PR #1, opened 2019-06-07
WIP: GET-49 Dockerize application and update necessary dependencies 

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-49

* Upgrade to Ruby version 2.6.3
* Upgrade Nodejs to 10.14.2 
Although our initial decision was 10.16, whilst attempting to build a docker image using existing packages the following mismatch happened:
- Nodejs 10.16 only exists on edge and not on alpine, and there are no Docker Ruby edge images:
https://hub.docker.com/_/ruby/
https://pkgs.alpinelinux.org/packages?name=nodejs&branch=edge
- attempting to use a pre-built Nodejs 10.16 alpine Docker image, or any alpine version does not contain the newest ruby versions but up to 2.5.5
https://pkgs.alpinelinux.org/packages?name=ruby&branch=edge
This meant we could not setup something like this:
https://blog.codeship.com/build-minimal-docker-container-ruby-apps/
- attempting to build a custom alpine image whilst referring to the Ruby alpine Docker image and previous Dockerfiles I've built seemed to show that building it would mean we will have huge Docker files with many dependencies: https://github.com/docker-library/ruby/blob/9ae0943fa2935b3a13c72ae7d6afa2439145d7fa/2.6/stretch/slim/Dockerfile

I personally prefer something simple we could maintain, as from previous experience maintaining so many dependancies has been quite unwieldy and breaks frequently. I have checked the CVEs for Nodejs 10.14 and there seem to be none: https://www.cvedetails.com/vulnerability-list/vendor_id-12113/Nodejs.html
so I think it would be safe for us to use. Happy to hear people's thoughts on this.


Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+# Update and install all of the required packages.
+# At the end, remove the apk cache
+RUN apk add --update $BUILD_PACKAGES && \
+    rm -rf /var/cache/apk/*
+
+# Generate Assets
+WORKDIR /app
+COPY Gemfile Gemfile.lock .ruby-version ./
+RUN bundle install
```
smoothcontract: Is is worthwhile excluding development dependencies here? I think we'll need test deps to allow CI to run and then use the same image on production for simplicity. Omitting any dev stuff will likely keep the image a bit smaller.

smoothcontract: After discussion this section only builds assets so could also exclude `test` dependencies.

PR #4, opened 2019-06-11
GET-40 Job profile scraper

smoothcontract:
### Context
Exploring the possibility of scraping job profile and associated data directly from the NCS "Explore my careers" site as no API is currently available we can use. It certainly works although isn't an ideal technique.

For now I've setup a simplified model structure of skills, categories and job profiles. This is just a quick spike so no test coverage yet and the code is far from robust.

The sitemap from NCS is first parsed for a list of interesting links to category and job profile pages. Placeholder records are created for each of these with `slug` and `original_url` attributes.

Once the site map is imported the individual category and job profile pages can be scraped. The category page is scraped for links to job profiles - we assume that means the job profiles are within that category so we can join them together. We can also parse a more nicely formatted category name from the page.

Job profile pages are scraped for name and description, and a list of individual job skills. A skill record is created for each found skill (matching just by the full text) and joined to the job skill. This is limited to processing 10 job profile pages for now as there's no point scraping the whole site (800+ job profiles) until we decide which content is needed.

### Guidance to review
Run migrations and then the three rake tasks in order i.e.

```
bundle exec rails db:migrate
bundle exec rails data_import:import_sitemap
bundle exec rails data_import:scrape_categories
bundle exec rails data_import:scrape_job_profiles
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,12 @@ gem 'foreman'
 # Canonical meta tag
 gem 'canonical-rails'
 
+# Process web and ai requests
+gem 'httparty'
+
+# XML and HTML parsing
+gem 'nokogiri', '~> 1.8'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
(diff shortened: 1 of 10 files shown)


smoothcontract:
Closing in favour of https://github.com/DFE-Digital/get-help-to-retrain/pull/24 - this was just a quick hacky spike implementation.

PR #5, opened 2019-06-13
Get 58 capybara cucumber setup

jaffamonkey:
Added standard cucumber-rails setup, added feature files and steps definitions.


README.md
```diff
@@ -35,6 +35,12 @@ bundle exec rake
 bundle exec rspec
 ```
 
+## Running Cucumber/Capybara test
+```
+cd tests
```
smoothcontract: I don't think we need to run `cucumber` in this directory now (plus the name has changed).

features/GET-30.feature
```diff
@@ -0,0 +1,26 @@
+Feature: As a customer
```
smoothcontract: I don't think I'm totally sold on naming the features to match the Jira tickets. It's likely that we'll have features that span multiple Jira stories, plus it's less immediately obvious when working with the feature specs where we can find specific features; instead we need to cross reference each feature. Maybe add a tag with the Jira story within each feature, but name the feature files with something that identifies the feature e.g. for this one `journey_overview.feature`. What do you think?

features/GET-62.feature
```diff
@@ -0,0 +1,22 @@
...
+        Given I am on the "careers-search" page
+        When I click the text link "<Occupation>"
+        Then I should see a list of occupations for "<Occupation>"
+        Examples:
+            | Occupation     |
+            | Healthcare     |
+            | Administration |
+            | Animal Care    |
+
+    Scenario Outline: Find a Course Service unavailable
```
smoothcontract: I don't think we'll be using the Course API for some time. Is there a way to maybe mark this as a pending feature? Or perhaps annotate it for Phase 2?

features/step_definitions/test_steps.rb
```diff
@@ -0,0 +1,69 @@
+Given("I am on the {string} page") do |string|
+    visit 'http://localhost:3000/pages/' + string
```
smoothcontract: Can we use 2 spaces for indentation please? We'll shortly be introducing Rubocop to the project which will enforce/autocorrect formatting stuff like this so no biggie at present.

features/GET-30.feature
jaffamonkey: Its all open for change, anyone can make suggestions. And I have never had case of one feature (file) spanning multiple user stories so will need to see example of how that happens, as that sounds like a very differing approach. My approach is the feature is derived from user story (a more 1-2-1). After Sprint end, the feature file may "die" (removed from context of automation), it may live on usefully, it may expand. But in the last case there will be a be a new user story anyway, assuming it's not just a user story rolled over from previous Sprint). Tagging, naturally good idea. At the moment I keep tags to failing tests (@bdd) and the ones we always expect to be green (@ci). This branch is actually dead and should be deleted.  WIl do new PR and add tags for user-stories too. I should add I have kept this light with expectation of change.

README.md
jaffamonkey: OK, will change this in other branch - or can just keep to this one?

features/step_definitions/test_steps.rb
jaffamonkey: Ok, will adjust the detault

features/GET-30.feature
jaffamonkey: regards feature file naming convention, that's fine - i did similar, but your looks better so will change - I added jira ticket number as tags for scenario anyway.

PR #7, opened 2019-06-14
Add some factories and specs for models

smoothcontract:
### Context
I'll be using these factories shortly for setting up some test data fixtures, but I think this is already useful as a standalone PR.

This adds some factory bot factories for the key models using Faker to populate attributes. Also added some very simplistic tests to debug the has many relationships - all working nicely now.

### Guidance to review

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -1,4 +1,4 @@
 class Category < ApplicationRecord
   has_many :job_profile_categories
-  has_many :job_profiles, through: :job_profile_categories
+  has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
 end
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -1,3 +1,4 @@
 class Category < ApplicationRecord
+  has_many :job_profile_categories
   has_many :job_profiles, through: :job_profile_categories
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -1,6 +1,6 @@
 class JobProfile < ApplicationRecord
   has_many :job_profile_categories
   has_many :job_profile_skills
-  has_many :categories, through: :job_profile_categories
-  has_many :skills, through: :job_profile_skills
+  has_many :categories, through: :job_profile_categories, inverse_of: :job_profiles
+  has_many :skills, through: :job_profile_skills, inverse_of: :job_profiles
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -1,4 +1,6 @@
 class JobProfile < ApplicationRecord
+  has_many :job_profile_categories
+  has_many :job_profile_skills
   has_many :categories, through: :job_profile_categories
   has_many :skills, through: :job_profile_skills
 end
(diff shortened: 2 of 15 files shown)


app/models/category.rb
```diff
@@ -1,3 +1,4 @@
 class Category < ApplicationRecord
+  has_many :job_profile_categories
```
cwrw: could we add the `inverse_of` for all the associations?

app/models/job_profile.rb
```diff
@@ -1,4 +1,6 @@
 class JobProfile < ApplicationRecord
+  has_many :job_profile_categories
```
cwrw: If a `JobProfileCategory` is deleted, do we want to persist Job Profiles?

app/models/category.rb
smoothcontract: Hadn’t come across that before so googled - seems in Rails > 4 those are automatically determined 🤘

https://makandracards.com/makandra/12331-rails-when-to-use-inverse_of-in-has_many-has_one-or-belongs_to-associations

spec/models/category_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+RSpec.describe Category do
+  let(:category) { FactoryBot.build_stubbed(:category) }
```
cwrw: you can call `build_stubbed` directly without using `FactoryBot` as so: `let(:category) {  build_stubbed(:category) }`

app/models/category.rb
cwrw: It seems for the `through` relation they aren't:
```
Below there are some known options where a automatically determination of the inverse is not working:

:conditions, :through ...
```
Could we add them there?

app/services/seed_service.rb
```diff
@@ -0,0 +1,42 @@
...
+  class << self
+    def seed
+      only_development!
+      delete_all
+      seed_categories
+      seed_skills
+      seed_job_profiles
+    end
+
+    private
```
cwrw: this is odd, according the linter:
```
Linting ruby...
app/services/seed_service.rb:11:5: C: [Corrected] Layout/AccessModifierIndentation: Outdent access modifiers like private. (https://github.com/rubocop-hq/ruby-style-guide#indent-public-private-protected)
    private
    ^^^^^^^
```

cwrw: I would have assumed what you did was correct 🤷‍♂

smoothcontract: Hmm.. maybe private class methods are not a thing. I'll refactor to move some of that stuff to instance methods.

smoothcontract:
> I don't usually test associations as they are part of the framework. When we add implementation we will be testing all these associations and behaviours fully, so I would recommend we remove those for now. What are your thoughts?

I'd agree with this, but having the specs within this PR at least has been useful to get the relationships defined correctly. Once we have some more realistic test coverage (i.e. anything that actually exercises the models) we can remove the association tests as part of the relevant PR.

app/models/job_profile.rb
smoothcontract: Good question - I'd assume yes for now. I may revise this when working further on the web scraper.

PR #8, opened 2019-06-16
GET-49 Set up CI with Azure Pipelines

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-49 

* Docker image now builds test dependencies (needs rethink)
* Add azure pipeline to run tests and build images per commit and tag with build number and Git SHA
* If this is master is will build without cache and with the tag: latest
* If build is successful it will push images to DockerHub
* Things to look out for:
  - link is deprecated, could we use docker-compose instead?
  - could not make use of service containers as network would not allow connections properly
  - using separate jobs for testing and deploy (deploy job) did not allow us to share the image build 
    meaning we had to rebuild it, so for now everything is in a single job

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,18 +11,11 @@ RUN apk add --update $BUILD_PACKAGES && \
 # Generate Assets
 WORKDIR /app
 COPY Gemfile Gemfile.lock .ruby-version ./
-RUN bundle install --clean --without "test development"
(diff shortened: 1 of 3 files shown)


mafonso:
LGTM as a starting point.
While we should look for options to avoid the deprecated `--link´ this is functional and provides all we need at this point.

PR #9, opened 2019-06-17
Add Rubocop ruby linter and fix existing offences

tavy88:
### Context

Adding Rubocop ruby linter and fix existing offences.

### Notes

Strongly recommend configuring Rubocop with the preferred code editor. Very useful. 

### Screenshots

```ruby
➜  get-help-to-retrain git:(GET-96-rubocop) ✗ rubocop
Inspecting 34 files
..................................

34 files inspected, no offenses detected
```


Gemfile
```diff
@@ -54,6 +54,10 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+
+  # Ruby code analyzer and formatter
+  gem 'rubocop'
```
cwrw: could we add the versions here?

.rubocop.yml
```diff
@@ -0,0 +1,94 @@
+AllCops:
+  DisplayCopNames: true
```
cwrw: Im more inclined to start with the bare minimum set of rules in `.rubocop.yml` and build them as we go along, and include the folders we can't fix, and attempt to fix the ones that are simple like: `spec/spec_helper.rb` etc.
```
AllCops:
  DisplayCopNames: true
  DisplayStyleGuide: true
  TargetRubyVersion: 2.6.3
  Exclude:
    - 'db/schema.rb'
    - 'vendor/**/*'
    - bin/**
    - db/**/*
    - 'node_modules/**/*'
    - 'tmp/**/*'
    - 'lib/tasks/cucumber.rake'
    - 'config/initializers/canonical_rails.rb'
Style/FrozenStringLiteralComment:
  Enabled: false
Style/Documentation:
  Enabled: false
Metrics/LineLength:
  Max: 120
```
My main reasoning is that it is preferable to start with the vanilla rubocop settings and then see which ones we might not want as they crop along which I'm sure we will. What do you think?

Gemfile
tavy88: Do you want to lock it to a version?

.rubocop.yml
tavy88: This is something I was using at a previous place. So I'm open to any kind of rules the team agrees to. Mine might probably be too strict.

PR #10, opened 2019-06-17
Add SimpleCov is a code coverage analysis tool

tavy88:
### Context

Add SimpleCove code coverage analysis tool

### Notes

Open `coverage/index.html` after running rspec if you want to see the report.

### Screenshots

image


Gemfile
```diff
@@ -63,6 +63,9 @@ group :test do
   gem 'selenium-webdriver'
   # Easy installation and use of chromedriver to run system tests with Chrome
   gem 'chromedriver-helper'
+
+  # Test coverage reporting
+  gem 'simplecov', require: false
```
smoothcontract: Would be nice to tighten up the version a little bit here if possible, maybe

```ruby
gem 'simplecov', '~> 0.16', require: false
```

tavy88: Sure thing. I don't usually lock to a specific version or tighten it up but I am flexible on it.

smoothcontract: I think it's good practice to lock to a minor release point (assuming most gems use semantic versioning). I know we have `Gemfile.lock` but that gets ignored if we do a wholesale upgrade of a lot of gems and then it's tricky to figure out which gems would work with the app. I like this approach as we can still upgrade gems (within reason), hopefully without any breaking changes, in a more controlled fashion. See https://github.com/thoughtbot/guides/tree/master/best-practices#bundler

PR #12, opened 2019-06-17
Add security checker tool brakeman

tavy88:
### Context

Adding brakeman for security check.

### Notes

Run `brakeman` or `brakeman -o output.html` - if you prefer a report generated in a nice way.

Check https://brakemanscanner.org/docs/ for more options.


Gemfile
```diff
@@ -47,6 +47,9 @@ group :development, :test do
 end
 
 group :development do
+  # Security scanner
+  gem 'brakeman'
```
cwrw: could we add this in the section `group :development, :test do` as we will be running it in our CI build, also could we tie this to a certain version to avoid issues down the line: `gem 'brakeman', '~> 4.5'`

README.md
```diff
@@ -46,3 +46,22 @@ or
...
+To get the report simply run:
+
+```bash
+  brakeman
+```
+
+or 
+
+```bash
+  brakeman -o report.html
```
cwrw: could you add this to our azure-pipeline as step to run as part of the CI build? https://github.com/DFE-Digital/get-help-to-retrain/blob/master/azure-pipelines.yml#L60 as 
```
- script: |
        $DOCKER_RUN bundle exec brakeman
      displayName: Brakeman
      env:
        DOCKER_RUN: $(dockerRun)
```

Gemfile
smoothcontract: +1

README.md
tavy88: Yep was actually going to ask if we are ok to put it as part of the Azure checks. Glad you confirmed that. Thanks.

.gitignore
```diff
@@ -27,3 +27,4 @@ yarn-debug.log*
 
 .env
 *.log
+coverage
```
cwrw: could you add `/coverage/*` instead?

cwrw:
Build is red because of our first brakeman failure, if you run `bundle exec brakeman` locally you'll be able to see/fix it

tavy88:
@cwrw Some brakeman warnings will never be fixed (some are just false positives). I think we need to find a way to not block the deploy because of the brakeman. I had similar issues in the past. We just ended up just keeping it as a check but not block the merge/deploy unless the severity was medium - high.

PR #15, opened 2019-06-18
GET-56 Build Job Profile Search page and Results page

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-56

* Build Job Profile responsive search and results page 
* Search is case insensitive
* Add base for beta banner and breadcrumbs

### DIFF
--- a/app/controllers/job_profiles_controller.rb
+++ b/app/controllers/job_profiles_controller.rb
@@ -1,0 +1,15 @@
+class JobProfilesController < ApplicationController
+  def search
+    @job_categories = Category.with_job_profiles
+  end
+
+  def index
+    @job_profiles = JobProfile.search(job_profile_params[:name])
+  end
+
+  private
+
+  def job_profile_params
+    params.permit(:name)
+  end
+end
--- a/app/controllers/job_profiles_controller.rb
+++ b/app/controllers/job_profiles_controller.rb
@@ -1,8 +1,4 @@
 class JobProfilesController < ApplicationController
-  def search
-    @job_categories = Category.with_job_profiles
-  end
-
   def index
     @job_profiles = JobProfile.search(job_profile_params[:name])
   end
@@ -10,6 +6,6 @@ class JobProfilesController < ApplicationController
   private
 
   def job_profile_params
-    params.permit(:name)
+    params.require(:job_profile).permit(:name)
   end
 end
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -4,6 +4,7 @@ class Category < ApplicationRecord
 
   def self.with_job_profiles
     includes(:job_profile_categories)
-      .where.not(job_profile_categories: { job_profile_id: nil })
+      .where
+      .not(job_profile_categories: { job_profile_id: nil })
   end
 end
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -1,4 +1,9 @@
 class Category < ApplicationRecord
   has_many :job_profile_categories
   has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
+
+  def self.with_job_profiles
+    includes(:job_profile_categories)
+      .where.not(job_profile_categories: { job_profile_id: nil })
+  end
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -3,4 +3,8 @@ class JobProfile < ApplicationRecord
   has_many :job_profile_skills
   has_many :categories, through: :job_profile_categories, inverse_of: :job_profiles
   has_many :skills, through: :job_profile_skills, inverse_of: :job_profiles
+
+  def self.search(name)
+    where('name LIKE ?', "%#{name}%") if name
+  end
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -5,6 +5,6 @@ class JobProfile < ApplicationRecord
   has_many :skills, through: :job_profile_skills, inverse_of: :job_profiles
 
   def self.search(name)
-    where('name LIKE ?', "%#{name}%") if name
+    where('name ILIKE ?', "%#{name}%")
   end
 end
--- a/app/views/job_profiles/index.html.erb
+++ b/app/views/job_profiles/index.html.erb
@@ -8,13 +8,13 @@
   <li class="govuk-breadcrumbs__list-item"><%= t('breadcrumb.job_profiles_result') %></li>
 <% end %>
 
-<div class="govuk-grid-row govuk-!-margin-top-7">
+<div class="govuk-grid-row govuk-!-padding-top-7">
   <div class="govuk-grid-column-two-thirds" >
     <h3 class="govuk-heading-l"><%= t('.title') %></h1>
     <div class="govuk-grid-column-full govuk-!-padding-0">
       <%= form_tag job_profiles_path, method: 'get' do %>
         <div class="govuk-form-group">
-          <%= text_field_tag(:name, nil, class: 'search-input govuk-input govuk-!-width-one-half', placeholder: t('.placeholder')) %>
+          <%= text_field(:job_profile, :name, value: params[:job_profile][:name], class: 'search-input govuk-input govuk-!-width-one-half', placeholder: t('.placeholder')) %>
           <%= button_tag('', class: 'search-button-results', name: nil) %>
         </div>
       <% end %>
--- a/app/views/job_profiles/index.html.erb
+++ b/app/views/job_profiles/index.html.erb
@@ -1,16 +1,40 @@
-<%= form_tag job_profiles_path, method: 'get' do %>
-  <p>
-    <%= text_field_tag(:name, nil, placeholder: 'e.g. Software Developer') %>
-    <%= submit_tag('Search Job Profiles', name: nil) %>
-  </p>
+<% content_for :breadcrumb do %>
+  <li class="govuk-breadcrumbs__list-item">
+    <%= link_to(t('breadcrumb.home'), root_path, class: 'govuk-breadcrumbs__link') %>
+  </li>
+  <li class="govuk-breadcrumbs__list-item">
+    <%= link_to(t('breadcrumb.job_profiles_search'), search_job_profiles_path, class: 'govuk-breadcrumbs__link') %>
+  </li>
+  <li class="govuk-breadcrumbs__list-item"><%= t('breadcrumb.job_profiles_result') %></li>
 <% end %>
 
-<% if @job_profiles.empty? %>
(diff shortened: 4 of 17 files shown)


spec/models/job_profile_spec.rb
```diff
@@ -30,4 +30,24 @@
...
+      job_profile = create(:job_profile, name: 'Beverage Dissemination Officer')
+
+      expect(described_class.search('Dissemination')).to contain_exactly(job_profile)
+    end
+
+    it 'returns nothing if no job profile is matched' do
+      create(:job_profile, name: 'Dream Alchemist')
+
+      expect(described_class.search('Beverage Dissemination Officer')).to be_empty
+    end
```
cwrw: add extra test for `nil` name

app/models/category.rb
```diff
@@ -1,4 +1,9 @@
 class Category < ApplicationRecord
   has_many :job_profile_categories
   has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
+
+  def self.with_job_profiles
+    includes(:job_profile_categories)
+      .where.not(job_profile_categories: { job_profile_id: nil })
```
cwrw: new line at `.not`

app/webpacker/styles/_job-profile.scss
```diff
@@ -0,0 +1,27 @@
...
+    color: govuk-colour("white")
+  }
+}
+.search-button {
+  border: 0;
+  width:40px;
+  height:40px;
+  float: left;
+  background-color: govuk-colour("black");
+  background-image: url("../images/search_icon.svg");
```
cwrw: add assets url instead

app/models/job_profile.rb
```diff
@@ -3,4 +3,8 @@ class JobProfile < ApplicationRecord
   has_many :job_profile_skills
   has_many :categories, through: :job_profile_categories, inverse_of: :job_profiles
   has_many :skills, through: :job_profile_skills, inverse_of: :job_profiles
+
+  def self.search(name)
+    where('name LIKE ?', "%#{name}%") if name
```
smoothcontract: Maybe use ILIKE (case independent) ?

app/controllers/job_profiles_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def index
+    @job_profiles = JobProfile.search(job_profile_params[:name])
+  end
+
+  private
+
+  def job_profile_params
+    params.permit(:name)
```
cwrw: add require `.require(:job_profile)`

app/models/job_profile.rb
smoothcontract: Alternatively there’s a postgres column type that supports case independent comparison operators by default - we could swap to that. Would have to google that, can’t recall off the top of my head.

spec/models/category_spec.rb
```diff
@@ -13,4 +13,30 @@
...
+    end
+
+    it 'does not return categories not linked to job profiles' do
+      category = create(:category, job_profiles: [create(:job_profile)])
+      create(:category)
+      expect(Category.with_job_profiles).to contain_exactly(category)
+    end
+
+    it 'returns nothing if no categories linked to job profiles' do
+      create_list(:category, 4)
```
smoothcontract: Maybe just create 2 categories here? That’s enough to prove the spec without slowing it unnecessarily.

app/models/job_profile.rb
smoothcontract: See https://nandovieira.com/using-insensitive-case-columns-in-postgresql-with-citext

cwrw: I've went with `ILIKE` for now, I think we need to think generally about search and what we would like it's behaviour to be like before committing to anything past queries

app/views/shared/_contact_us.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="govuk-grid-column-one-third">
+  <div class="govuk-related-items govuk-!-padding-bottom-4 top-border__container">
+    <h2 class="govuk-heading-m"><%= t('contact_us.title') %></h2>
+    <p class="govuk-body-m">
+      <%= t('contact_us.body') %> <a href="tel:+442079476330" <%= t('contact_us.telephone_link') %> class="govuk-!-padding-left-1 govuk-link"><%= t('contact_us.telephone') %></a>
```
cwrw: fix href

spec/models/job_profile_spec.rb
```diff
@@ -30,4 +30,60 @@
       expect(recommended_job.skills).to match [skill]
     end
   end
+
+  describe '.search' do
+    it 'returns a job profile if a name matches exactly' do
+      job_profile = create(:job_profile, name: 'Beverage Dissemination Officer')
```
tavy88: could we use `let(:job_profile)/let!` ?

spec/models/job_profile_spec.rb
```diff
@@ -30,4 +30,60 @@
...
+
+      expect(described_class.search('Beverage Dissemination Officer')).to contain_exactly(job_profile)
+    end
+
+    it 'returns a job profile if a name is like an existing job profile name' do
+      job_profile = create(:job_profile, name: 'Beverage Dissemination Officer')
```
tavy88: could we use let(:job_profile)/let! ?

spec/models/job_profile_spec.rb
cwrw: I think there are two approaches, sharing objects across tests vs. creating tests which are self contained. I am following the trend of self contained tests where we create everything in there. If there is something bigger or more intricate we can move it out to a let

spec/models/job_profile_spec.rb
cwrw: same above

app/views/job_profiles/index.html.erb
```diff
@@ -0,0 +1,40 @@
...
+    <%= link_to(t('breadcrumb.job_profiles_search'), search_job_profiles_path, class: 'govuk-breadcrumbs__link') %>
+  </li>
+  <li class="govuk-breadcrumbs__list-item"><%= t('breadcrumb.job_profiles_result') %></li>
+<% end %>
+
+<div class="govuk-grid-row govuk-!-padding-top-7">
+  <div class="govuk-grid-column-two-thirds" >
+    <h3 class="govuk-heading-l"><%= t('.title') %></h1>
+    <div class="govuk-grid-column-full govuk-!-padding-0">
+      <%= form_tag job_profiles_path, method: 'get' do %>
```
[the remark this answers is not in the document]
cwrw: I think for now it's best to keep them separate as this form is different across the index/search pages regarding styles.

app/controllers/job_profiles_controller.rb
```diff
@@ -0,0 +1,15 @@
+class JobProfilesController < ApplicationController
+  def search
+    @job_categories = Category.with_job_profiles
+  end
```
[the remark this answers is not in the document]
cwrw: This was added initially as we thought we'd be including a list of categories on this page, I'll remove as it's not needed on the page anymore

app/controllers/job_profiles_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def index
+    @job_profiles = JobProfile.search(job_profile_params[:name])
+  end
+
+  private
+
+  def job_profile_params
+    params.require(:job_profile).permit(:name)
+  end
```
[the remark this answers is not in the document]
cwrw: yes, we want to only allow name for now

app/models/category.rb
```diff
@@ -1,4 +1,10 @@
 class Category < ApplicationRecord
   has_many :job_profile_categories
   has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
+
+  def self.with_job_profiles
+    includes(:job_profile_categories)
+      .where
+      .not(job_profile_categories: { job_profile_id: nil })
+  end
```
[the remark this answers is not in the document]
cwrw: I think for now it's safe to leave as is until we get to work on this in the job category ticket, as it seems we don't need this anymore in this ticket

app/controllers/job_profiles_controller.rb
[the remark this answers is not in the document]
tavy88: 👍

app/models/category.rb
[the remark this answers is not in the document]
tavy88: 👍

app/views/job_profiles/index.html.erb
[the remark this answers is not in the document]
tavy88: I doesn't need to be a shared partial, but it's more about drying this index page.

spec/models/job_profile_spec.rb
tavy88: Ok, but using a let would dry your test steps as you are calling create multiple times.

app/views/job_profiles/index.html.erb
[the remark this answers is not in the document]
cwrw: I think both the search and index page are quite small, and not worth extracting partials out right now (unless they are reused then that would be a good case to start). If we notice the pages becoming unwieldy we can start doing that.

app/views/job_profiles/search.html.erb
```diff
@@ -0,0 +1,22 @@
+<% content_for :breadcrumb do %>
+  <li class="govuk-breadcrumbs__list-item">
+    <%= link_to(t('breadcrumb.home'), root_path, class: 'govuk-breadcrumbs__link') %>
+  </li>
+  <li class="govuk-breadcrumbs__list-item"><%= t('breadcrumb.job_profiles_search') %></li>
+<% end %>
+
```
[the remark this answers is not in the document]
cwrw: same as above

app/views/job_profiles/index.html.erb
```diff
@@ -0,0 +1,40 @@
+<% content_for :breadcrumb do %>
+  <li class="govuk-breadcrumbs__list-item">
+    <%= link_to(t('breadcrumb.home'), root_path, class: 'govuk-breadcrumbs__link') %>
+  </li>
+  <li class="govuk-breadcrumbs__list-item">
+    <%= link_to(t('breadcrumb.job_profiles_search'), search_job_profiles_path, class: 'govuk-breadcrumbs__link') %>
+  </li>
+  <li class="govuk-breadcrumbs__list-item"><%= t('breadcrumb.job_profiles_result') %></li>
+<% end %>
+
```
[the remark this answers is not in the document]
cwrw: I think breadcrumbs is a good candidate for extraction to shared partials. It would be good to see a pattern emerge first (further from 2 views) so we can build a helper for those. For now we can keep the same until then

PR #17, opened 2019-06-19
Introduce Version model

smoothcontract:
### Context
The intent here is to allow a future/draft version of the job profile and associated models to be created without breaking the existing job profile data in the meantime.

Once the screen scraping completes successfully the draft version becomes the current version. I've also added support for keeping a previous version - not sure if that will be useful but no harm in adding it for now and removing if it proves unnecessary.

This means that when working with the core `JobProfile`, `Category` and `Skill` models we will instead need to refer to `Version.current.job_profiles`, `Version.current.categories` and `Version.current.skills` respectively.

### Guidance to review
This is still WIP and just a proposed approach for now - if we like this then I'll flesh it out more.

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -1,4 +1,5 @@
 class Category < ApplicationRecord
-  has_many :job_profile_categories
+  belongs_to :version
+  has_many :job_profile_categories, dependent: :delete_all
   has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -1,4 +1,5 @@
 class JobProfile < ApplicationRecord
+  belongs_to :version
   has_many :job_profile_categories
   has_many :job_profile_skills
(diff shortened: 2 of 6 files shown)


tavy88:
I like the idea of versioning. I can't see a set of tests, but wondering if it would happen to revert to a previous version if one has already promoted especially if there are structural changes in the schema e.g renaming a column/dropping one. 

Have you checked https://github.com/paper-trail-gem/paper_trail?

It's a very good gem for versioning.

smoothcontract:
> I like the idea of versioning. I can't see a set of tests, but wondering if it would happen to revert to a previous version if one has already promoted especially if there are structural changes in the schema e.g renaming a column/dropping one.
> 
> Have you checked https://github.com/paper-trail-gem/paper_trail?
> 
> It's a very good gem for versioning.

I've used papertrail previously but don't think it's quite the right fit for this usecase. We don't need to track changes to individual model attributes.

tavy88:
Yep fair. Agreed as we discussed.

smoothcontract:
Closing for now.

PR #18, opened 2019-06-19
Qa sprint2

jaffamonkey:
Sprint 2 Update for passing stories GET-35 and GET-56

The two stories GET-35 and GET-56 have passing scenarios, and tagged 'ci'.  And added code to enable running tests through various browsers/methods (chrome headless currently).


features/support/env.rb
```diff
@@ -36,6 +36,10 @@
   raise "You need to add database_cleaner to your Gemfile (in the :test group) if you wish to use it."
 end
 
+Before do
+  $session = Capybara::Session.new(:selenium_chrome_headless)
+  # $session = Capybara::Session.new(:selenium, browser: :chrome)
```
tavy88: This is more like a note: do we promote commented code to production?

tavy88: Also, I think we should move this session init in a shared step and avoid using global variables. Then you can just do the init in every scenario and use `@session`.

cwrw: I've removed this completely for now. I think it might be easier for us to use drivers/tags instead

features/step_definitions/test_steps.rb
```diff
@@ -1,21 +1,29 @@
...
+end
+
+Then("the current page contains text {string}") do |content|
+  expect(page).to have_content(content)
 end
 
 Then("the first search result title should contain {string}") do |string|
-  expect(page).to have_content(string)
+  find('a', text: string)
+  # expect(page).to have_selector 'h3', text: string
```
tavy88: No expectation here?

features/support/env.rb
```diff
@@ -1,58 +1,18 @@
...
-# 2) Set the value below to true. Beware that doing this globally is not
-# recommended as it will mask a lot of errors for you!
-#
 ActionController::Base.allow_rescue = false
 
-# Remove/comment out the lines below if your app doesn't have a database.
-# For some databases (like MongoDB and CouchDB) you may need to use :truncation instead.
 begin
-  DatabaseCleaner.strategy = :transaction
+  DatabaseCleaner.strategy = :truncation
```
tavy88: Truncation is really slow. Wonder why did you need to switch to this?

cwrw: This was quite interesting. So when we created an instance of Job Profile, because it was wrapped in a transaction, the controller couldn't access it, so unfortunately we will have to use truncation. Also its the default setup for cucumber as seen in : https://github.com/DatabaseCleaner/database_cleaner#cucumber-example

features/step_definitions/test_steps.rb
cwrw: looks like a rogue step, I'll remove

features/support/env.rb
jaffamonkey: > This is more like a note: do we promote commented code to production?

it's junk :)

features/step_definitions/test_steps.rb
jaffamonkey: More junk, yes :)

PR #22, opened 2019-06-21
Include assets folder as well as packs

cwrw:
Images from node modules are included under different directory
include that when porting assets to final container


mafonso:
👍

PR #24, opened 2019-06-24
GET-80 Explore My Careers Scraper

smoothcontract:
### Context
Adds rake tasks for scraping job profiles, categories and skills from National Careers Service "Explore My Careers" service.

https://dfedigital.atlassian.net/browse/GET-80

The sitemap from NCS is first parsed for a list of interesting links to category and job profile pages. Placeholder records are created for each of these with `slug` and `original_url` attributes. Introduced `wombat` gem which has a nice DSL for parsing XML and HTML documents.

Once the site map is imported the individual category and job profile pages can be scraped. The category page is scraped for links to job profiles - we assume that means the job profiles are within that category so we can join them together. We can also parse a more nicely formatted category name from the page.

Job profile pages are scraped for name and description, and a list of individual job skills. A skill record is created for each found skill (matching just by the full text) and joined to the job skill. The entire scraped page is also stored as `content` attribute for further parsing later.

**TODO:** Add validation to models to ensure that content has been scraped correctly, also ensure that existing content is not destroyed if scraping fails part way through. (This will be covered in a follow up PR).

### Guidance to review
Run the three rake tasks in order i.e.

```
bundle exec rails data_import:import_sitemap
bundle exec rails data_import:scrape_categories
bundle exec rails data_import:scrape_job_profiles
```


Gemfile
```diff
@@ -83,6 +86,10 @@ group :test do
 
     # Test coverage reporting
   gem 'simplecov', '~> 0.16', require: false
+
+  # Web request caching for tests
+  gem 'vcr', '~> 5.0', require: false
```
cwrw: discussed in person but for posterity: I think we should avoid using VCR as it doesn't create human readable fixtures which will be highly problematic down the line when we want to change fixtures. I think if we add the html itself to the fixtures and then use webmock we should be fine for now

smoothcontract: I like the ease of use VCR adds but agree it's not ideal having the cassette stored in an encoded format which makes diffs hard. I'm having a quick look at that now to see if it's an easy fix that can be included in this PR.

spec/scrapers/sitemap_scraper_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  describe '#scrape' do
+    subject { scraper.scrape }
+
+    it 'returns an array of urls' do
+      expect(subject).to have_key 'urls'
+      expect(subject['urls']).to be_an(Array)
+    end
+  end
+
+  describe
```
cwrw: ✂️

spec/rails_helper.rb
```diff
@@ -1,9 +1,9 @@
 require 'spec_helper'
 ENV['RAILS_ENV'] ||= 'test'
 require File.expand_path('../config/environment', __dir__)
-abort('The Rails environment is running in production mode!') if Rails.env.production?
```
cwrw: What we the reason this was required? I think it's important we add it back

.rubocop.yml
```diff
@@ -23,3 +23,6 @@ Style/GuardClause:
   MinBodyLength: 2
 Layout/AccessModifierIndentation:
   Enabled: false
+Lint/AmbiguousBlockAssociation:
```
cwrw: would we add this to `spec/.rubocop.yml` instead as the failures were in the tests?

spec/rails_helper.rb
smoothcontract: Rebase snafoo, restored

.rubocop.yml
```diff
@@ -23,3 +23,4 @@ Style/GuardClause:
   MinBodyLength: 2
 Layout/AccessModifierIndentation:
   Enabled: false
+
```
cwrw: 👻 ✂️

PR #25, opened 2019-06-24
GET-62 Add Explore Occupations page with category list

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-62

* add css linter to pipeline
* move check your existing skills out of Job Profile controller into its own controller/views
* add explore occupations controller/views with categories list
* tidy up views and add partials for shared forms

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -1,7 +1,7 @@
 class Category < ApplicationRecord
   has_many :job_profile_categories
   has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
-  default_scope { order(name: :asc) }
+  scope :by_name, -> { order(:name) }
 
   def self.with_job_profiles
     includes(:job_profile_categories)
@@ -10,6 +10,9 @@ class Category < ApplicationRecord
   end
 
   def self.with_job_profiles_without(category)
-    Category.with_job_profiles - [category]
+    Category
+      .with_job_profiles
+      .by_name
+      .where.not(id: category)
   end
 end
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -1,10 +1,15 @@
 class Category < ApplicationRecord
   has_many :job_profile_categories
   has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
+  default_scope { order(name: :asc) }
 
   def self.with_job_profiles
     includes(:job_profile_categories)
       .where
       .not(job_profile_categories: { job_profile_id: nil })
+  end
+
+  def self.with_job_profiles_without(category)
+    Category.with_job_profiles - [category]
   end
 end
--- a/app/views/categories/show.html.erb
+++ b/app/views/categories/show.html.erb
@@ -1,0 +1,38 @@
+<% content_for :breadcrumb do %>
+  <li class="govuk-breadcrumbs__list-item">
+    <%= link_to(t('breadcrumb.home'), root_path, class: 'govuk-breadcrumbs__link') %>
+  </li>
+  <li class="govuk-breadcrumbs__list-item">
+    <%= link_to(t('breadcrumb.explore_occupations'), explore_occupations_path, class: 'govuk-breadcrumbs__link') %>
+  </li>
+  <li class="govuk-breadcrumbs__list-item"><%= t('breadcrumb.job_category') %></li>
+<% end %>
+
+<div class="govuk-grid-row govuk-!-padding-top-7">
+  <div class="govuk-grid-column-two-thirds" >
+    <h3 class="govuk-heading-l"><%= @category.name %></h3>
+      <ul class="govuk-list">
+        <% @category.job_profiles.each do |job_profile| %>
+          <li class="govuk-!-padding-bottom-2 govuk-!-padding-top-4">
+            <h3 class="govuk-!-margin-0">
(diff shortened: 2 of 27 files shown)


app/models/category.rb
```diff
@@ -1,10 +1,15 @@
 class Category < ApplicationRecord
   has_many :job_profile_categories
   has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
+  default_scope { order(name: :asc) }
```
smoothcontract: I'd really like to avoid adding default scopes - could a scope of `by_name` or something similar be added explicitly here please and then include where needed ?

app/views/categories/show.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :breadcrumb do
+    generate_breadcrumbs(
+      t('breadcrumb.job_category'),
+      [
+        [t('breadcrumb.home'), root_path],
+        [t('breadcrumb.explore_occupations'), explore_occupations_path]
+      ]
+    )
+  end
+%>
```
smoothcontract: This is okay for now but I'm hoping in time we can come up with a better mechanism that doesn't rely on each view having to know which breadcrumbs were followed to reach this page.

app/views/explore_occupations/results.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      <ul class="govuk-list">
+        <% @job_profiles.each do |job_profile| %>
+          <li class="govuk-!-padding-bottom-1">
+            <h3 class="govuk-!-margin-0">
+              <a class="govuk-link no-text-decoration" href="#"><%= job_profile.name %></a>
+            </h3>
+            <p class="govuk-!-margin-0 govuk-!-margin-bottom-3"><%= job_profile.description %></p>
+            <p class="govuk-body-s govuk-!-margin-bottom-1 strong">Salary: £15,500 to £21,000</p>
+            <p class="govuk-body-s muted-text">Found in:
+              <%= job_profile_category_list(job_profile) %>
```
smoothcontract: Not sure if it's necessary to add a helper for this, could iterate over the collection of categories in the view similar to the approach used here: https://github.com/DFE-Digital/get-help-to-retrain/pull/25/files#diff-ffb75260f330b885690ab8599cab45f9R22

I think no need to change this, but maybe let's not add more helpers than necessary going forwards? Open to discussion on this too.

azure-pipelines.yml
```diff
@@ -64,6 +64,12 @@ jobs:
       env:
         DOCKER_RUN: $(dockerRun)
 
+    - script: |
```
smoothcontract: Is this change related to the PR?

spec/models/category_spec.rb
```diff
@@ -42,4 +42,17 @@
...
+
+  describe '.with_job_profiles_without' do
+    it 'returns categories with job profiles without passed in category' do
+      category1 = create(:category, name: 'Administration', job_profiles: [create(:job_profile)])
+      category2 = create(:category, name: 'Education', job_profiles: [create(:job_profile)])
+      expect(described_class.with_job_profiles_without(category1)).to contain_exactly(category2)
+    end
+
+    it 'returns categories with job profiles if category empty' do
+      category = create(:category, name: 'Administration', job_profiles: [create(:job_profile)])
```
smoothcontract: It might be nice to add a trait to the `category` factory e.g. `with_job_profile` to support this. I don't want to use default associations within the factories though as this results in unnecessary models being created for specs that don't need them.

app/views/explore_occupations/results.html.erb
cwrw: This helper was added to comma delimit the category list, and not just iterating over the categories

azure-pipelines.yml
cwrw: Could extract this to it's own pr, but thought I'd lump it in here as I needed to touch more styles

app/views/categories/show.html.erb
cwrw: I couldn't think of anything past `content_for` and specific links, do you have anything in mind which is cleaner?

azure-pipelines.yml
smoothcontract: Okay, it's cool to include it here.

app/views/categories/show.html.erb
smoothcontract: Nothing concrete yet, let's see how this evolves as we add more pages

app/models/category.rb
```diff
@@ -1,10 +1,15 @@
 class Category < ApplicationRecord
   has_many :job_profile_categories
   has_many :job_profiles, through: :job_profile_categories, inverse_of: :categories
+  scope :by_name, -> { order(name: :asc) }
```
tavy88: ORDER ASC is by default.

app/models/category.rb
```diff
@@ -1,10 +1,15 @@
...
 
   def self.with_job_profiles
     includes(:job_profile_categories)
       .where
       .not(job_profile_categories: { job_profile_id: nil })
   end
+
+  def self.with_job_profiles_without(category)
+    Category.with_job_profiles.by_name - [category]
```
tavy88: Why not use a `.where.not` clause here to filter out?

app/models/category.rb
cwrw: its default per id not name

PR #26, opened 2019-06-24
Use last commit sha

mafonso:
### Context
We want to use the last commit (short) sha to identify our builds. 

Use the short sha of the last commit (head)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -18,7 +18,7 @@ jobs:
 
   steps:
     - script: |
-        git_sha=$(echo $(Build.SourceVersion) | cut -c 1-7)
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
   steps:
     - script: |
-        git_sha=$(echo $(Build.SourceVersion) | cut -c 1-7)
+        git_sha=$(git log -n 1 | head -n 1 | sed -e 's/^commit //' | head -c 7)
```
cwrw: I think we might get what we want with `git rev-parse --short HEAD`, what do you think?

mafonso: yep, it's cleaner and yields the same result. 👍

PR #31, opened 2019-06-25
Postgres firewall rules

mafonso:
### Context

In order for the app to access the database we need to define some firewall rules.

Allow webapp outbound IPs to access the database

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -38,6 +38,12 @@
         "description": "The default port for the psql server."
       }
     },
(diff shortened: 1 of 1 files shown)


azure/template.json
```diff
@@ -252,4 +282,4 @@
       "value": "[variables('workerAppServiceName')]"
     }
   }
-}
+}
```
cwrw: ✂️

PR #32, opened 2019-06-25
Initialise the DB

mafonso:
### Context

Run DB Migrations on startup

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -1,0 +1,6 @@
+#! /bin/sh
+set -e
+
+bundle exec rake db:create
+bundle exec rake db:migrate
+bundle exec rails server -b 0.0.0.0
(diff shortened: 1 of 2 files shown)


startup.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/sh
+set -e
+
+bundle exec rake db:create
```
smoothcontract: I don't think this should run `db:create` every time the container starts - assumption would be that we create an initial empty database as a one-off activity.

mafonso: I checked with @tavy88 that is was idempotent.  Trying to avoid one-off tasks as much as possible.

smoothcontract: Okay then, in that case I'm happy.

cwrw: I agree with @smoothcontract we have to be very explicit on `db:create` and ideally should not include that in the startup script

PR #33, opened 2019-06-26
Build skills page

tavy88:
### Context

Direct to job skills page from 'explore my skills' results page.

### Ticket

https://dfedigital.atlassian.net/browse/GET-31


app/views/job_profiles/skills.html.erb
```diff
@@ -0,0 +1,25 @@
...
+        [t('breadcrumb.search_results'), results_check_your_skills_path]
+      ]
+    )
+  end
+%>
+
+<div class="govuk-grid-row govuk-!-margin-top-7 govuk-body">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-xl">Your current skills</span>
+    <%= render partial: 'brief' %>
```
cwrw: we don't need to include the word partial here but use `<%= render 'brief" %>`

app/views/job_profiles/skills.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <%= @job_profile.skills_list %>
+    <div class="govuk-grid-column-full">
+      <%= link_to('Explore jobs you could do', explore_occupations_path, class: 'govuk-button govuk-!-margin-top-6  govuk-!-margin-bottom-9') %>
+    </div>
+  </div>
+
+  <%= render partial: 'shared/contact_us' %>
```
cwrw: same here

app/controllers/job_profiles_controller.rb
```diff
@@ -1,11 +1,13 @@
 class JobProfilesController < ApplicationController
-  def show
-    @job_profile = JobProfileDecorator.new(job_profile)
-  end
+  before_action :decorate_job_profile, on: %i[show skills]
```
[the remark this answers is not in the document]
tavy88: I'm just trying to reduce code redundancy. What's the harm in drying it like this? Surely we don't want 2 methods that have the exact same code.

app/views/job_profiles/skills.html.erb
tavy88: Ok, it's just a preference thing.

app/views/job_profiles/skills.html.erb
tavy88: Ok, it's just a preference thing.

app/controllers/job_profiles_controller.rb
```diff
@@ -1,11 +1,13 @@
...
 
   private
 
   def job_profile
     @job_profile ||= JobProfile.find_by(slug: params[:id])
   end
+
+  def decorate_job_profile
```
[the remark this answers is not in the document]
tavy88: Yeah, we discussed this before. I will try to name @job_profile to @job_profile_decorated.

app/controllers/job_profiles_controller.rb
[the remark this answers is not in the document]
cwrw: You can dry up the methods, but you should avoid using `before_action` to do that as that's not what it's intended for

app/views/shared/search/_results_list.html.erb
```diff
@@ -2,7 +2,7 @@
   <% job_profiles.each do |job_profile| %>
     <li class="govuk-!-padding-bottom-2 govuk-!-padding-top-4">
       <h3 class="govuk-!-margin-0">
-        <%= link_to job_profile.name, job_profile_path(job_profile.slug), class: 'govuk-link no-text-decoration' %>
+        <%= link_to job_profile.name, show_skills_only ? skills_job_profile_path(job_profile.slug) : job_profile_path(job_profile.slug), class: 'govuk-link no-text-decoration' %>
```
cwrw: this fails if you click through to a category example :`http://localhost:3000/categories/1`

cwrw: It's because this partial is being reused in listing category job profiles, and needs a `show_skills_only?` variable as well

tavy88: No, it shouldn't. That is optional. I will check. See this is an example of why we should have maybe controller tests to just check if the page renders without error.

app/controllers/job_profiles_controller.rb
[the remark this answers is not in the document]
tavy88: Is there an article/document that backs your statement? I would be curious to see it. Also, how would you alternatively dry this?

app/views/shared/search/_results_list.html.erb
tavy88: I might just move it to its own controller (skills_controller) - using custom methods inside Rails controllers is considered bad practice usually anyway. (might try to stick to CRUD)

app/controllers/job_profiles_controller.rb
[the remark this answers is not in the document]
tavy88: I might just move it to its own controller (skills_controller) - using custom methods inside Rails controllers is considered bad practice usually anyway. (might try to stick to CRUD)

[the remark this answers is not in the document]
cwrw: I would have done something similar to this:
```
  def show
    @job_profile = decorate_job_profile
  end

  def skills
    @job_profile = decorate_job_profile
  end

  private

  def decorate_job_profile
    JobProfileDecorator.new(
      JobProfile.find_by(slug: params[:id])
    )
  end
end
``` 
But yes as you suggested in the last comment, since we already have a skills model, it makes more sense to use the controller for it, and next time we pick up a scraper ticket we can populate the skills table directly in scraping itself and move that out of the decorator.

app/views/shared/search/_results_list.html.erb
cwrw: The error is:
```
ActionView::Template::Error (undefined local variable or method `show_skills_only')
```
which is being called from: https://github.com/DFE-Digital/get-help-to-retrain/blob/master/app/views/categories/show.html.erb#L16 

We are certainly missing test coverage, but as discussed earlier, this should be reflected in the cucumber tests, as that is the direction we have chosen to test our user journey. We seem to be missing specifically tests for the category page as that ticket has been completed and moved to done. I think it's worth discussing this today to see how we can rectify that

cwrw: I have found the test here: /Users/abeersalameh/workspace/dfe/get-help-to-retrain/features/62-Job-Categories.feature

It seems it is tagged not to run normally. I'll talk to Paul and decide which tags are to run normally, as currently it only runs CI tags,

app/controllers/job_profiles_controller.rb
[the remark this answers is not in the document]
smoothcontract: Skills are already parsed out separately when scraping, and persisted as skills against a job profile. So I wouldn't expect there to be any need to further decorate those. I'd agree this makes more sense as a skills controller index page, nested under job profile route.

app/controllers/job_profiles_controller.rb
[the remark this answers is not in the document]
smoothcontract: I don't think it makes sense to rename the instance variable, unless you need to work with both the decorated and undecorated version of the job profile in the view (which would be a bit weird). So better to leave the instance variable name as `@job_profile` and combine these two methods together as per the suggestion above.

app/views/shared/search/_results_list.html.erb
smoothcontract: I'm not averse to also adding request specs to demonstrate that views render correctly. Not much of a fan of view specs though, and controller specs are deemed deprecated.

config/routes.rb
```diff
@@ -12,7 +12,9 @@
     get :results, on: :collection
   end
 
-  resources :job_profiles, only: %i[show]
+  resources :job_profiles, only: %i[show] do
+    get :skills, on: :member
+  end
```
smoothcontract: This would make more sense as

```ruby
resources :job_profiles, only: %i[show] do
    resources :skills, only: %i[index]
  end
```

That should give the same url but in a more restful style.

spec/decorators/job_profile_decorator_spec.rb
```diff
@@ -1,10 +1,24 @@
...
+
+    let(:skill) { build_stubbed(:skill, name: 'test skill') }
+
+    let(:expected_html) {
+      '<ul class="govuk-list govuk-list--bullet"><li>test skill</li></ul>'
+    }
+
+    it 'returns an unordered HTML list of skills' do
+      expect(job_profile.skills_list).to eq expected_html
+    end
```
smoothcontract: Again, it feels like overkill to use a decorator for something as simple as this. I'd prefer it to be moved to the view - even maybe as a partial for each skill bullet point then you can pass the skills collection to be rendered.

app/controllers/job_profiles_controller.rb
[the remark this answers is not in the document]
tavy88: Ok, I just wanted to keep both versions available in case we might need both.

app/decorators/job_profile_decorator.rb
```diff
@@ -73,6 +75,14 @@ def section(xpath: nil)
     @doc.to_html.concat(separator_line)
   end
 
+  def skills_list
```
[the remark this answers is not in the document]
tavy88: Is that not the purpose of decorators to move out logic from views? That's the reason I added it here as it does a bit of logic. Why do you feel this approach is wrong?

spec/decorators/job_profile_decorator_spec.rb
tavy88: I'm not convinced having this in a decorator method is wrong. It does a bit of logic to build that list and I feel the decorator is the right place for it.

tavy88: However, if you feel strongly against it I will remove it.

app/decorators/job_profile_decorator.rb
```diff
@@ -1,5 +1,7 @@
 class JobProfileDecorator < SimpleDelegator
   include ActionView::Helpers::TagHelper
+  include ActionView::Context
```
cwrw: ✂️

app/views/skills/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <span class="govuk-caption-xl">Your current skills</span>
+    <%= render 'job_profiles/brief' %>
+    <%= render 'skills_list' %>
+    <div class="govuk-grid-column-full">
+      <%= link_to('Explore jobs you could do', explore_occupations_path, class: 'govuk-button govuk-!-margin-top-6  govuk-!-margin-bottom-9') %>
+    </div>
+  </div>
+
+  <%= render 'shared/contact_us' %>
+</div>
```
cwrw: 🦂

app/decorators/job_profile_decorator.rb
tavy88: ha true true

app/controllers/job_profiles_controller.rb
```diff
@@ -1,11 +1,11 @@
...
-    @job_profile = JobProfileDecorator.new(job_profile)
+    @job_profile = JobProfileDecorator.new(resource)
   end
 
   private
 
-  def job_profile
-    @job_profile ||= JobProfile.find_by(slug: params[:id])
+  def resource
+    @resource ||= JobProfile.find_by(slug: params[:id])
```
smoothcontract: This instance variable isn't necessary - the method is only being called once so can be simplified to

```ruby
JobProfile.find_by(slug: params[:id])
```

app/controllers/skills_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+    @skills = job_profile.skills
+  end
+
+  private
+
+  def job_profile
+    @job_profile ||= JobProfileDecorator.new(
+      JobProfile.find_by(slug: params[:job_profile_id])
+    )
+  end
```
smoothcontract: Two different styles used in this controller and the one above for finding and decorating the job profile. Could you standardise on one approach maybe please? I think I prefer this one.

app/views/skills/_skills_list.html.erb
```diff
@@ -0,0 +1,5 @@
+<ul class='govuk-list govuk-list--bullet'>
+  <% @skills.each do |skill| %>
+    <li><%= skill.name %></li>
+  <% end %>
+</ul>
```
smoothcontract: Thankyou - I think this looks cleaner than it did in the decorator 👍

app/views/skills/index.html.erb
smoothcontract: @cwrw crab emoji for missing blank line? is that a thing?

app/decorators/job_profile_decorator.rb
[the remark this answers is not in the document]
smoothcontract: I don't think it counts as logic - it's simply showing the list of skills. ERB is a better implementation for that use case. If there was something more complex going on maybe some filtering etc. then I'd agree a decorator is the right thing to use.

app/views/skills/index.html.erb
cwrw: it is now :D

PR #38, opened 2019-06-26
GET-161 Change copy on job profile button

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-161


cwrw:
I initially wanted to go with that. But in the end we agreed we shouldn't be using locales for everything and mainly in places which are reused like title, breadcrumbs or partials. For instance this text will only be used here so it's ok

PR #42, opened 2019-06-27
GET-141 scrape salary range

smoothcontract:
### Context
Salary range was previously rendered on the fly using Nokogiri to parse the job profile body content. This was too slow when viewing a large number of job profiles on the search results page.

Added new attributes to job profile for salary min and max values. These are populated when scraping the job profile, and then used within the decorator. Also cleaned up a little related/dead code.

As the salary values are held as integers, they need to be reformatted back to currency values. I've changed the default locale to `en-GB` to accomodate this, otherwise we would need to specify currency symbol every time. Open to other approaches though.

Search results page (for all 857 jobs) is now nice and quick:

```
  Rendering explore_occupations/results.html.erb within layouts/application
  Rendered shared/search/_results_form.html.erb (0.8ms)
  Rendered shared/_contact_us.html.erb (0.5ms)
  Rendered explore_occupations/results.html.erb within layouts/application (408.6ms)
  Rendered shared/navigation/_main_header.html.erb (0.7ms)
Completed 200 OK in 755ms (Views: 429.6ms | ActiveRecord: 124.8ms)
```

### Guidance to review
Job profiles need to be re-scraped by running `bundle exec rails data_import:scrape_job_profiles` - otherwise all job profiles will display `Varied` for salary range. It's not necessary to run the other scraper steps just for this.

### DIFF
--- a/app/decorators/job_profile_decorator.rb
+++ b/app/decorators/job_profile_decorator.rb
@@ -3,10 +3,9 @@
 # which should remove the need to disable rubocop rules here.
 class JobProfileDecorator < SimpleDelegator
   include ActionView::Helpers::TagHelper
+  include ActionView::Helpers::NumberHelper
 
   # rubocop:disable Metrics/LineLength
-  SALARY_MIN_XPATH = "//div[@id='Salary']//p[@class='dfc-code-jpsstarter']".freeze
-  SALARY_MAX_XPATH = "//div[@id='Salary']//p[@class='dfc-code-jpsexperienced']".freeze
   WORKING_HOURS_XPATH = "//div[@id='WorkingHours']//p[@class='dfc-code-jphours']".freeze
   WORKING_HOURS_PATTERNS_XPATH = "//div[@id='WorkingHoursPatterns']//p[@class='dfc-code-jpwpattern']".freeze
   HERO_COPY_XPATH = "//header[@class='job-profile-hero']//h1[@class='heading-xlarge']".freeze
@@ -25,13 +24,9 @@ class JobProfileDecorator < SimpleDelegator
   # rubocop:enable Metrics/LineLength
 
   def salary_range
-    min_salary = html_body.xpath(SALARY_MIN_XPATH).children[0]
+    return 'Variable' unless salary_min && salary_max
 
-    return 'Variable' unless min_salary
-
-    max_salary = html_body.xpath(SALARY_MAX_XPATH).children[0]
-
-    "#{min_salary.text.strip} - #{max_salary.text.strip}"
+    "#{number_to_currency(salary_min, precision: 0)} - #{number_to_currency(salary_max, precision: 0)}"
   end
 
   def working_hours
@@ -81,7 +76,7 @@ class JobProfileDecorator < SimpleDelegator
   private
 
   def html_body
-    @html_body ||= Nokogiri::HTML(__getobj__ .content)
+    @html_body ||= Nokogiri::HTML(content)
   end
 
   def mutate_html_body
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -18,37 +18,8 @@ class JobProfile < ApplicationRecord
     scraper = JobProfileScraper.new
     scraped = scraper.scrape(source_url)
 
-    update(name: scraped['title'], description: scraped['description'], content: scraped['body'])
-    self.skills = Skill.import(scraped['skills'])
-  end
-
-  def salary
-    {
-      min: html_body.xpath(SALARY_MIN_XPATH).children[0].text.strip,
-      max: html_body.xpath(SALARY_MAX_XPATH).children[0].text.strip
-    }
-  end
-
-  def working_hours
-    html_body.xpath(WORKING_HOURS_XPATH)
-             .children[0]
-             .text
-             .strip
-             .gsub('to', '-')
-             .delete(' ')
-  end
-
-  def working_hours_patterns
-    html_body.xpath(WORKING_HOURS_PATTERNS_XPATH)
-             .children[0]
-             .text
-             .strip
-             .capitalize
-  end
-
-  private
-
-  def html_body
-    @html_body ||= Nokogiri::HTML(content)
+    self.name = scraped.delete('title')
+    self.skills = Skill.import scraped.delete('skills')
+    update!(scraped)
   end
 end
--- a/app/scrapers/job_profile_scraper.rb
+++ b/app/scrapers/job_profile_scraper.rb
@@ -3,6 +3,15 @@ class JobProfileScraper
 
   title css: 'h1.heading-xlarge'
   description css: '.column-desktop-two-thirds p'
-  body 'css=body', :html
+
+  salary_min css: '#Salary p.dfc-code-jpsstarter/text()' do |min|
+    min.gsub(/\D/, '').to_i if min =~ /\d/
+  end
+
+  salary_max css: '#Salary p.dfc-code-jpsexperienced/text()' do |max|
+    max.gsub(/\D/, '').to_i if max =~ /\d/
+  end
+
+  content 'css=body', :html
   skills 'css=#Skills ul li', :list
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,5 +28,7 @@ module GetHelpToRetrain
     # the framework and any gems in your application.
 
     config.exceptions_app = routes
+
+    config.i18n.default_locale = :'en-GB'
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,4 +1,4 @@
-en:
+en-GB:
   breadcrumb:
     home: Home
     task_list: Get help to retrain
--- a/spec/scrapers/job_profile_scraper_spec.rb
+++ b/spec/scrapers/job_profile_scraper_spec.rb
@@ -21,8 +21,16 @@ RSpec.describe JobProfileScraper, vcr: { cassette_name: 'explore_my_careers_job_
       expect(scraped['description']).to match 'organising meetings'
     end
 
+    it 'parses job profile minimum salary' do
+      expect(scraped['salary_min']).to eq 14_000
+    end
+
+    it 'parses job profile maximum salary' do
+      expect(scraped['salary_max']).to eq 30_000
+    end
+
     it 'parses job profile full page' do
-      expect(scraped['body']).to match 'National Careers Service'
+      expect(scraped['content']).to match 'National Careers Service'
     end
 
     it 'parses skills' do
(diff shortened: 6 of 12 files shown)


app/models/job_profile.rb
```diff
@@ -18,37 +18,8 @@ def scrape
     scraper = JobProfileScraper.new
     scraped = scraper.scrape(source_url)
 
-    update(name: scraped['title'], description: scraped['description'], content: scraped['body'])
-    self.skills = Skill.import(scraped['skills'])
-  end
-
-  def salary
```
smoothcontract: This is all dead code - and breaks because the constants are not defined here. Looks like it was a remnant from introducing the decorator on a previous PR (where these methods can be found).

app/models/job_profile.rb
```diff
@@ -18,37 +18,8 @@ def scrape
...
-             .capitalize
-  end
-
-  private
-
-  def html_body
-    @html_body ||= Nokogiri::HTML(content)
+    self.name = scraped.delete('title')
+    self.skills = Skill.import scraped.delete('skills')
+    update!(scraped)
```
smoothcontract: It was getting a little tedious to specify each attribute here, so other than `title` (which can't be named `name` as it breaks Wombat), and `skills` which need to be handled differently this now just does a mass assignment of all scraped attributes.

app/decorators/job_profile_decorator.rb
```diff
@@ -76,7 +71,7 @@ def section(xpath: nil)
   private
 
   def html_body
-    @html_body ||= Nokogiri::HTML(__getobj__ .content)
+    @html_body ||= Nokogiri::HTML(content)
```
smoothcontract: Works just as well without the `__getobj__` stuff and cleanly delegates `content` to the underlying model.

app/scrapers/job_profile_scraper.rb
```diff
@@ -3,6 +3,15 @@ class JobProfileScraper
...
   title css: 'h1.heading-xlarge'
   description css: '.column-desktop-two-thirds p'
-  body 'css=body', :html
+
+  salary_min css: '#Salary p.dfc-code-jpsstarter/text()' do |min|
+    min.gsub(/\D/, '').to_i if min =~ /\d/
+  end
+
+  salary_max css: '#Salary p.dfc-code-jpsexperienced/text()' do |max|
+    max.gsub(/\D/, '').to_i if max =~ /\d/
```
smoothcontract: Not sure if there's a nicer way to implement this. I want a `nil` value returned if the salary is not present with some digits, or just the numeric value if it is.

config/locales/en.yml
```diff
@@ -1,4 +1,4 @@
-en:
+en-GB:
```
smoothcontract: 🤔

app/scrapers/job_profile_scraper.rb
```diff
@@ -3,6 +3,15 @@ class JobProfileScraper
...
+  end
+
+  content 'css=body', :html
```
smoothcontract: Renamed from `body` to match job profile attribute name

app/decorators/job_profile_decorator.rb
```diff
@@ -20,13 +19,9 @@ class JobProfileDecorator < SimpleDelegator
   OTHER_ROUTES_SECTION_XPATH = "//section[contains(@class, 'job-profile-subsection') and contains (@id, 'otherroutes')]".freeze
 
   def salary_range
-    min_salary = html_body.xpath(SALARY_MIN_XPATH).children[0]
+    return 'Variable' unless salary_min && salary_max
```
tavy88: No big deal but I think testing if only min exists should be enough? Probably won't make much performance difference.

app/decorators/job_profile_decorator.rb
```diff
@@ -1,8 +1,7 @@
 class JobProfileDecorator < SimpleDelegator
   include ActionView::Helpers::TagHelper
+  include ActionView::Helpers::NumberHelper
 
-  SALARY_MIN_XPATH = "//div[@id='Salary']//p[@class='dfc-code-jpsstarter']".freeze
-  SALARY_MAX_XPATH = "//div[@id='Salary']//p[@class='dfc-code-jpsexperienced']".freeze
   WORKING_HOURS_XPATH = "//div[@id='WorkingHours']//p[@class='dfc-code-jphours']".freeze
   WORKING_HOURS_PATTERNS_XPATH = "//div[@id='WorkingHoursPatterns']//p[@class='dfc-code-jpwpattern']".freeze
```
tavy88: I think you can remove these 2 as well now?

app/models/job_profile.rb
```diff
@@ -18,37 +18,8 @@ def scrape
...
-             .strip
-             .capitalize
-  end
-
-  private
-
-  def html_body
-    @html_body ||= Nokogiri::HTML(content)
+    self.name = scraped.delete('title')
+    self.skills = Skill.import scraped.delete('skills')
```
tavy88: Could we do `Skill.import scraped.delete('skills')` on a separate method? This line seems to be doing 2 things.

app/models/job_profile.rb
```diff
@@ -18,37 +18,8 @@ def scrape
...
-  end
-
-  def working_hours_patterns
-    html_body.xpath(WORKING_HOURS_PATTERNS_XPATH)
-             .children[0]
-             .text
-             .strip
-             .capitalize
-  end
-
```
tavy88: How are specs still passing if you are removing these methods? A bit confused... can't see them as attributes on the job_profile?

smoothcontract: This code is not called from anywhere else. I suspect you decided to move it to the decorator at some point and forgot to remove from the model.

app/decorators/job_profile_decorator.rb
smoothcontract: If max is nil then it still makes more sense to show Various I think. We’re scraping an external site so best not to bake in too many assumptions about when these values get populated.

app/decorators/job_profile_decorator.rb
smoothcontract: No, they’re still used in the decorator. I removed the working hours methods and other dead code from the model. The xpath for salary isn’t required now as that’s implemented in the job profile scraper.

app/models/job_profile.rb
tavy88: Ah sorry, you're right. I thought I was looking at the decorator class.

app/decorators/job_profile_decorator.rb
tavy88: Yep sorry I was misled by the removals in the job profile class.

app/decorators/job_profile_decorator.rb
tavy88: Yeah but I don't think you'll have min and ni max or vice versa. If you have variable the min will likely miss. And there's no point checking both upper and lower limits. But again it's just an optional thing.

config/application.rb
```diff
@@ -28,5 +28,7 @@ class Application < Rails::Application
     # the framework and any gems in your application.
 
     config.exceptions_app = routes
+
+    config.i18n.default_locale = :'en-GB'
```
cwrw: I'm surprised we need the gem for this, I thought it would come with Rails: https://guides.rubyonrails.org/i18n.html#setup-the-rails-application-for-internationalization
unless they don't have GB?

app/decorators/job_profile_decorator.rb
```diff
@@ -3,10 +3,9 @@
 # which should remove the need to disable rubocop rules here.
 class JobProfileDecorator < SimpleDelegator
   include ActionView::Helpers::TagHelper
+  include ActionView::Helpers::NumberHelper
```
cwrw: is this needed for precision?

app/scrapers/job_profile_scraper.rb
cwrw: Because we need `nil` I think this is ok

spec/scrapers/job_profile_scraper_spec.rb
```diff
@@ -21,8 +21,16 @@
       expect(scraped['description']).to match 'organising meetings'
     end
 
+    it 'parses job profile minimum salary' do
```
cwrw: can we add more tests to check for nil values, values with and without other characters?

config/locales/en.yml
cwrw: seems ok according to https://guides.rubyonrails.org/i18n.html#configure-the-i18n-module, what are your current thoughts?

app/decorators/job_profile_decorator.rb
smoothcontract: It's for `number_to_currency` helper

config/application.rb
smoothcontract: Annoyingly GB isn't included by default with Rails and is only included in this gem. `UNACCEPTABLE!`.

config/locales/en.yml
smoothcontract: It works fine, and we're targettng GB only so I'm happy with it.

app/decorators/job_profile_decorator.rb
smoothcontract: Proposing to leave this as is

app/models/job_profile.rb
smoothcontract: I don't think it would be any clearer TBH. Proposing to leave this as is.

spec/scrapers/job_profile_scraper_spec.rb
smoothcontract: I've added additional test cases for minimum salary; maximum salary is pretty much identical so just left the happy path test for that one.

spec/scrapers/job_profile_scraper_spec.rb
```diff
@@ -1,32 +1,87 @@
...
+  describe 'salary_max' do
+    it 'parses job profile maximum salary' do
+      expect(scraped['salary_max']).to eq 30_000
+    end
+  end
+
+  describe 'salary_min' do
+    let(:fake_page) { Mechanize::Page.new nil, nil, body, 200, scraper.mechanize }
+
+    around do |example|
```
cwrw: Can you explain what those steps are doing as Im not sure I follow?

smoothcontract: Wombat is a bit odd in parts. Here I’m building a mechanize page with defined body content rather than allowing it to be fetched from vcr cassette. 

The ‘metadata.page’ part is actually an assignment - method name is slightly weird as it forms part of the DSL.

So this swaps out the page for some known content, runs the spec and then resets to nil. The last step is crucial as page is lazily fetched otherwise and can pollute subsequent tests.

PR #45, opened 2019-06-27
GET-156 Business metrics (KPIs)

mafonso:
### Context

https://dfedigital.atlassian.net/browse/GET-156

This adds a javascript snippet to every page in `production` which collects data on every page impression served. It also creates a permanent user cookie and temporary session cookie to track repeat visits.

* Adding the gem and initialiser boilerplate for Azure application insights.
* Introduced page titles for each view to allow easier tracking of user journeys within app insights
* Use category slugs rather than id's for category paths so that these don't change after scraping new categories

### Guidance to review
This needs to be deployed to an Azure instance with a reasonable level of traffic to allow testing - there's also some lead time before data appears. I'd suggest we merge this and then evaluate how well it works with the updated page titles.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM ruby:2.6-alpine as assets
 
 ENV RAILS_ENV production
 ENV NODE_ENV production
-ENV BUILD_PACKAGES build-base nodejs yarn tzdata postgresql-dev
+ENV BUILD_PACKAGES build-base nodejs yarn tzdata postgresql-dev git
 
 # Update and install all of the required packages.
 # At the end, remove the apk cache
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,8 @@ gem 'foreman'
 # Canonical meta tag
 gem 'canonical-rails'
 
+gem 'application_insights'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ end
 ruby File.read('.ruby-version').chomp
 
 # To use Azure application insights
-gem 'application_insights', '~> 0.5.6'
+gem 'application_insights', github: 'microsoft/ApplicationInsights-Ruby', ref: 'a7429200'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.2.3'
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,9 @@ end
 
 ruby File.read('.ruby-version').chomp
 
+# To use Azure application insights
+gem 'application_insights', '~> 0.5.6'
+
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.2.3'
 gem 'rails-i18n', '~> 5.1'
@@ -32,8 +35,6 @@ gem 'foreman'
 
 # Canonical meta tag
 gem 'canonical-rails'
-
-gem 'application_insights'
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,8 @@
 module ApplicationHelper
+  def page_title(key)
+    content_for(:page_title, I18n.t(key, scope: 'page_titles'))
+  end
+
   def generate_breadcrumbs(current_page, previous_pages)
     safe_join(
       [
--- a/app/views/check_your_skills/results.html.erb
+++ b/app/views/check_your_skills/results.html.erb
@@ -1,4 +1,4 @@
-<% page_title :check_skills_results %>
+<% page_title :check_your_skills_results %>
 <% content_for :breadcrumb do
     generate_breadcrumbs(
       t('breadcrumb.search_results'),
--- a/app/views/check_your_skills/results.html.erb
+++ b/app/views/check_your_skills/results.html.erb
@@ -1,3 +1,4 @@
+<% page_title :check_skills_results %>
 <% content_for :breadcrumb do
     generate_breadcrumbs(
       t('breadcrumb.search_results'),
--- a/app/views/explore_occupations/results.html.erb
+++ b/app/views/explore_occupations/results.html.erb
@@ -1,4 +1,4 @@
-<% page_title :job_profile_results %>
+<% page_title :explore_occupations_results %>
 <% content_for :breadcrumb do
     generate_breadcrumbs(
       t('breadcrumb.search_results'),
--- a/app/views/explore_occupations/results.html.erb
+++ b/app/views/explore_occupations/results.html.erb
@@ -1,3 +1,4 @@
+<% page_title :job_profile_results %>
 <% content_for :breadcrumb do
     generate_breadcrumbs(
       t('breadcrumb.search_results'),
--- a/config/initializers/application_insights.rb
+++ b/config/initializers/application_insights.rb
@@ -1,4 +1,4 @@
-unless Rails.env.test?
+if Rails.env.production?
   Rails.application.configure do
     if (app_insights_key = ENV['APPINSIGHTS_INSTRUMENTATIONKEY']) && app_insights_key.present?
       # the optional extra params are buffer_size (= 500) and send_interval (= 60),
--- a/config/initializers/application_insights.rb
+++ b/config/initializers/application_insights.rb
@@ -1,11 +1,11 @@
 unless Rails.env.test?
   Rails.application.configure do
-    if (app_insights_key = ENV["APPINSIGHTS_INSTRUMENTATIONKEY"]) && app_insights_key.present?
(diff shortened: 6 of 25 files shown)


mafonso:
This branch was released to [Dev1](https://s108d01-dev-as.azurewebsites.net/) 
The insight data can be checked on the Azure portal [here](https://portal.azure.com/#@platform.education.gov.uk/resource/subscriptions/f35bd249-3028-48bd-9aab-d00e9d1a56b7/resourceGroups/s108d01-dev/providers/Microsoft.Insights/components/s108d01-dev-as/overview)

Gemfile
```diff
@@ -30,6 +30,8 @@ gem 'foreman'
 # Canonical meta tag
 gem 'canonical-rails'
 
+gem 'application_insights'
```
cwrw: can we add a version here?

config/initializers/application_insights.rb
```diff
@@ -0,0 +1,38 @@
+unless Rails.env.test?
```
cwrw: can we also add env.development as well as we don't want this running on development locally as well

mafonso: Yeah, this was taken from an example. As I understood, everything running in Azure will be on the ruby prod profile.
However the App insights is accessible from anywhere, not just Azure infra, so it you find it is useful to send metric and events there while developing, that is possible to do.

Gemfile
mafonso: We can, latest stable is 0.5.6.
The schools experience team is using an unreleased version.
https://github.com/DFE-Digital/schools-experience/commit/f1e8a0d062cf29131db28f3751b4b688d61d8294#diff-8b7db4d5cc4b8f6dc8feb7030baa2478

I think they have made contributions to the Gem project.

config/initializers/application_insights.rb
mafonso: Anyway, it only get's used if the `APPINSIGHTS_INSTRUMENTATIONKEY` is defined.

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ruby:2.6-alpine as assets
 
 ENV RAILS_ENV production
 ENV NODE_ENV production
-ENV BUILD_PACKAGES build-base nodejs yarn tzdata postgresql-dev
+ENV BUILD_PACKAGES build-base nodejs yarn tzdata postgresql-dev git
```
cwrw: Where do we need this?

Gemfile
```diff
@@ -5,6 +5,9 @@ end
 
 ruby File.read('.ruby-version').chomp
 
+# To use Azure application insights
+gem 'application_insights', github: 'microsoft/ApplicationInsights-Ruby', ref: 'a7429200'
```
cwrw: can we not use gem 'application_insights', '~> 0.5.6' ? Why do we need this specific reference?

config/initializers/application_insights.rb
cwrw: development environment is only used locally, I would exclude it as well as we don't need that running locally

Gemfile
mafonso: This is the reference to version 0.5.7. (not released yet) which should have support for the JavaScript tracking which we would also like to use.

There is an issue for another project in DfE asking for a release date. https://github.com/microsoft/ApplicationInsights-Ruby/issues/71

config/initializers/application_insights.rb
```diff
@@ -0,0 +1,38 @@
...
+    # # info here:
+    #
+    # # https://github.com/microsoft/ApplicationInsights-Ruby
+    # # https://github.com/microsoft/ApplicationInsights-Home/wiki#getting-an-application-insights-instrumentation-key
+    #
+    sender = ApplicationInsights::Channel::AsynchronousSender.new
+    queue = ApplicationInsights::Channel::AsynchronousQueue.new(sender)
+    channel = ApplicationInsights::Channel::TelemetryChannel.new(nil, queue)
+
+    TELEMETRY_CLIENT = ApplicationInsights::TelemetryClient.new(app_insights_key, channel).tap do |tc|
```
tavy88: Where is being used out of curiosity?

smoothcontract: @tavy88 good catch - we're not using this yet so have commented out. It's useful to keep as documentation/sample code for now; planning to use extra telemetry features in near future.

mafonso: We will need this for the custom metrics for the KPIs, won't we?  So for we are only capturing browser side page views and rack requests.
This seems very little to troubleshoot any issue with a particular request after go live.
We are not doing any server side tracing or using the custom events for logging. We should at the very list add tracing in key point of the app, as currently we cannot examine in detail a particular request.
me and @tavy88 did this exercise with some errored requests and we were a bit blind on what was going on.

config/locales/en.yml
```diff
@@ -11,6 +11,21 @@ en-GB:
     search_results: Search results
     job_profile: Job profile details
     your_skills: Your skills
+
+  page_titles:
+    default: Get help to retrain
```
mafonso: In what render situations this default will be used?

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,8 @@
 module ApplicationHelper
+  def page_title(key)
+    content_for(:page_title, I18n.t(key, scope: 'page_titles'))
```
cwrw: is there anything that required us in using `I18n.t()` over `t()`?

app/views/explore_occupations/results.html.erb
```diff
@@ -1,3 +1,4 @@
+<% page_title :job_profile_results %>
```
cwrw: do we want to name this `:explore_occupation_results` for consistency?

app/views/check_your_skills/results.html.erb
```diff
@@ -1,3 +1,4 @@
+<% page_title :check_skills_results %>
```
cwrw: could we name this `:check_your_skills_results` for consistency?

config/initializers/application_insights.rb
```diff
@@ -0,0 +1,38 @@
...
+    if (app_insights_key = ENV['APPINSIGHTS_INSTRUMENTATIONKEY']) && app_insights_key.present?
+      # the optional extra params are buffer_size (= 500) and send_interval (= 60),
+      # leaving for now as they appear sensible
+      config.middleware.use(ApplicationInsights::Rack::TrackRequest, app_insights_key)
+      if ENV['APPINSIGHTS_JAVASCRIPT_ENABLED'] == 'true'
+        config.middleware.use(ApplicationInsights::Rack::InjectJavaScriptTracking, app_insights_key)
+      end
+    end
+
+    # # This isn't needed _until_ we want to track events and metrics in addition
```
cwrw: can we remove the unused setup?

config/locales/en.yml
```diff
@@ -11,6 +11,21 @@ en-GB:
...
+    home: Get help to retrain - Home
+    task_list: Get help to retrain - Tasks
+    check_your_skills: Get help to retrain - Check your skills
+    check_skills_results: Get help to retrain - Check skills results
+    your_skills: Get help to retrain - Your skills
+    explore_occupations: Get help to retrain - Explore occupations
+    job_category: Get help to retrain - Job category
+    job_profile_results: Get help to retrain - Search results
```
cwrw: could we change this to `Get help to retrain - Explore occupations results`

cwrw:
looks good! Just come consistency comments but otherwise exciting 🎈

config/initializers/application_insights.rb
smoothcontract: @mafonso yes, we probably will - but I want to see how much we get out of the box first. It's not clear yet what the KPI requirements are so will keep this around for documentation only (it wasn't being called so is dead code at present) and reintroduce it when needed.

config/locales/en.yml
smoothcontract: This gets used if we forget to add a custom page title to any of the views.

app/helpers/application_helper.rb
smoothcontract: yep, `t()` doesn't work in a helper, only a view.

config/initializers/application_insights.rb
smoothcontract: Not yet, I'd prefer to keep it as documentation for the time being. It's likely we may need some of this for additional KPI's - to be explored in next sprint. If the out of the box features are sufficient without this then it can be removed. But lets see the insights working first.

config/locales/en.yml
```diff
@@ -11,6 +11,24 @@ en-GB:
     search_results: Search results
     job_profile: Job profile details
     your_skills: Your skills
+
+  page_titles:
+    default: Get help to retrain
+    home_index: Get help to retrain - Home
+    pages_task_list: Get help to retrain - Tasks
+    check_your_skills_index: Get help to retrain - Check your skills
+    check_your_skills_results: Get help to retrain - Check skills results
```
cwrw: could this be `Check your skills results`?

config/initializers/application_insights.rb
mafonso: Fair enough and I'm happy to go with this first.  Just pointing that we will need it for the log at least, as we should had the app tracing (and maybe dependency) calls as well as the very least before go live. These last ones are automatic with other AppInsights library, just not on the Ruby one (yet).

PR #46, opened 2019-06-27
GET-136 rubocop

smoothcontract:
### Context
Fixes rubocop violations and adds a step to Azure pipeline so that we can use rubocop locally. The gov.uk linter seems to ignore a lot of violations so this is more thorough.

Ignored class length and line length violations in JobProfileDecorator for now - that will be refactored over time to resolve those.

### Guidance to review
Check build passes - examine Azure devops output.

```
##[section]Starting: Rubocop linter
==============================================================================
Task         : Command line
Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
Version      : 2.151.1
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
==============================================================================
Generating script.
Script contents:
$DOCKER_RUN bundle exec rubocop
========================== Starting Command Output ===========================
[command]/bin/bash --noprofile --norc /home/vsts/work/_temp/4a75aa44-fa52-443d-b6b4-a3bc27ad6927.sh
Inspecting 72 files
........................................................................

72 files inspected, no offenses detected
##[section]Finishing: Rubocop linter
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
-git_source(:github) { |repo| "https://github.com/#{repo}.git" }
+git_source(:github) do |repo|
+  "https://github.com/#{repo}.git"
+end
 
 ruby File.read('.ruby-version').chomp
 
@@ -84,7 +86,7 @@ group :test do
   # Easy installation and use of browser drivers to run system tests with different browsers
   gem 'webdrivers', '~> 4.0'
 
-    # Test coverage reporting
+  # Test coverage reporting
   gem 'simplecov', '~> 0.16', require: false
 
   # Web request caching for tests
--- a/config/spring.rb
+++ b/config/spring.rb
@@ -3,4 +3,6 @@
   .rbenv-vars
   tmp/restart.txt
   tmp/caching-dev.txt
-].each { |path| Spring.watch(path) }
+].each do |path|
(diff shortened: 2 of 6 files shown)


Gemfile
```diff
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
-git_source(:github) { |repo| "https://github.com/#{repo}.git" }
+git_source(:github) do |repo|
+  "https://github.com/#{repo}.git"
+end
```
smoothcontract: Maybe the rubocop rules are still not quite right here. The previous inline block format seems fine to me. Rubocop auto corrected it to add `do` and `end` on the same line (D'oh) so I edited it. Rubocop is happy with this now anyway.

config/spring.rb
```diff
@@ -3,4 +3,6 @@
   .rbenv-vars
   tmp/restart.txt
   tmp/caching-dev.txt
-].each { |path| Spring.watch(path) }
+].each do |path|
+  Spring.watch(path)
+end
```
smoothcontract: Same block issue here as per `Gemfile`

tavy88: You should be allowed to use `{}` as long as the block is on just one line like it originally was.... `Do` is recommended if the block has multiple lines.

PR #47, opened 2019-06-28
Get 138 fix breadcrumbs

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-138

* users don't want to go back home but to tasklist from all pages
* keep home on task list breadcrumbs
* remove title of job and only have job details instead
* Move results list to specific directories to avoid confusion
  as now we are being explicit with passing params around
* Change breadcrumbs helper to rely on params rather than referrer
  request as that was flakey and not explicit causing unexpected
  behaviour from outside requests and if someone presses the back
  button on the browser

Breadcrumbs paths we support now (this is not the exact content but just to show the paths):
1) Homepage: none
2) Task list page: Home > Task list
3) Check your existing skills page: Home: Task list > Check your skills
4) Check your existing skills results page: Home: Task list > Check your skills > Search results
5) Skills page: Home: Task list > Check your skills > Search results > Skills (if you go back should now reserve your results)
6) Explore occupations page: Home: Task list > Explore Occupations
7) Categoires page: Home: Task list > Explore Occupations > Category
8) Job profile through category: Home: Task list > Explore Occupations > Category > Job profile
9) Explore occupations results page: Home: Task list > Explore Occupations > Search results
10) Job profile through results: Home: Task list > Explore Occupations >  Search results > Job profile (if you go back should now reserve your results)
11) apply for course: Home: Task list > Apply for course
12) Next steps: Home: Get help to retrain > Next steps

### DIFF
--- a/app/helpers/job_profiles_helper.rb
+++ b/app/helpers/job_profiles_helper.rb
@@ -1,5 +1,11 @@
 module JobProfilesHelper
-  def job_profile_breadcrumbs_for(url)
-    URI(url).path =~ /categories/ ? t('breadcrumb.job_category') : t('breadcrumb.search_results')
+  def job_profile_breadcrumbs_for(params)
+    if params[:category]
+      [t('breadcrumb.job_category'), category_path(params[:category])]
+    else
+      [t('breadcrumb.search_results'), results_explore_occupations_path(
+        name: params[:explore_occupations_result]
+      )]
+    end
   end
 end
--- a/app/views/categories/show.html.erb
+++ b/app/views/categories/show.html.erb
@@ -2,8 +2,7 @@
     generate_breadcrumbs(
       t('breadcrumb.job_category'),
       [
-        [t('breadcrumb.home'), root_path],
-        [t('breadcrumb.task_list'), task_list_path],
+        [t('breadcrumb.task_list_home'), task_list_path],
         [t('breadcrumb.explore_occupations'), explore_occupations_path]
       ]
     )
--- a/app/views/categories/show.html.erb
+++ b/app/views/categories/show.html.erb
@@ -12,7 +12,7 @@
 <div class="govuk-grid-row govuk-!-margin-top-7">
   <div class="govuk-grid-column-two-thirds" >
     <h3 class="govuk-heading-l"><%= @category.name %></h3>
-    <%= render "shared/search/results_list", job_profiles: @category.job_profiles %>
+    <%= render "results_list", job_profiles: @category.job_profiles %>
   </div>
   <%= render "shared/contact_us" %>
   <div class="govuk-grid-column-one-third">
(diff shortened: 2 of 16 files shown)


tavy88:
@cwrw Neither the ticket nor the description of this PR illustrates what the expected breadcrumbs are expected to be. Should we follow the prototype ones?

app/helpers/job_profiles_helper.rb
```diff
@@ -1,5 +1,11 @@
...
-  def job_profile_breadcrumbs_for(url)
-    URI(url).path =~ /categories/ ? t('breadcrumb.job_category') : t('breadcrumb.search_results')
+  def job_profile_breadcrumbs_for(params)
+    if params[:category]
+      [t('breadcrumb.job_category'), category_path(params[:category])]
+    else
+      [t('breadcrumb.search_results'), results_explore_occupations_path(
+        name: params[:explore_occupations_result]
+      )]
+    end
```
tavy88: Could we do 

```ruby

def job_profile_breadcrumbs_for(params)
  if params[:category]
     breadcrumb_entry(
        t('breadcrumb.job_category'),
        category_path(params[:category])
     )
  else
     breadcrumb_entry(
        t('breadcrumb.search_results'),
        results_explore_occupations_path(name: params[:explore_occupations_result]
     )
  end

def breadcrumb_entry(label, path)
  [label, path]
end
```

tavy88: Feels a bit more readable this way.

app/views/categories/show.html.erb
```diff
@@ -2,8 +2,7 @@
     generate_breadcrumbs(
       t('breadcrumb.job_category'),
       [
-        [t('breadcrumb.home'), root_path],
-        [t('breadcrumb.task_list'), task_list_path],
+        [t('breadcrumb.task_list_home'), task_list_path],
```
tavy88: And then you can do here:

```ruby
breadcrumb_entry(t('breadcrumb.task_list_home'), task_list_path) etc
```

app/helpers/job_profiles_helper.rb
smoothcontract: @tavy88 I disagree I'm afraid - that's almost 3 x more code and requires lots more mental jogging to piece it all together.

cwrw: I think currently as the method stand, its doing a simple if/else which I find quite easier to follow, so proposition is to keep as is

app/views/categories/show.html.erb
cwrw: I think pushing everything to a method to add it to an array is not needed at this stage

cwrw:
@tavy88 In terms of the specs I'll update the pr summary to include the specific differences

app/helpers/job_profiles_helper.rb
tavy88: Single responsibility principle is broken here because your method does the following:

- evaluates an expression
- returns array
- defines the structure of the returned array

My approach was breaking that responsibility to multiple methods that only do 1 thing. @smoothcontract @cwrw. However, there are 2 votes against my suggestion, so go for it.

PR #51, opened 2019-07-01
Display related careers on skills/job profile pages

tavy88:
### Context

As a user
I want to see occupations related to mine
So that I am made aware about other potential occupations

### Screenshots

Screen Shot 2019-07-02 at 14 54 29
Screen Shot 2019-07-02 at 14 54 45

### DEV NOTES

```ruby
irb(main):002:0> JobProfile.second.related_job_profiles
  JobProfile Load (0.7ms)  SELECT  "job_profiles".* FROM "job_profiles" ORDER BY "job_profiles"."id" ASC LIMIT $1 OFFSET $2  [["LIMIT", 1], ["OFFSET", 1]]
  JobProfile Load (0.6ms)  SELECT  "job_profiles".* FROM "job_profiles" INNER JOIN "related_job_profiles" ON "job_profiles"."id" = "related_job_profiles"."related_job_profile_id" WHERE "related_job_profiles"."job_profile_id" = $1 LIMIT $2  [["job_profile_id", 1816], ["LIMIT", 11]]
=> #<ActiveRecord::Associations::CollectionProxy []>
irb(main):003:0> JobProfile.second.related_job_profiles << JobProfile.all[5]
  JobProfile Load (0.7ms)  SELECT  "job_profiles".* FROM "job_profiles" ORDER BY "job_profiles"."id" ASC LIMIT $1 OFFSET $2  [["LIMIT", 1], ["OFFSET", 1]]
  JobProfile Load (81.5ms)  SELECT "job_profiles".* FROM "job_profiles"
   (0.2ms)  BEGIN
  JobProfile::HABTM_RelatedJobProfiles Create (0.3ms)  INSERT INTO "related_job_profiles" ("job_profile_id", "related_job_profile_id") VALUES ($1, $2)  [["job_profile_id", 1816], ["related_job_profile_id", 1824]]
   (5.8ms)  COMMIT
  JobProfile Load (0.9ms)  SELECT  "job_profiles".* FROM "job_profiles" INNER JOIN "related_job_profiles" ON "job_profiles"."id" = "related_job_profiles"."related_job_profile_id" WHERE "related_job_profiles"."job_profile_id" = $1 LIMIT $2  [["job_profile_id", 1816], ["LIMIT", 11]]
=> #<ActiveRecord::Associations::CollectionProxy [#<JobProfile id: 1824, slug: "advertising-account-planner", name: "Advertising account planner", source_url: "https://nationalcareers.service.gov.uk/job-profile...", description: "Advertising account planners develop and plan adve...", recommended: false, content: "<!-- Google Tag Manager (noscript) --> <noscript> ...", created_at: "2019-06-24 15:09:41", updated_at: "2019-06-27 11:19:31", salary_min: 18000, salary_max: 45000>]>
irb(main):004:0> JobProfile.second.related_job_profiles
  JobProfile Load (0.6ms)  SELECT  "job_profiles".* FROM "job_profiles" ORDER BY "job_profiles"."id" ASC LIMIT $1 OFFSET $2  [["LIMIT", 1], ["OFFSET", 1]]
  JobProfile Load (1.0ms)  SELECT  "job_profiles".* FROM "job_profiles" INNER JOIN "related_job_profiles" ON "job_profiles"."id" = "related_job_profiles"."related_job_profile_id" WHERE "related_job_profiles"."job_profile_id" = $1 LIMIT $2  [["job_profile_id", 1816], ["LIMIT", 11]]
=> #<ActiveRecord::Associations::CollectionProxy [#<JobProfile id: 1824, slug: "advertising-account-planner", name: "Advertising account planner", source_url: "https://nationalcareers.service.gov.uk/job-profile...", description: "Advertising account planners develop and plan adve...", recommended: false, content: "<!-- Google Tag Manager (noscript) --> <noscript> ...", created_at: "2019-06-24 15:09:41", updated_at: "2019-06-27 11:19:31", salary_min: 18000, salary_max: 45000>]>
```

Screen Shot 2019-07-01 at 11 49 38
Screen Shot 2019-07-02 at 14 54 45

Needs re-scraping of the job profiles.

### JIRA

https://dfedigital.atlassian.net/browse/GET-153

### DIFF
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -22,16 +22,12 @@ class JobProfile < ApplicationRecord
     end
   end
 
-  def self.bulk_import(slugs)
-    where(slug: slugs)
-  end
-
   def scrape(scraper = JobProfileScraper.new)
     scraped = scraper.scrape(source_url)
 
     self.name = scraped.delete('title')
     self.skills = Skill.import scraped.delete('skills')
-    self.related_job_profiles = JobProfile.bulk_import(scraped.delete('related_profiles') - [slug])
+    self.related_job_profiles = JobProfile.where(slug: scraped.delete('related_profiles') - [slug])
 
     update!(scraped)
   end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -22,11 +22,17 @@ class JobProfile < ApplicationRecord
     end
   end
 
+  def self.bulk_import(slugs)
+    slugs.map { |slug| find_or_create_by(slug: slug) }
+  end
+
   def scrape(scraper = JobProfileScraper.new)
     scraped = scraper.scrape(source_url)
 
     self.name = scraped.delete('title')
     self.skills = Skill.import scraped.delete('skills')
+    self.related_job_profiles = JobProfile.bulk_import scraped.delete('related_profiles')
+
     update!(scraped)
   end
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -23,7 +23,7 @@ class JobProfile < ApplicationRecord
   end
 
   def self.bulk_import(slugs)
-    slugs.map { |slug| find_by(slug: slug) }.compact
+    where(slug: slugs)
   end
 
   def scrape(scraper = JobProfileScraper.new)
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -68,8 +68,6 @@ class JobProfile < ApplicationRecord
       )
     )
 
-    clean_related_profiles
-
     self.related_job_profiles = imported_related_job_profiles
   end
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -4,6 +4,8 @@ class JobProfile < ApplicationRecord
   has_many :categories, through: :job_profile_categories, inverse_of: :job_profiles
   has_many :skills, through: :job_profile_skills, inverse_of: :job_profiles
 
+  attr_accessor :scraped
+
   has_and_belongs_to_many :related_job_profiles,
                           class_name: 'JobProfile',
                           join_table: :related_job_profiles,
@@ -27,24 +29,13 @@ class JobProfile < ApplicationRecord
   end
 
   def scrape(scraper = JobProfileScraper.new)
-    scraped = scraper.scrape(source_url)
+    self.scraped = scraper.scrape(source_url)
 
     ActiveRecord::Base.transaction do
       self.name = scraped.delete('title')
 
-      skills_to_import = Skill.import scraped.delete('skills')
-
-      related_profiles_to_import = JobProfile.bulk_import(
-        remove_value_from(
-          collection: scraped.delete('related_profiles'),
-          value: slug
-        )
-      )
-
-      clean_related_profiles
-
-      self.skills = skills_to_import
-      self.related_job_profiles = related_profiles_to_import
+      update_skills
+      update_related_jobs
 
       update!(scraped)
     rescue StandardError => e
@@ -62,5 +53,23 @@ class JobProfile < ApplicationRecord
 
   def clean_related_profiles
     related_job_profiles.destroy_all
+  end
+
+  def update_skills
+    imported_skills = Skill.import(scraped.delete('skills'))
+    self.skills = imported_skills
+  end
+
+  def update_related_jobs
+    imported_related_job_profiles = JobProfile.bulk_import(
+      remove_value_from(
+        collection: scraped.delete('related_profiles'),
+        value: slug
+      )
+    )
+
+    clean_related_profiles
+
+    self.related_job_profiles = imported_related_job_profiles
   end
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -23,7 +23,7 @@ class JobProfile < ApplicationRecord
   end
 
   def self.bulk_import(slugs)
-    slugs.map { |slug| find_or_create_by(slug: slug) }
+    slugs.map { |slug| find_by(slug: slug) }.compact
   end
 
   def scrape(scraper = JobProfileScraper.new)
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -29,15 +29,27 @@ class JobProfile < ApplicationRecord
   def scrape(scraper = JobProfileScraper.new)
     scraped = scraper.scrape(source_url)
 
-    self.name = scraped.delete('title')
+    ActiveRecord::Base.transaction do
+      self.name = scraped.delete('title')
 
-    self.skills = Skill.import scraped.delete('skills')
+      skills_to_import = Skill.import scraped.delete('skills')
 
-    scraped_profile_slugs = scraped.delete('related_profiles')
+      related_profiles_to_import = JobProfile.bulk_import(
+        remove_value_from(
+          collection: scraped.delete('related_profiles'),
+          value: slug
+        )
+      )
 
-    self.related_job_profiles = JobProfile.bulk_import remove_value_from(collection: scraped_profile_slugs, value: slug)
+      clean_related_profiles
 
-    update!(scraped)
+      self.skills = skills_to_import
+      self.related_job_profiles = related_profiles_to_import
+
+      update!(scraped)
+    rescue StandardError => e
+      puts "Could not resume scraping, rolling back. Reason: #{e.message}"
+    end
   end
 
   private
@@ -46,5 +58,9 @@ class JobProfile < ApplicationRecord
     collection.delete(value) if value
 
     collection
+  end
+
+  def clean_related_profiles
+    related_job_profiles.destroy_all
   end
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -30,9 +30,21 @@ class JobProfile < ApplicationRecord
     scraped = scraper.scrape(source_url)
 
     self.name = scraped.delete('title')
+
     self.skills = Skill.import scraped.delete('skills')
-    self.related_job_profiles = JobProfile.bulk_import scraped.delete('related_profiles')
+
+    scraped_profile_slugs = scraped.delete('related_profiles')
+
+    self.related_job_profiles = JobProfile.bulk_import remove_value_from(collection: scraped_profile_slugs, value: slug)
 
     update!(scraped)
+  end
+
+  private
+
+  def remove_value_from(collection:, value:)
+    collection.delete(value) if value
+
+    collection
   end
 end
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -4,8 +4,6 @@ class JobProfile < ApplicationRecord
   has_many :categories, through: :job_profile_categories, inverse_of: :job_profiles
   has_many :skills, through: :job_profile_skills, inverse_of: :job_profiles
 
-  attr_accessor :scraped
-
   has_and_belongs_to_many :related_job_profiles,
                           class_name: 'JobProfile',
                           join_table: :related_job_profiles,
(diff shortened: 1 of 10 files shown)


app/webpacker/styles/_job-profile.scss
```diff
@@ -101,3 +101,9 @@
     width: 150% !important;
   }
 }
+
+.related-profiles {
+  @include govuk-media-query($from: tablet) {
+    margin-top: 180px;
```
cwrw: can we use the custom govuk margins instead?

tavy88: How it goes up to 60px? https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale

spec/models/job_profile_spec.rb
```diff
@@ -126,30 +134,46 @@
...
+        create :job_profile, slug: 'data-entry-clerk'
+
+        job_profile.scrape
+      end
+
+      it 'updates the related_profiles with scraped related job profiles data' do
+        expect(job_profile.related_job_profiles.pluck(:slug)).to contain_exactly(
+          'data-entry-clerk'
+        )
+      end
```
smoothcontract: Could you explain the purpose of this test please? Why does this need to differentiate between being previously scraped or not? Seems simpler to just have the one context as before.

smoothcontract: On reading this again, I can see that an existing job profile record is needed to setup the related job profiles correctly. In that case I think it would make more sense to structure the tests a little differently to make this more explicit - also it doesn't matter if the existing related job has been scraped previously, so how about this:

```ruby
describe '#scrape', vcr: { cassette_name: 'explore_my_careers_job_profile' } do
  let(:url) { 'https://nationalcareers.service.gov.uk/job-profiles/admin-assistant' }
  let(:job_profile) { build :job_profile, source_url: url }
  let!(:customer_service) { create :skill, name: 'customer service skills' }

  before { job_profile.scrape }

  it 'updates name with scraped title' do
    expect(job_profile.name).to eq 'Admin assistant'
  end

  it 'updates description with scraped description' do
    expect(job_profile.description).to match 'organising meetings'
  end

  it 'updates salary minimum with scraped salary' do
    expect(job_profile.salary_min).to eq 14_000
  end

  it 'updates salary maximum with scraped salary' do
    expect(job_profile.salary_max).to eq 30_000
  end

  it 'updates content with scraped body' do
    expect(job_profile.content).to match 'National Careers Service'
  end

  it 'updates skills with scraped skill names' do
    expect(job_profile.skills).to include customer_service
  end

  it 'updates the related_profiles with matching scraped related job profiles' do
    create :job_profile, slug: 'data-entry-clerk' 
    create :job_profile, slug: 'unrelated-job-profile'
    expect(job_profile.related_job_profiles.pluck(:slug)).to contain_exactly(
      'data-entry-clerk'
    )
  end
end
```

tavy88: This won't work unfortunately as the scrape will run before I have seeded the data. That's the reason I separated it into different contexts.

smoothcontract: If that's the case then I think there's a bug as that would suggest the order in which job profiles are scraped can affect the results.

tavy88: Sorry, @smoothcontract  I am talking about the way the before block works in the specs. As you've suggested the before block kick in first, and then the create factories will kick in.

smoothcontract: The process of scraping is in 3 steps:

1. create placeholder job profiles and categories for each of those found in the site map.
2. scrape category pages to link job profiles to categories
3. scrape job profile pages to populate all remaining job profile attributes and build skills as needed

At present, those steps are idempotent and will work irrespective of the order in which we find the data. I would expect us to be able to attach a job profile which is being scraped currently to an existing job profile record that has not yet been scraped. There's an easy way to check that - just run the test I suggested :)

tavy88: Let's discuss - the test will fail - I have done it like this previously. If you check bulk_import it does not do a find_or_create it just does find_by.

smoothcontract: Discussed offline and separate contexts make sense to allow another job profile to be persisted before calling `scrape` method.

app/models/job_profile.rb
```diff
@@ -14,11 +20,45 @@ def self.import(slug, url)
...
+      skills_to_import = Skill.import scraped.delete('skills')
+
+      related_profiles_to_import = JobProfile.bulk_import(
+        remove_value_from(
+          collection: scraped.delete('related_profiles'),
+          value: slug
+        )
+      )
+
+      clean_related_profiles
```
smoothcontract: I think this is not necessary. You're assigning to the `related_job_profiles` collection property in line 45 - that will automatically create and update the relevant related job profile entries I believe. Hence the transaction wrapper around here is not required either.

app/models/job_profile.rb
```diff
@@ -14,11 +20,45 @@ def self.import(slug, url)
...
 
-    self.name = scraped.delete('title')
-    self.skills = Skill.import scraped.delete('skills')
-    update!(scraped)
+    ActiveRecord::Base.transaction do
+      self.name = scraped.delete('title')
+
+      skills_to_import = Skill.import scraped.delete('skills')
+
+      related_profiles_to_import = JobProfile.bulk_import(
```
smoothcontract: I think this could be a little clearer without adding the extra method (seems a bit too verbose maybe). Perhaps something like:

```ruby
related_skills_slugs = scraped.delete('related_profiles').reject { |rs| slug == rs }
self.related_job_profiles = JobProfile.bulk_import related_skills_slugs
```

taking this a step further, `where` I believe accepts an array argument, so you could also remove the `bulk_import` method here and replace with:

```ruby
related_skills_slugs = scraped.delete('related_profiles').reject { |rs| slug == rs }
self.related_job_profiles = JobProfile.where(slug: related_skills_slugs)
```

Lines 33 and 44 can be recombined in a single line once all the transaction stuff is removed too.

tavy88: @smoothcontract I've done a bit of refactoring to fix violations.  Let's chat when you have a moment.

app/models/job_profile.rb
tavy88: What about the deprecated relations if they change something on their source page?

tavy88: Also re transaction wrapper, what if something fails e.g during the Import? Or one of the previous steps fail?

app/models/job_profile.rb
```diff
@@ -14,11 +20,17 @@ def self.import(slug, url)
...
+  def self.bulk_import(slugs)
+    slugs.map { |slug| find_by(slug: slug) }.compact
+  end
+
   def scrape(scraper = JobProfileScraper.new)
     scraped = scraper.scrape(source_url)
 
     self.name = scraped.delete('title')
     self.skills = Skill.import scraped.delete('skills')
+    self.related_job_profiles = JobProfile.bulk_import(scraped.delete('related_profiles') - [slug])
```
smoothcontract: This is why I love PR's - implementation has been reduced to a single LOC ❤️

app/models/job_profile.rb
```diff
@@ -14,11 +20,17 @@ def self.import(slug, url)
     end
   end
 
+  def self.bulk_import(slugs)
+    slugs.map { |slug| find_by(slug: slug) }.compact
```
smoothcontract: Could you try `where(slug: slugs)` please to see if that works instead (it's much simpler code if indeed it works).

app/models/job_profile.rb
smoothcontract: Discussed offline - it will break, but we expect that and will resolve the wider issue of making the entire scraping process atomic somehow.

app/models/job_profile.rb
tavy88: Yeah, agree I might have overreacted with my original implementation.

app/models/job_profile.rb
tavy88: Ah yeah haha good point. Not sure how I missed that.

app/webpacker/styles/_job-profile.scss
cwrw: Let me play around with this locally and see how we can get it to not rely on a fixed margin. Also another question around adding an extra blue bar like we do to the call us section (something to ask the designers)

cwrw: alright I was able to cobble up something: 
```
<div class="govuk-grid-row govuk-!-margin-top-7 govuk-body">
      <span class="govuk-caption-xl">Your current skills</span>
      <%= render 'job_profiles/brief' %>
    </div>
    <div class="govuk-grid-column-full govuk-!-padding-0 extended-container">
      <div class="govuk-grid-row">
          <%= render 'skills_list' %>
          <div class="govuk-grid-column-full">
            <%= link_to('Explore jobs you could do', explore_occupations_path, class: 'govuk-button govuk-!-margin-top-6  govuk-!-margin-bottom-9') %>
          </div>
        </div>
        <div class='govuk-grid-column-one-third govuk-!-padding-left-5'>
          <%= render 'job_profiles/related_profiles' %>
        </div>
      </div>
    </div>
  </div>

  <%= render 'shared/contact_us' %>
</div>
```
This does not use a margin but relies on the div, which isn't too much better and relies on the content being there, so maybe its safe to keep your solution for now. We need to ask the designers about the blue bar though. I think we can have a session with them as they are flexible regarding things

spec/models/job_profile_spec.rb
```diff
@@ -126,30 +134,49 @@
     let(:job_profile) { build :job_profile, source_url: url }
     let!(:customer_service) { create :skill, name: 'customer service skills' }
 
-    before { job_profile.scrape }
+    context 'with no persisted job profiles' do
```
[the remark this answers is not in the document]
tavy88: Yes because we need already scraped profiles when we're linking the related ones. Does it make sense?

app/webpacker/styles/_job-profile.scss
tavy88: I think it's a bit too late to block this. This should have been discussed prior. They have not specified any blue bar so we are not going to implement it in this iteration I would say.

tavy88: Also, I'm not sure why we are against using custom margins when we have to? I feel we should be a bit more flexible...

app/models/job_profile.rb
tavy88: There are cases when on the NCS site they link a job profile with self. I found an example. Hence why we remove the self.slug

tavy88: What is wrong if it's nil?

```ruby
irb(main):004:0> JobProfile.where(slug: nil)
  JobProfile Load (12.5ms)  SELECT  "job_profiles".* FROM "job_profiles" WHERE "job_profiles"."slug" IS NULL LIMIT $1  [["LIMIT", 11]]
=> #<ActiveRecord::Relation []>
```

spec/scrapers/job_profile_scraper_spec.rb
```diff
@@ -30,6 +30,18 @@
...
+  describe 'related profiles' do
+    it 'parses the right slugs' do
+      expect(scraped['related_profiles']).to contain_exactly(
+        'local-government-administrative-assistant',
+        'data-entry-clerk',
+        'hotel-receptionist',
+        'personal-assistant',
+        'health-records-clerk'
+      )
+    end
```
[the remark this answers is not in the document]
tavy88:
```ruby
irb(main):004:0> JobProfile.where(slug: nil)
  JobProfile Load (12.5ms)  SELECT  "job_profiles".* FROM "job_profiles" WHERE "job_profiles"."slug" IS NULL LIMIT $1  [["LIMIT", 11]]
=> #<ActiveRecord::Relation []>
```

No need to test an AR method?

app/models/job_profile.rb
```diff
@@ -14,11 +20,17 @@ def self.import(slug, url)
     end
   end
 
+  def self.bulk_import(slugs)
```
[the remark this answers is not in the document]
tavy88: To keep it consistent with skills import I would prefer to keep it named `import`.

spec/scrapers/job_profile_scraper_spec.rb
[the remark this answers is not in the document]
cwrw: this is in the scraper and not the model

app/webpacker/styles/_job-profile.scss
cwrw: > so maybe its safe to keep your solution for now
As you see from above I was suggesting we keep your solution in my previous comment as couldn't find a better solution. Re the blue bar its one line change so I don't think its a blocker. I think it's important for us to feel able to talk to designers and suggest things we might see that require change

app/models/job_profile.rb
cwrw: Ensuring types are maintained across helps mitigate any unforeseen behaviour/errors. If :list from the scraper returns an [] we should be ok

app/models/job_profile.rb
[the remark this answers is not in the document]
cwrw: import in skills is creating records rather than search https://github.com/DFE-Digital/get-help-to-retrain/blob/master/app/models/skill.rb#L6

app/webpacker/styles/_job-profile.scss
cwrw: talked in person for posterity keeping things as is here

spec/models/job_profile_spec.rb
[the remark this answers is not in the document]
cwrw: talked in person and keeping things as is here

spec/scrapers/job_profile_scraper_spec.rb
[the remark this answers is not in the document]
tavy88: Yeah, I see what you mean.

app/models/job_profile.rb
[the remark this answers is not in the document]
tavy88: Ok I guess you have a point.

app/models/job_profile.rb
tavy88: Wombat returns an empty list if the applied XPath finds nothing.

```ruby
related_profiles 'css=.job-profile-related ul li a/@href', :list
```

Just tested.

spec/scrapers/job_profile_scraper_spec.rb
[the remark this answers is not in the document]
cwrw: I ran this locally and verified that list from wombat returns `[]` so its good

PR #52, opened 2019-07-01
GET-140 Add RSpec feature tests

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-140

* Remove test page view and test
* Make sure to cover full user journey
* Move specific tests to their own files
* Remove cucumber from pipeline as now we have feature tests in RSpec


spec/features/check_your_skills_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  it 'User checks their current skills' do
+    visit(task_list_path)
+    click_on('Check your existing skills')
+    fill_in('name', with: 'Bodyguard')
+    find('.search-button').click
+    click_on('Bodyguard')
+
+    expect(page).to have_text('Patience and the ability to remain calm in stressful situations')
```
tavy88: Are you testing 2 things here?

spec/features/check_your_skills_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    create(
+      :job_profile,
+      name: 'Bodyguard',
+      skills: [
+        create(:skill, name: 'Patience and the ability to remain calm in stressful situations')
+      ]
+    )
+  end
+
+  it 'User checks their current skills' do
```
tavy88:
```ruby
it `searches for a skill` maybe?
```

spec/features/check_your_skills_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    visit(task_list_path)
+    click_on('Check your existing skills')
+    fill_in('name', with: 'Bodyguard')
+    find('.search-button').click
+    click_on('Bodyguard')
+
+    expect(page).to have_text('Patience and the ability to remain calm in stressful situations')
+  end
+
+  it 'User continues journey to explore their careers' do
```
tavy88:
```ruby 
it 'continues journey to explore their careers' 
```
maybe ?

spec/features/explore_occupations_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+RSpec.describe 'Explore Occupations', type: :feature do
+  let!(:job_profile) do
+    create(
+      :job_profile,
+      name: 'Zombie Killer',
+      content: file_fixture('zombie_killer_content.html').read,
+      salary_min: 17_000,
+      salary_max: 28_000,
+      categories: [create(:category, name: 'Apocalyptic services')]
```
tavy88: Could we use `let(:category) { create(:category, name: 'Apocalyptic services')  }` here?

spec/features/explore_occupations_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  it 'User cannot find occupation through search' do
+    visit(explore_occupations_path)
+    fill_in('name', with: 'Embalmer')
+    find('.search-button').click
+
+    expect(page).to have_text('0 results found')
+  end
+
+  it 'User continues journey to finding a training course' do
```
cwrw: move to task list and add path to go to training course from occupation

spec/fixtures/files/zombie_killer_content.html
```diff
@@ -0,0 +1,261 @@
+<main role="main">
```
tavy88: Why do you need this page?

spec/features/check_your_skills_spec.rb
cwrw: There is one expectation, unless you mean the path starts from task list which was intentional, as I would like to test the full path from task list to a job profile skills view, could you clarify what you mean?

spec/features/check_your_skills_spec.rb
cwrw: so I started with using `scenario` which fitted more into those tests description, and rubocop wasn't happy with that and forces the use of `it`. Let me check the usual syntax for scenario descriptions and fix them

spec/fixtures/files/zombie_killer_content.html
cwrw: we need the content for a job, if we don't add content the decorator falls over

spec/features/explore_occupations_spec.rb
cwrw: I think it's ok to keep as is until we have multiple categories, then will move it out

spec/features/check_your_skills_spec.rb
tavy88: Ok fair enough.

spec/features/check_your_skills_spec.rb
cwrw: alright have a look now, should make more sense with scenario

spec/fixtures/files/zombie_killer_content.html
tavy88: Hmm, it shouldn't break the page if the content is missing.

tavy88: Yep ok fair.

spec/features/check_your_skills_spec.rb
tavy88: Do you think it's worth having a test that just checks the rendered page of results? Right now you click on a result and navigate away.

cwrw: My strategy was to cover that by showing that we have a result to click on. Maybe when we care about the actual results like pagination we can add a specific scenario then

spec/features/check_your_skills_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  scenario 'User continues journey to explore their careers' do
+    visit(job_profile_skills_path(job_profile.slug))
+    click_on('Explore jobs you could do')
+
+    expect(page).to have_text('Explore occupations')
+  end
+
+  scenario 'User cannot find occupation through search' do
+    visit(explore_occupations_path)
+    fill_in('name', with: 'Embalmer')
```
smoothcontract: 😂

spec/features/explore_occupations_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+
+RSpec.feature 'Explore Occupations', type: :feature do
+  let!(:job_profile) do
+    create(
+      :job_profile,
+      name: 'Zombie Killer',
+      content: file_fixture('zombie_killer_content.html').read,
```
smoothcontract: Could this be moved to the job profile factory with a specific `:zombie` trait perhaps?

smoothcontract: or maybe something with a more generic trait name that indicates the job profile has content resembling a scraped page from NCS.

spec/fixtures/files/zombie_killer_content.html
```diff
@@ -0,0 +1,261 @@
...
+  <div class="sfPublicWrapper" id="PublicWrapper">
+    <header class="job-profile-hero">
+      <div data-sf-element="Row">
+        <div id="MainContentTop_T41A29498007_Col00" class="sf_colsIn" data-sf-element="Container" data-placeholder-label="Job Profile Hero Container"><div class="content-container">
+          <div class="breadcrumbs govuk-breadcrumbs">
+            <ol class="govuk-breadcrumbs__list">
+              <li class="govuk-breadcrumbs__list-item"><a class="govuk-breadcrumbs__link" href="/">Home: Explore careers</a></li>
+              <li class="govuk-breadcrumbs__list-item">Zombie Killer                </li>
```
smoothcontract: Just thinking ahead a little, maybe this could run through ERB and make use of some placeholders for things like job profile name and description? Then we could have a factory that generates a reasonably faithful NCS page for any given job profile we want to test? It wouldn't be perfect, but may be handy. Perhaps something for later. Thoughts?

PR #54, opened 2019-07-02
GET-169 Add Alternative Titles

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-169

* When scraping job profile pages Persist alternative titles in the database in an array column
* Remove sub hero content from Job Profile decorator
* Add alternative titles lists in check your skills and explore occupations search results
* Substitute alternative titles for sub hero content in Job Profile page
* fix some styles where we are missing some margins

### DIFF
--- a/db/migrate/20190701155237_add_alternative_titles_to_job_profiles.rb
+++ b/db/migrate/20190701155237_add_alternative_titles_to_job_profiles.rb
@@ -1,0 +1,5 @@
+class AddAlternativeTitlesToJobProfiles < ActiveRecord::Migration[5.2]
+  def change
+    add_column :job_profiles, :alternative_titles, :string, array: true, null: false, default: []
+  end
+end
--- a/app/decorators/job_profile_decorator.rb
+++ b/app/decorators/job_profile_decorator.rb
@@ -1,4 +1,3 @@
-# rubocop:disable Metrics/ClassLength
 # TODO: Most of the xpath expressions within this class will be migrated to JobProfileScraper over time
 # which should remove the need to disable rubocop rules here.
 class JobProfileDecorator < SimpleDelegator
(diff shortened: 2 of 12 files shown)


cwrw:
I went with the helper as we aren't instantiating the decorator in all controllers, and we intend to phase the decorator out

smoothcontract:
> I went with the helper as we aren't instantiating the decorator in all controllers, and we intend to phase the decorator out

I quite like the decorator pattern, perhaps even over the helper pattern as it provides a nice scoping mechanism rather than having everything in the global helper namespace. But I'd agree we want to phase out the on fly html parsing aspects of the decorator.

cwrw:
I think it's a good point though, will keep it in mind for next time, maybe we'll get some form of pattern which we can use when we move away from the current state

db/migrate/20190701155237_add_alternative_titles_to_job_profiles.rb
```diff
@@ -0,0 +1,5 @@
+class AddAlternativeTitlesToJobProfiles < ActiveRecord::Migration[5.2]
+  def change
+    add_column :job_profiles, :alternative_titles, :string, array: true, null: false, default: []
```
tavy88: I think having a tags table would have been a better option especially if we want to exploit these tags later on (e.g filtering by tags). I assume the reason you stored them as an array was exactly to maybe use them in future?

cwrw: We discussed this in person, and agreed that we can easily move to a tags table later on if we decide to search by alternative titles, for now its safe to leave it here until the need arises

tavy88: Ok fair. However, keep in mind we are doing the same work twice if that will be the case. We will need a new migration etc etc.

PR #56, opened 2019-07-03
GET-104 Job title blank search

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-104

* Guard against search results being nil or empty on the backend
* Add a new search-js class to disable search form if field is empty. Note that due to turbolinks messing up the relaod and events we have to listen on `turbolink:load` each time the page changes
* Add feature tests for disabling search, which meant we have to enable js on the test. VCR complains about needing to hit outer links so we allow that unless a cassette is defined

### DIFF
--- a/app/models/job_profile.rb
+++ b/app/models/job_profile.rb
@@ -5,6 +5,8 @@ class JobProfile < ApplicationRecord
   has_many :skills, through: :job_profile_skills, inverse_of: :job_profiles
 
   def self.search(name)
+    return none unless name.present?
+
     where('name ILIKE ?', "%#{name&.squish}%")
   end
 
--- a/app/webpacker/packs/search-form.js
+++ b/app/webpacker/packs/search-form.js
@@ -9,7 +9,9 @@ function SearchForm () {
   }
 
   function disableEmptyForm($form, $field) {
+
     if ($form) {
+      $field.required = false;
       $form.addEventListener('submit', event => {
         if ($field.value.trim().length == 0) {
           event.preventDefault();
--- a/app/webpacker/packs/search-form.js
+++ b/app/webpacker/packs/search-form.js
@@ -9,7 +9,6 @@ function SearchForm () {
   }
 
   function disableEmptyForm($form, $field) {
-
     if ($form) {
       $field.required = false;
       $form.addEventListener('submit', event => {
--- a/app/webpacker/packs/search-form.js
+++ b/app/webpacker/packs/search-form.js
@@ -1,0 +1,22 @@
+function SearchForm () {
+  this.start = function () {
+    document.addEventListener("turbolinks:load", function() {
+      var $jobProfileName = document.querySelector('#name');
+      var $jobProfileSearchForm = document.querySelector('#job-profile-search');
+
+      disableEmptyForm($jobProfileSearchForm, $jobProfileName);
+    });
+  }
+
+  function disableEmptyForm($form, $field) {
+    if ($form) {
+      $form.addEventListener('submit', event => {
+        if ($field.value.trim().length == 0) {
+          event.preventDefault();
+        }
+      });
+    }
+  }
+}
+
+module.exports = new SearchForm();
--- a/spec/features/check_your_skills_spec.rb
+++ b/spec/features/check_your_skills_spec.rb
@@ -34,14 +34,4 @@ RSpec.feature 'Check your skills', type: :feature do
 
     expect(page).to have_text('0 results')
   end
-
-  feature 'when searching with no data', js: true, driver: :selenium_chrome_headless do
-    scenario 'User cannot see any results' do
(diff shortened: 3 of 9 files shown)


spec/features/check_your_skills_spec.rb
```diff
@@ -28,10 +28,20 @@
...
   scenario 'User cannot find occupation through search' do
-    visit(explore_occupations_path)
+    visit(check_your_skills_path)
     fill_in('name', with: 'Embalmer')
     find('.search-button').click
 
     expect(page).to have_text('0 results')
   end
+
+  feature 'when searching with no data', js: true do
```
cwrw: what are people's thoughts in regards to these tests?

app/models/job_profile.rb
```diff
@@ -5,6 +5,8 @@ class JobProfile < ApplicationRecord
   has_many :skills, through: :job_profile_skills, inverse_of: :job_profiles
 
   def self.search(name)
+    return none unless name.present?
```
smoothcontract: Interesting, I hadn't come across the `none` keyword before used like this. Pulled the branch and it returns an empty AR relation so all good. TIL 😀

app/webpacker/packs/search-form.js
```diff
@@ -0,0 +1,22 @@
...
+      $form.addEventListener('submit', event => {
+        if ($field.value.trim().length == 0) {
+          event.preventDefault();
+        }
+      });
+    }
+  }
+}
+
+module.exports = new SearchForm();
```
smoothcontract: Very clean pattern 👍

spec/features/check_your_skills_spec.rb
```diff
@@ -28,10 +28,20 @@
   end
 
   scenario 'User cannot find occupation through search' do
-    visit(explore_occupations_path)
+    visit(check_your_skills_path)
```
smoothcontract: I'm guessing this spec was previously incorrectly testing the wrong form?

spec/rails_helper.rb
```diff
@@ -19,6 +19,7 @@
   vcr.hook_into :webmock
   vcr.ignore_localhost = true
   vcr.configure_rspec_metadata!
+  vcr.allow_http_connections_when_no_cassette = true
```
smoothcontract: I disabled this line locally on the branch and am still seeing specs passing. Are you sure it's needed?

spec/features/check_your_skills_spec.rb
cwrw: yea unfortunately 🥇

app/webpacker/packs/search-form.js
cwrw: went with singleton for now, might need to rethink if things change later

spec/rails_helper.rb
cwrw: oh your right, I thought I needed it but since I restructured the tests I don't anymore, will remove

tavy88:
BTW build is failing and needs sorting out the merge conflicts.

spec/rails_helper.rb
cwrw: Actually correction these are needed in docker itself. The gem apparently pulls in drivers: https://dfe-ssp.visualstudio.com/S108-Get-Help-To-Retrain/_build/results?buildId=12581

cwrw:
It seems to add the JS testing we need to include some Chrome binaries. Also we need some extra setup which isn't ideal as we are running this container in production. I think we should first do some work to separate testing from production when building containers, introduce docker compose to include the selenium setup and then add JS tests. What do you think?

spec/features/explore_occupations_spec.rb
```diff
@@ -46,4 +46,14 @@
...
     expect(page).to have_text('Find and apply to a training course near you')
   end
+
+  feature 'when searching with no data', js: true, driver: :selenium_chrome_headless do
+    scenario 'User cannot see any results' do
+      visit(explore_occupations_path)
+      fill_in('name', with: '')
+      find('.search-button').click
+
+      expect(page).not_to have_text('Search results for')
```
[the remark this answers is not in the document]
cwrw: removed the tests (see comment below)

app/webpacker/packs/search-form.js
```diff
@@ -0,0 +1,23 @@
...
+      var $jobProfileName = document.querySelector('#name');
+      var $jobProfileSearchForm = document.querySelector('#job-profile-search');
+
+      disableEmptyForm($jobProfileSearchForm, $jobProfileName);
+    });
+  }
+
+  function disableEmptyForm($form, $field) {
+    if ($form) {
+      $field.required = false;
```
smoothcontract: 👍

cwrw:
Unfortunately I've tried the basic test without JS but it navigates to the next page and ignores the require on the input field

PR #57, opened 2019-07-05
Add loggly for production logging

smoothcontract:
### Context
This is a short term solution until we move to a native Azure logging approach.

I've added support for an optional `LOGGLY_TOKEN` environment variable. If this is set, then loggly is used as the preferred logger, with all log traffic captured and forwarded.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,3 +1,4 @@
+# rubocop:disable Metrics/BlockLength
 Rails.application.configure do
   # Verifies that versions and hashed value of the package contents in the project's package.json
   config.webpacker.check_yarn_integrity = false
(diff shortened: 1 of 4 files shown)


config/environments/production.rb
```diff
@@ -77,10 +77,11 @@
...
   # Use a different logger for distributed setups.
-  # require 'syslog/logger'
-  # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
-
-  if ENV['RAILS_LOG_TO_STDOUT'].present?
+  if ENV['LOGGLY_TOKEN'].present?
+    loggly_url = "https://logs-01.loggly.com/inputs/#{ENV['LOGGLY_TOKEN']}/tag/ruby/"
+    config.logger = Logglier.new(loggly_url, threaded: true)
+    config.logger.formatter = config.log_formatter
+  elsif ENV['RAILS_LOG_TO_STDOUT'].present?
```
cwrw: can we not send `STDOUT` to loggly generally?

tavy88: Also, why are LOGGLY_TOKEN & RAILS_LOG_TO_STDOUT mutually exclusive (according to this config and how would that work for our envs)?

smoothcontract: It's possible - there's an alternative transport mechanism that essentially pipes STDOUT to loggly via another process (presumably that runs in the container alongside rails). I think we're trying to stick to the model of one process only in the containers though so that's not ideal. I'll see if there's any other way to tee the logs to both - worst case we could fork the logglier gem to add that in.

smoothcontract: Found a technique to allow logs to be chained together, so supports both at once now.

PR #60, opened 2019-07-08
GET-177 Add Full Text Search to Job Profiles

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-177

Trying to use alternative titles in its current state yielded strange results when attempting to search. It mainly felt like we were fighting the schema by turning the string into an array, and then back to a string everywhere else to use including search and views. 
While attempting to search alternative titles when in an array form it took 50ms for a the following query:
```
joins('CROSS JOIN LATERAL unnest(alternative_titles) AS alternative_title')
  .where("to_tsvector('english', name) @@ to_tsquery('english', :q) or to_tsvector('english', alternative_title) @@ to_tsquery('english', :q) or to_tsvector('english', description) @@ to_tsquery('english', :q)", q: names)
  .order("#{rank} desc")
  .uniq
```

This was due to the need to unnest the array field and then search against single words. Also indexing this field proved problematic and required extra work which was advised against usually, where it required us to create a custom function to create immutable array object and then pass it in as follows:
```
add_index :job_profiles, "to_tsvector('english', COALESCE(f_array_to_text(alternative_titles),''))", using: 'gin', name: 'job_profiles_alternative_titles_idx'

```

This didn't help much with the results.
Instead, after switching the column back to a string, and indexing it, the current query takes 1.3ms on average.

Please see individual commits for more detail especially about sanitising query text and preventing SQL injection


app/models/job_profile.rb
```diff
@@ -10,10 +10,32 @@ class JobProfile < ApplicationRecord
...
+  def self.search(string)
+    return none unless string.present?
 
-    where('name ILIKE ?', "%#{name&.squish}%")
+    query_string = QueryStringFormatter.format(string)
+
+    select(:id, :name, :description, :alternative_titles, :slug, :salary_min, :salary_max)
+      .where(build_text_search_query, query_string: query_string)
+      .order(
+        Arel.sql(build_rank_query(query_string)) => :desc
```
cwrw: if we don't use `Arel.sql` we get deprecation warnings: DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s). Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql()

Also brakeman still complains about SQL injection even after we have formatted the string heavily and are also escaping and quoting it on line 34

db/migrate/20190707130437_change_alternative_titles_type_to_string.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeAlternativeTitlesTypeToString < ActiveRecord::Migration[5.2]
+  def change
+    change_column :job_profiles, :alternative_titles, :string, array: false, null: true, default: nil, using: "(array_to_string(ARRAY[alternative_titles], ', '))"
```
smoothcontract: I suspect this won't work in reverse - perhaps this could be split into separate `up` / `down` methods. Not critical in this case but generally a nice practice. What happens to the column if this migration is reversed - is it left with `nil` value?

app/views/check_your_skills/results.html.erb
```diff
@@ -17,7 +17,7 @@
     <% if @job_profiles.blank? %>
       <p class="govuk-body-m"><%= render '/shared/search/no_results' %></p>
     <% else %>
-      <p class="govuk-body-m" ><%= t('.results', count: @job_profiles.count) %></p>
+      <p class="govuk-body-m" ><%= t('.results', count: @job_profiles.length) %></p>
```
cwrw: This is needed as when calling attempting a `.count`, a `COUNT` query is added to the original query making it incorrect and fails

app/models/query_string_formatter.rb
```diff
@@ -0,0 +1,17 @@
+module QueryStringFormatter
```
smoothcontract: Is there anywhere better to place this within the app than as a model (that's really not a model but a module) ?

spec/models/query_string_formatter_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    it 'converts characters into their compatibility characters' do
+      expect(described_class.normalize('2⁵')).to eq('25')
+    end
+
+    it 'composes all characters into single code points' do
+      expect(described_class.normalize('Ƶ̧̈')).to eq('Ƶ')
+    end
+
+    it 'removes any non alphanumeric characters from the string' do
+      expect(described_class.normalize('<script> ("pwned!")123 </script>')).to eq('script pwned123 script')
```
tavy88: Why not strip all tags?

app/models/query_string_formatter.rb
cwrw: could move to lib folder

db/migrate/20190707130437_change_alternative_titles_type_to_string.rb
cwrw: 🦅 eyes :D I tried creating an up/down but it kept complaining about immutable objects even after creating custom functions for `string_to_array` as the one out of the box wouldn't work. Will give it another try

app/models/query_string_formatter.rb
```diff
@@ -0,0 +1,17 @@
...
+  def format(string)
+    normalize(string)
+      .squish
+      .split(/\s/)
+      .map { |str| str + ':*' }
+      .join(' | ')
+  end
```
tavy88: What is it you're trying to achieve here exactly?

app/models/job_profile.rb
```diff
@@ -10,10 +10,32 @@ class JobProfile < ApplicationRecord
...
+  def self.build_text_search_query
+    <<-SQL
+      to_tsvector('english', name) @@ to_tsquery('english', :query_string)
+      OR to_tsvector('english', alternative_titles) @@ to_tsquery('english', :query_string)
+      OR to_tsvector('english', description) @@ to_tsquery('english', :query_string)
+    SQL
+  end
+
+  def self.build_rank_query(query_string)
+    qouted_query_string = ActiveRecord::Base.connection.quote(query_string)
```
tavy88: is it not `quoted`?

app/models/query_string_formatter.rb
tavy88: you could use a `formatters` folder as well?

app/views/check_your_skills/results.html.erb
tavy88: How about `size`?

https://mensfeld.pl/2014/09/activerecord-count-vs-length-vs-size-and-what-will-happen-if-you-use-it-the-way-you-shouldnt/

app/models/query_string_formatter.rb
cwrw: As per the commit message: 
To allow words to be used as prefixes use the ":*" notation to
be appended to each word's ending, and an OR between multiple words
to allow the query to search across words entered. 
So for instance a string "Cook Chef" will be changed to "Cook:* | Chef:* " this means that we can search for any words that are prefixed with those words (due to :*) and use an OR between them so we search for both words in our Job Profiles rather than an &

app/models/job_profile.rb
cwrw: The method we are using is from: https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Quoting.html#method-i-quote

It does some extra work to make sure a string is quoted correctly, have a look at: https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L164

spec/models/query_string_formatter_spec.rb
cwrw: I opted out from introducing specific things like ignoring tags" as that doesn't serve much purpose. I went with ignoring anything that isn't alphanumeric (a-A and 0-9) This way we make sure to strip everything regardless

app/models/query_string_formatter.rb
tavy88: Would you consider using `concat` instead of `+`? https://stackoverflow.com/questions/23763215/ruby-array-concat-versus-speed

My concern is that there is a lot of transformations happening on the fly here. So every improvement helps.

app/models/query_string_formatter.rb
```diff
@@ -0,0 +1,17 @@
+module QueryStringFormatter
+  def format(string)
+    normalize(string)
+      .squish
+      .split(/\s/)
+      .map { |str| str + ':*' }
```
tavy88: Would you consider using concat instead of +? https://stackoverflow.com/questions/23763215/ruby-array-concat-versus-speed

My concern is that there is a lot of transformations happening on the fly here. So every improvement helps.

spec/models/query_string_formatter_spec.rb
tavy88: Sure but taking out the whole tags would also contribute to performance as it's less tokenization happening and fewer words to look for. I don't, however, anticipate people using tags when searching. But think if someone copy-pastes a whole JS snippet in the search?

app/models/query_string_formatter.rb
cwrw: We are using `String` `+` not `Array` `+` here: https://ruby-doc.org/core-2.4.0/String.html#method-i-2B so I think that is quite different.
This take ~ 1 millisecond to complete, if we do see some issues in performance we can revisit

tavy88: Yeah, it's the same discussion with strings: https://coderwall.com/p/ac5j9g/or-concat-what-is-faster-for-appending-string-in-ruby (sorry for my previous example as it was targetting arrays).

db/migrate/20190707130437_change_alternative_titles_type_to_string.rb
cwrw: This has now been added

tavy88:
Also, @cwrw be aware that your change is not backwards compatible.

spec/models/query_string_formatter_spec.rb
cwrw: That is interesting. I'd say let's wait and see what people put into this field and act accordingly. In terms of injections we are safe, in terms of long strings we can gauge that with user testing and add a limit

app/models/job_profile.rb
tavy88: No haha sorry I meant the var name. It's a typo.

PR #62, opened 2019-07-08
Custom domain names

mafonso:
### Context
https://dfedigital.atlassian.net/browse/GET-158
https://dfedigital.atlassian.net/browse/GET-159

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -14,6 +14,13 @@
         "description": "The abbreviation of the service name to be used when naming a resource."
       }
     },
+    "customHostName": {
+      "type": "string",
+      "defaultValue": "",
+      "metadata": {
+        "description": "The custom hostname to add to the app service."
+      }
+    },
     "databaseName": {
       "type": "string",
       "metadata": {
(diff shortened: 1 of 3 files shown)


mafonso:
> Lots of configs.

yeah... domain, cert, vault and it's RG.. :(

azure/template.json
```diff
@@ -314,6 +398,12 @@
                 "value": "[parameters('secretKeyBase')]"
               }
             ]
+          },
+          "customHostName": {
+            "value": "[parameters('customHostName')]"
+          },
+          "certificateThumbprint": {
+            "value": "[if(greater(length(parameters('customHostname')), 0), reference('app-service-certificate', '2018-11-01').outputs.certificateThumbprint.value, '')]"
```
cwrw: talked in person but for posterity double check the use of a version, and if we don't need to pass it in

mafonso: yeah, I don't think I need to pin the version of the API here. It is already pined on the nested template I'm importing for this resource.  It should be pinned somewhere as I'm passing parameters to that version of the API as if there is a change upstream it will stop working.  
Azure does not deprecate resource API versions so the one we're using will always be there.

PR #64, opened 2019-07-08
GET-183 Remove Turbolinks

smoothcontract:
### Context
Turbolinks is intended to improve page load times, by avoiding the need to fetch and parse CSS and JS assets described in the `head` section of the page. In practice this works to a degree, but has complications.

Already we have had to adapt the progressive enhancement for search to listen to custom turbolinks events. The Azure App Insights tracking script is working intermittently since this is only firing on initial page load or manual refresh, not when user navigates to a new page by clicking links.

There are numerous anecdotal reports of various issues with the library, including double loading of pages, accessibility limitations etc. Overall, it's not adding value so good to remove.

Completely remove turbolinks library and any associated setup. Use standard window load event to hook search functionality.

### DIFF
--- a/app/webpacker/packs/search-form.js
+++ b/app/webpacker/packs/search-form.js
@@ -1,11 +1,9 @@
 function SearchForm () {
   this.start = function () {
-    document.addEventListener("turbolinks:load", function() {
-      var $jobProfileName = document.querySelector('#name');
-      var $jobProfileSearchForm = document.querySelector('#job-profile-search');
+    const $jobProfileName = document.querySelector('#name');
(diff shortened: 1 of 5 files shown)


app/webpacker/packs/search-form.js
```diff
@@ -1,6 +1,6 @@
 function SearchForm () {
   this.start = function () {
-    document.addEventListener("turbolinks:load", function() {
+    window.addEventListener('load', function() {
```
cwrw: do we still need this, can we take it fully out?

smoothcontract: It seems we can 😀

PR #69, opened 2019-07-10
GET-134 Add link to smart survey for feedback

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-134

Add extra section on task list page for link to survey. This should
load through the same page. Add link to feedback in banner, this one
should go to a different tab.

### DIFF
--- a/app/views/pages/task_list.html.erb
+++ b/app/views/pages/task_list.html.erb
@@ -74,6 +74,21 @@
           </li>
         </ul>
       </li>
+      <li>
+        <h2 class="app-task-list__section">
+          <span class="app-task-list__section-number">5. </span> Help improve this service
+        </h2>
+        <ul class="app-task-list__items">
+          <li class="app-task-list__item">
+            <div class="app-task-list__task-name">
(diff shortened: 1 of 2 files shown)


app/views/pages/task_list.html.erb
```diff
@@ -74,6 +74,21 @@
           </li>
         </ul>
       </li>
+      <li>
+        <h2 class="app-task-list__section">
```
smoothcontract: Is the double underscore thing a GDS naming convention?

cwrw: yea, they are using the BEM syntax for css classes http://getbem.com/introduction/

cwrw:
My go to is usually including the jira ticket which has the mockups but it seems those are missing, and the person can pull in the branch locally and play around. I could include screenshots, but I prefer not to as it doesn't help when you have many different use cases as well as paths. Maybe deploying to an environment might be good, but sometimes it becomes worse if there are changes requested on a pr and you have to go through many iterations. In this case I think I could deploy it as it's pretty straight forward

PR #74, opened 2019-07-11
Add client side tracking for User Exit Survey Link

tavy88:
### Context

Adding client-side tracking for when the user clicks both of the User Exit Survey links. 
Tracking done inside AppInsights

### Screenshots

Screen Shot 2019-07-11 at 15 10 39

Screen Shot 2019-07-12 at 08 54 51

### Ticket

https://dfedigital.atlassian.net/browse/GET-211

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -51,7 +51,7 @@
           <p class="govuk-phase-banner__content">
             <strong class="govuk-tag govuk-phase-banner__content__tag ">beta</strong>
             <span class="govuk-phase-banner__text">
-              This is a new service – your <%= link_to 'feedback', 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', target: '_blank', data: { tracked_event: 'User Exit Survey Link', type: 'link' }%> will help us to improve it.
+              This is a new service – your <%= link_to 'feedback', 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', target: '_blank', data: { tracked_event: 'Banner - User Exit Survey Link' }%> will help us to improve it.
             </span>
           </p>
         </div>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -51,7 +51,7 @@
           <p class="govuk-phase-banner__content">
             <strong class="govuk-tag govuk-phase-banner__content__tag ">beta</strong>
             <span class="govuk-phase-banner__text">
-              This is a new service – your <%= link_to 'feedback', 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', target: '_blank', data: { tracked__event: 'User Exit Survey Link', type: 'link' }%> will help us to improve it.
+              This is a new service – your <%= link_to 'feedback', 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', target: '_blank', data: { tracked_event: 'User Exit Survey Link', type: 'link' }%> will help us to improve it.
             </span>
           </p>
         </div>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -51,7 +51,7 @@
           <p class="govuk-phase-banner__content">
             <strong class="govuk-tag govuk-phase-banner__content__tag ">beta</strong>
             <span class="govuk-phase-banner__text">
-              This is a new service – your <%= link_to 'feedback', 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', target: '_blank'%> will help us to improve it.
+              This is a new service – your <%= link_to 'feedback', 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', target: '_blank', data: { tracked__event: 'User Exit Survey Link', type: 'link' }%> will help us to improve it.
             </span>
           </p>
         </div>
--- a/app/webpacker/packs/track-events.js
+++ b/app/webpacker/packs/track-events.js
@@ -1,6 +1,6 @@
 function TrackEvents () {
   this.start = function () {
-    if (appInsights) {
+    if (typeof(appInsights) === 'object') {
       document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
         $this = element;
 
--- a/app/webpacker/packs/track-events.js
+++ b/app/webpacker/packs/track-events.js
@@ -1,19 +1,25 @@
 function TrackEvents () {
   this.start = function () {
-    if (typeof(appInsights) === 'object') {
+    if (typeof(appInsights) !== 'undefined') {
       document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
-        if (element.dataset.type === 'link') {
-          element.onclick = function sendEvent() {
-            if (element.dataset.trackedEvent) {
-              setTimeout(function () {
-                appInsights.trackEvent(element.dataset.trackedEvent);
-              }, 0);
-            }
-          };
-        }
+        trackComponentOnClick(element);
       });
     }
   };
+
+  function sendEvent() {
+    var eventLabel = this.dataset.trackedEvent;
+
+    setTimeout(function () {
+      appInsights.trackEvent(eventLabel);
+    }, 0);
+  };
+
+  function trackComponentOnClick(element) {
+    if (element.dataset.trackedEvent) {
+      element.addEventListener('click', sendEvent, false);
+    }
+  }
 }
 
 export default new TrackEvents();
--- a/app/webpacker/packs/track-events.js
+++ b/app/webpacker/packs/track-events.js
@@ -1,0 +1,19 @@
+function TrackEvents () {
+  this.start = function () {
+    if (appInsights) {
+      document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
+        $this = element;
+
+        if ($this.dataset.type === 'link') {
+          $this.onclick = function sendEvent(e) {
+            if ($this.dataset.trackedEvent) {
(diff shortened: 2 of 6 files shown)


app/webpacker/packs/track-events.js
```diff
@@ -0,0 +1,19 @@
...
+              appInsights.trackEvent($this.dataset.trackedEvent)
+            }
+          }
+        }
+      });
+    }
+  }
+}
+
+module.exports = new TrackEvents();
```
tavy88: I tried making this as generic as possible and as reusable as possible.

app/views/layouts/application.html.erb
```diff
@@ -51,7 +51,7 @@
           <p class="govuk-phase-banner__content">
             <strong class="govuk-tag govuk-phase-banner__content__tag ">beta</strong>
             <span class="govuk-phase-banner__text">
-              This is a new service – your <%= link_to 'feedback', 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', target: '_blank'%> will help us to improve it.
+              This is a new service – your <%= link_to 'feedback', 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', target: '_blank', data: { tracked__event: 'User Exit Survey Link', type: 'link' }%> will help us to improve it.
```
smoothcontract: This has two underscores in the `data` tag for `tracked__event`

app/webpacker/packs/track-events.js
```diff
@@ -0,0 +1,19 @@
+function TrackEvents () {
+  this.start = function () {
+    if (appInsights) {
```
smoothcontract: I don't think this is sufficient to check that `appInsights` variable is defined. Tried this in a browser console on a site without the tracking script loaded (here on github actually) and this seems to raise an error:

```
if (appInsights) { console.log('yes') };
VM373:1 Uncaught ReferenceError: appInsights is not defined
    at <anonymous>:1:1
```

Take a look at this - better to use `typeof` operator I think. https://stackoverflow.com/questions/5113374/javascript-check-if-variable-exists-is-defined-initialized

app/webpacker/packs/track-events.js
```diff
@@ -0,0 +1,19 @@
...
+      document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
+        $this = element;
+
+        if ($this.dataset.type === 'link') {
+          $this.onclick = function sendEvent(e) {
+            if ($this.dataset.trackedEvent) {
+              appInsights.trackEvent($this.dataset.trackedEvent)
```
smoothcontract: Is this a synchronous call? What happens if the call takes a while and the link seems unresponsive so user clicks again?

app/webpacker/packs/track-events.js
smoothcontract: Cool 👍

app/webpacker/packs/track-events.js
tavy88: Fair point.

app/webpacker/packs/track-events.js
tavy88: True

app/webpacker/packs/track-events.js
tavy88: However, adding this:

```ruby
function TrackEvents () {
  this.start = function () {
    if (typeof(appInsights) === 'object') {
      document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
        $this = element;

        if ($this.dataset.type === 'link') {
          $this.onclick = function sendEvent(e) {
            if ($this.dataset.trackedEvent) {
              appInsights.trackEvent($this.dataset.trackedEvent)
              
              setTimeout(alertFunc, 6000);
            }
          }
        }
      });
    }
  }

  function alertFunc() {
    alert("Hello!");
  }
}
```

Does not prevent the redirect nor stops the page. Even if something breaks in the script, you get an error in the console but you still get redirected to the survey.

tavy88: However, if the code fails it still redirects and the error can be seen in the console.

tavy88: To be safer I decided to use one of JS's native async function `setTimeout`. I initially played with Promises but turns out sync code & promises doesn't play nice. Here's an example of the initial Promise implementation:

```ruby
function trackEvent(eventLabel) {
    const sendEvent = new Promise(function (resolve, reject) {
      try {
        var i = 0;
        while(i < 20) {
          console.log('alert');
          i = i + 1;
        }
        appInsights.trackEvent(eventLabel);
        resolve();
      }
      catch (err) {
        reject(err);
      }
    });
    
    sendEvent
      .then(function success(response) {
        return response;
      })
      .catch(function failure(err) {
        console.error(err);
      });
    }
```

Everything is fine until I add the following block: 

```
var i = 0;
        while(i < 20) {
          console.log('alert');
          i = i + 1;
        }
```
which blocks the main thread, hence my decision to use setTimeout - a native async function.

app/views/layouts/application.html.erb
tavy88: Oops...Copy paste typo.

app/webpacker/packs/track-events.js
```diff
@@ -0,0 +1,19 @@
...
+              }, 0);
+            }
+          };
+        }
+      });
+    }
+  };
+}
+
+export default new TrackEvents();
```
tavy88: To be safer I decided to use one of JS's native async function setTimeout. I initially played with Promises but turns out sync code & promises doesn't play nice. Here's an example of the initial Promise implementation:

function trackEvent(eventLabel) {
    const sendEvent = new Promise(function (resolve, reject) {
      try {
        var i = 0;
        while(i < 20000) {
          console.log('alert');
          i = i + 1;
        }
        appInsights.trackEvent(eventLabel);
        resolve();
      }
      catch (err) {
        reject(err);
      }
    });
    
    sendEvent
      .then(function success(response) {
        return response;
      })
      .catch(function failure(err) {
        console.error(err);
      });
    }
Everything is fine until I add the following block:

var i = 0;
        while(i < 20000) {
          console.log('alert');
          i = i + 1;
        }
which blocks the main thread, hence my decision to use setTimeout - a native async function.

cwrw:
I've had a look through the code, and was hoping to see something use made of javascript functions, as the code currently stands, I found it really hard to follow. Maybe something that looks similar to this might be easier to understand and generic enough as well, what do you think? 
```
function TrackEvents () {
  this.start = function () {
    if(typeof appInsights !== 'undefined') {
      trackLinks();
    }
  }

  function trackLinks() {
    document.querySelectorAll('a[data-tracked-event]').forEach(function(element) {
      element.addEventListener('click', trackInsights, false);
    });
  }

  function trackInsights() {
    if (this && this.hasAttribute('data-type')) {
      appInsights.trackEvent(this.dataset.trackedEvent);
    }
  }
}

module.exports = new TrackEvents();
```

cwrw:
> TrackLinks && TrackInsights is a bit misleading as we do not track 2 different things. We track custom event on links. 
This is an example of how we can structure. Feel free to rename the functions

> Also, your code runs the reporting to App insights sync.
You can add the `setTimeout` inside one of the functions as well

> Also why is typeof appInsights !== 'undefined' better than typeof appInsights === 'object' ?
When having a look through
https://stackoverflow.com/questions/5113374/javascript-check-if-variable-exists-is-defined-initialized
says that However  typeof null will return "object" so we need to actually check for undefined as well as null and not just object

I think generally this is a guidance to make the code more readable than it currently stands and split one function into something procedural to follow

tavy88:
Yeah, I'm not against breaking the code into more functions -- will try to sexy it up a bit.

app/webpacker/packs/track-events.js
```diff
@@ -0,0 +1,19 @@
...
+  };
+
+  function sendEvent() {
+    setTimeout(function () {
+      appInsights.trackEvent(this.dataset.trackedEvent);
+    }, 0);
+  };
+}
+
+export default new TrackEvents();
```
cwrw: Is there a difference between `export default new TrackEvents();` and `module.exports = new TrackEvents();` ? If there isn't could we stick to `module.exports = new TrackEvents();` to maintain consistency. see https://github.com/DFE-Digital/get-help-to-retrain/blob/master/app/webpacker/packs/search-form.js#L21

app/webpacker/packs/track-events.js
```diff
@@ -0,0 +1,19 @@
+function TrackEvents () {
+  this.start = function () {
+    if (appInsights !== null && typeof(appInsights) !== 'undefined') {
+      document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
```
cwrw: could we also move this to a method? As `this.start` is more of a constructor or initialiser for the application, and ideally will be as  clean as possible, and if we were to extend it with more events to track it would be easier, as so: 
```
 this.start = function () {
    if (appInsights !== null && typeof(appInsights) !== 'undefined') {
       track_clicks()
       track_something_else()
     }
  };
```
It will also help with the readability

app/webpacker/packs/track-events.js
tavy88: It's the more modern way of doing it with ES6

app/webpacker/packs/track-events.js
tavy88: document.querySelectorAll('[data-tracked-event]') is not tracking just links but tracks any component link, button etc that has this data attributes on. I'd like to keep it like this for now.

app/webpacker/packs/track-events.js
```diff
@@ -0,0 +1,25 @@
...
+  this.start = function () {
+    if (typeof(appInsights) !== 'undefined') {
+      document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
+        trackComponentOnClick(element);
+      });
+    }
+  };
+
+  function sendEvent() {
+    var eventLabel = this.dataset.trackedEvent
```
cwrw: 🦂 `;`

tavy88: Ha ha fair...

PR #75, opened 2019-07-11
GET-101 Add Pagination

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-101

* Use Kaminari gem to paginate results pages for both check your skills and explore occupations search
* Add Kaminari views and edit styles according to HMRC pagination:
http://hmcts-design-system.herokuapp.com/components/pagination
This has been reviewed by Steve and approved to be used. So disregard ticket stylings

This is deployed to for review: 
https://s108d02-dev-as.azurewebsites.net/check_your_skills/results?utf8=%E2%9C%93&name=e
https://s108d02-dev-as.azurewebsites.net/explore_occupations/results?utf8=%E2%9C%93&name=ex

### DIFF
--- a/app/controllers/explore_occupations_controller.rb
+++ b/app/controllers/explore_occupations_controller.rb
@@ -4,7 +4,8 @@ class ExploreOccupationsController < ApplicationController
   end
 
   def results
-    @job_profiles = JobProfile.search(job_profile_params[:name]).includes(:categories).map { |job_profile|
+    @search_results = JobProfile.search(job_profile_params[:name]).includes(:categories).page(params[:page])
+    @job_profiles = @search_results.map { |job_profile|
       JobProfileDecorator.new(job_profile)
     }
   end
--- a/Gemfile
+++ b/Gemfile
@@ -39,6 +39,9 @@ gem 'canonical-rails'
 # Logging
 gem 'logglier', '~> 0.5'
 
+# Pagination
+gem 'kaminari', '~> 1.1', '>= 1.1.1'
+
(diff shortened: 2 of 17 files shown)


app/controllers/explore_occupations_controller.rb
```diff
@@ -4,7 +4,8 @@ def index
   end
 
   def results
-    @job_profiles = JobProfile.search(job_profile_params[:name]).includes(:categories).map { |job_profile|
+    @search_results = JobProfile.search(job_profile_params[:name]).includes(:categories).page(params[:page])
```
smoothcontract: Does `@search_results` need to be an instance variable rather than a local variable here?

smoothcontract: Nevermind, I saw it used later in the view.

cwrw:
@smoothcontract feel free to play around on it here:
https://s108d02-dev-as.azurewebsites.net/check_your_skills/results?utf8=%E2%9C%93&name=e
https://s108d02-dev-as.azurewebsites.net/explore_occupations/results?utf8=%E2%9C%93&name=ex

cwrw:
> I presume the kaminari partials are customised versions of the default ones provided? 
Yea its generated as part of the gem if you want to customise. It's a bit strange as wasn't very happy to include partials with the gem name but I don't think there's another option

Regarding tests I'll add something to spec features 👍

smoothcontract:
Cool, thanks for adding some feature coverage 👍

PR #76, opened 2019-07-11
azure-pipelines.yml overhauling. 

mafonso:

- Remake the pipeline using Azure Tasks
- Introduce multi-stage to make room to include releases
- Introduces service connections to the Docker HUB registry to avoid managing credentials to use it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,11 @@ RUN apk add --update $BUILD_PACKAGES && \
 # Generate Assets
 WORKDIR /app
 COPY Gemfile Gemfile.lock .ruby-version ./
-RUN bundle install --clean --force --without "development"
+RUN bundle install --clean --force --without "development" \
+     && rm -rf /usr/local/bundle/cache/*.gem \
+     && find /usr/local/bundle/gems/ -name "*.c" -delete \
+     && find /usr/local/bundle/gems/ -name "*.o" -delete
+
 COPY package.json yarn.lock ./
 RUN yarn install
 COPY . ./
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,8 @@ group :development, :test do
   # GOV.UK interpretation of rubocop for linting Ruby
   gem 'govuk-lint'
 
+  gem 'rspec_junit_formatter', '~> 0.4.1'
+
   # Debugging
   gem 'pry-byebug'
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,104 +6,137 @@ trigger:
 
 pr: none
 
-jobs:
-- job: TestsAndBuild
-  displayName: Run Tests And Build Image
+variables:
+  IMAGE_NAME: '$(dockerHubUserName)/get-help-to-retrain'
+  postgresUrl: postgres://test:test@localhost:5432/get-help-to-retrain
 
-  pool:
-    vmImage: 'ubuntu-16.04'
+resources:
+  containers:
+  - container: 'postgres'
+    image: 'postgres:11-alpine'
+    ports:
+      - 5432:5432/tcp
+    env:
+      POSTGRES_DB: 'get-help-to-retrain'
+      POSTGRES_USER: 'test'
+      POSTGRES_PASSWORD: 'test'
 
-  variables:
-  - group: get-help-to-retrain
-  - name: postgresImage
-    value: postgres:11-alpine
-  - name: postgresPassword
-    value: test
-  - name: postgresDatabaseUrl
-    value: postgres://postgres:test@postgres/get-help-to-retrain
-  - name: dockerRun
-    value: docker run --rm --link $(postgresImage) -e DATABASE_URL=$(postgresDatabaseUrl) -e RAILS_ENV=test -e RACK_ENV=test -e NODE_ENV=test $(dockerRegistry)/$(imageName):$(imageTag)
+stages:
+- stage: Build
+  displayName: Build image
 
-  steps:
-    - script: |
-        git_sha=$(git rev-parse --short HEAD)
-        echo "##vso[build.updatebuildnumber]$git_sha"
-        echo "##vso[task.setvariable variable=ImageTag;]$git_sha"
-      displayName: 'Set Image Tags'
+  jobs:
+  - job: Build
+    displayName: Build
+    pool:
+      vmImage: 'ubuntu-16.04'
 
-    - script: docker login -u $(dockerRegistry) -u $(dockerId) -p $pswd
-      env:
-        pswd: $(dockerPassword)
-      displayName: Docker Login
+    services:
+      postgres: postgres
 
-    - script: |
-        docker run --name=postgres -e POSTGRESS_PASSWORD=$(postgresPassword) -d $(postgresImage)
-      displayName: Launch Postgres
+    steps:
+      - script: |
+          git_sha=$(git rev-parse --short HEAD)
+          echo "##vso[build.updatebuildnumber]$git_sha"
+          echo "##vso[task.setvariable variable=ImageTag;]$git_sha"
+        displayName: 'Set Image Tags'
 
-    - script: |
-        docker pull $(dockerRegistry)/$(imageName):latest || true
-      displayName: Retrieve latest Docker build to use as cache
-      condition: ne(variables['Build.SourceBranch'], 'refs/heads/master')
+      - task: Docker@1
+        displayName: Docker Hub login
+        inputs:
+          command: login
+          containerregistrytype: Container Registry
+          dockerRegistryEndpoint: DFE-Digital Docker Hub
 
-    - script: |
-        docker build -f Dockerfile --cache-from=$(dockerRegistry)/$(imageName):latest -t $(dockerRegistry)/$(imageName):$(imageTag) .
-      displayName: Build Docker Image using Cache
-      condition: ne(variables['Build.SourceBranch'], 'refs/heads/master')
+      - script: docker pull $(IMAGE_NAME):latest || true
+        displayName: "Pull latest docker image to cache"
 
-    - script: |
-        docker build -f Dockerfile -t $(dockerRegistry)/$(imageName):$(imageTag) -t $(dockerRegistry)/$(imageName):latest .
-      displayName: Build Docker Image without Cache
-      condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
+      - task: Docker@1
+        displayName: Build docker image
+        inputs:
+          command: build
+          imageName: $(IMAGE_NAME):$(Build.BuildNumber)
+          dockerFile: Dockerfile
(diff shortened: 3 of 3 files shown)


mafonso:
Best with Multi Stage Pipelines enabled.

msp(1)

Gemfile
```diff
@@ -49,6 +49,8 @@ group :development, :test do
   # GOV.UK interpretation of rubocop for linting Ruby
   gem 'govuk-lint'
 
+  gem 'rspec_junit_formatter'
```
cwrw: can we version this gem?

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+          imageName: $(IMAGE_NAME):$(Build.BuildNumber)
+          dockerFile: Dockerfile
+          arguments: '--cache-from $(IMAGE_NAME):latest'
+
+      - task: Docker@1
+        displayName: Run Rubocop
+        inputs:
+          command: run
+          imageName: $(IMAGE_NAME):$(Build.BuildNumber)
+          containerCommand: bundle exec rubocop app config db lib spec --format clang
```
cwrw: can we keep the original command `bundle exec rubocop`, is there a reason you changed it?

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+          imageName: $(IMAGE_NAME):$(Build.BuildNumber)
+          containerCommand: bundle exec brakeman
+          runInBackground: false
+
+      - task: Docker@1
+        displayName: Run GovUK Sass linter
+        inputs:
+          command: Run an image
+          imageName: $(IMAGE_NAME):$(Build.BuildNumber)
+          containerCommand: bundle exec govuk-lint-sass app/webpacker/styles
```
cwrw: can we keep the original command as well `bundle exec govuk-lint-sass`

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+          runInBackground: false
+
+      - task: Docker@1
+        displayName: Database Setup
+        inputs:
+          command: run
+          imageName: $(IMAGE_NAME):$(Build.BuildNumber)
+          containerCommand: bundle exec rake db:create db:test:prepare
+          runInBackground: false
+          arguments: --net host -e DATABASE_URL=$(postgresUrl) -e RAILS_ENV=test -e RACK_ENV=test -e NODE_ENV=test
```
cwrw: the arguments `-e RAILS_ENV=test -e RACK_ENV=test -e NODE_ENV=test` need to be defined on all tests not just rspec and test db creation

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+
+      - task: Docker@1
+        displayName: Run Rspec tests
+        inputs:
+          command: Run an image
+          imageName: $(IMAGE_NAME):$(Build.BuildNumber)
+          volumes: $(Build.SourcesDirectory):/build
+          containerCommand: bundle exec rspec spec --format RspecJunitFormatter --out /build/TEST-rspec.xml
```
cwrw: no need to define `spec` in `bundle exec rspec spec`

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+          arguments: --net host -e DATABASE_URL=$(postgresUrl) -e RAILS_ENV=test -e RACK_ENV=test -e NODE_ENV=test
+
+      - task: PublishTestResults@2
+        condition: succeededOrFailed()
+        inputs:
+          testRunner: JUnit
+          testResultsFiles: '*.xml'
+          failedTaskOnFailedTest: true
+
+      - task: Docker@2
```
cwrw: does this get pushed if the tests fail?

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+        inputs:
+          command: login
+          containerregistrytype: Container Registry
+          dockerRegistryEndpoint: DFE-Digital Docker Hub
+
+      - script: docker pull $(IMAGE_NAME):latest || true
+        displayName: "Pull latest docker image to cache"
+
+      - task: Docker@1
+        displayName: Build docker image
```
cwrw: we are missing extra step of rebuilding without cache, could we add that back for master?

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+          failedTaskOnFailedTest: true
+
+      - task: Docker@2
+        displayName: Push new image with current tag
+        inputs:
+          command: push
+          repository: $(IMAGE_NAME)
+          tags: $(Build.BuildNumber)
+
+      - task: Docker@2
```
cwrw: couldn't find in docs, whats the @1, @2 etc. significance on tasks?

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+        displayName: Push tagged image (latest) if master
```
cwrw: is it possible to move those to a deploy job?

azure-pipelines.yml
```diff
@@ -6,104 +6,134 @@ trigger:
...
+
+      - task: Docker@1
+        displayName: Run Rspec tests
+        inputs:
+          command: Run an image
+          imageName: $(IMAGE_NAME):$(Build.BuildNumber)
+          volumes: $(Build.SourcesDirectory):/build
+          containerCommand: bundle exec rspec spec --format RspecJunitFormatter --out /build/TEST-rspec.xml
+          runInBackground: false
+          arguments: --net host -e DATABASE_URL=$(postgresUrl) -e RAILS_ENV=test -e RACK_ENV=test -e NODE_ENV=test
```
cwrw: 🌟

Gemfile
mafonso: I don't see why not. I just needed this to format the test results. 
But as I understand this is not baked into the image, it will be bundled in run time when we run the tests.

azure-pipelines.yml
mafonso: Yes, I changed it in order to try to get a more complete report that would show on the build test results, although not with much success.

Not being too versed in the ruby world this is how I saw it in documentation and examples from other teams.  This is also how it's is being used in the GovUK linter, and so was hoping for having similar report formats.  (Reference: https://github.com/alphagov/govuk-lint/issues/100) 

I return the question :)  is there a special reason we just run the simplified command?  What problems could arise from running it this way?

azure-pipelines.yml
mafonso: Once more, following the documentation in https://github.com/alphagov/govuk-lint and be specific rather then let it assume defaults

azure-pipelines.yml
mafonso: Including static code tests?  As I understand it no app code will be executed (unlike rspec).

I assumed it was only in all of them for convenience since we were using the same variable all over. Here we don't need to so I cleaned up where possible.

But happy to add them if those arguments will have any real effect.

azure-pipelines.yml
mafonso: I know. Just being specific even if it is the default.

azure-pipelines.yml
mafonso: Yes, as we want to have the test report for failed builds too.

azure-pipelines.yml
mafonso: We can, but it turns out that we were still using cache for master, as docker will always reuse unchanged layers. We can add it and force it have no cache if we add `--no-cache` to docker build.  But since we were building with cache implicitly anyway and it was not making any difference, I simplified.

azure-pipelines.yml
mafonso: Its the task Version.

azure-pipelines.yml
mafonso: The pushing is part of the build and it needs to be pushed from the same Job where we build it. 
Jobs start a fresh instance of the VM, so the previously built image would not be in the local docker engine.

Dockerfile
```diff
@@ -12,12 +12,17 @@ RUN apk add --update $BUILD_PACKAGES && \
 # Generate Assets
 WORKDIR /app
 COPY Gemfile Gemfile.lock .ruby-version ./
-RUN bundle install --clean --force --without "development"
+RUN bundle install --clean --force --without "development" \
+     && rm -rf /usr/local/bundle/cache/*.gem \
+     && find /usr/local/bundle/gems/ -name "*.c" -delete \
+     && find /usr/local/bundle/gems/ -name "*.o" -delete
+
```
mafonso: Removing the gem build files saves ~100MB

Gemfile
cwrw: We usually add versions to gems to prevent major versions being pulled in when we run `bundle install` thus causing any unwanted behaviour or breakage. You can add `gem 'rspec_junit_formatter', '~> 0.4.1'` instead here

azure-pipelines.yml
cwrw: Ok that's fair if we were always using the cache. Do you have an example of this happening?

azure-pipelines.yml
cwrw: GovUK linter was swapped out for rubocop for not giving us the coverage we need for linting. Also GovUK linter was being tied to specific folders (the ones you saw in the examples from other teams). We want rubocop to run on *all* files and folders, and the ones we want to exclude are specified in the .rubocop.yml file specifically. The logic for this should not be in the command but in the config we specify for linting, leaving the command as simple as possible. The way it's been added was intentional so I think we should revert back to the original

azure-pipelines.yml
cwrw: We might add styles in other folders, so I would like this command to also run on those, and keep it as general as possible. These are quite fast and not causing issues in terms of speed, so no need to be specific

azure-pipelines.yml
cwrw: There is no need to be specific as this is the default command in ruby, and doesn't affect performance and will never read from a different folder, so could you please change this back

azure-pipelines.yml
cwrw: I agree that we have to have the report, but I don't think we should be pushing a docker image to dockerhub if the tests fail

azure-pipelines.yml
cwrw: Yes there are many reasons to include those variables. We are building all the ENVs in the docker image under production, and when running any kind of test we have to change the ENVs to test, as that usually will cause surprising behaviour tied to config and inside those tests. Also the current libraries and gems are defined in the test section of the Gemfile. We added the `bundle test` section in our docker image for now as a *hack* until we separate them and do this correctly. But when we do the separation this will all start breaking as the ENV will be incorrect.

Gemfile
mafonso: Gem version pined

PR #77, opened 2019-07-12
--force is now mandatory on bundle install --clean

mafonso:
This PR fixes a now mandatory use of `--force` on `bundle install --clean`


mafonso:
image

PR #82, opened 2019-07-15
GET-216 Dashify all routes

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-216

* Dashify all routes
* Change query params to make sure search is preserved to "search" to
  avoid long urls

deployed and ready for testing: https://s108d01-dev-as.azurewebsites.net/

### DIFF
--- a/app/views/check_your_skills/_results_list.html.erb
+++ b/app/views/check_your_skills/_results_list.html.erb
@@ -2,7 +2,7 @@
   <% job_profiles.each do |job_profile| %>
     <li class="govuk-!-padding-bottom-3 govuk-!-padding-top-4">
       <h3 class="govuk-!-margin-0">
-        <%= link_to job_profile.name, job_profile_skills_path(job_profile.slug, check_your_skills_result: params[:name]), class: 'govuk-link no-text-decoration' %>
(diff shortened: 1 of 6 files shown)


tavy88:
@cwrw Failed spec alert.

app/views/check_your_skills/_results_list.html.erb
```diff
@@ -2,7 +2,7 @@
   <% job_profiles.each do |job_profile| %>
     <li class="govuk-!-padding-bottom-3 govuk-!-padding-top-4">
       <h3 class="govuk-!-margin-0">
-        <%= link_to job_profile.name, job_profile_skills_path(job_profile.slug, check_your_skills_result: params[:name]), class: 'govuk-link no-text-decoration' %>
+        <%= link_to job_profile.name, job_profile_skills_path(job_profile.slug, 'search' => params[:name]), class: 'govuk-link no-text-decoration' %>
```
smoothcontract: Any particular reason to change this from a symbol to a string? Same comment applies to change in `job_profiles_helper` above. I think a symbol looks nicer in both places and even if search query string is eventually passed back as a string this can still be accessed via `params[:search]` as params uses `ActiveSupport::HashWithIndifferentAccess`.

cwrw: sure I'll change to symbol, it was not really intentional was because I dashified the first param then decided on search

PR #84, opened 2019-07-16
GET-217 Landing page content changes

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-217

* add uncommited bundle change from 
https://github.com/DFE-Digital/get-help-to-retrain/pull/76/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R52


Gemfile.lock
```diff
@@ -369,6 +371,7 @@ DEPENDENCIES
   rails (~> 5.2.3)
   rails-i18n (~> 5.1)
   rspec-rails (~> 3.8)
+  rspec_junit_formatter (~> 0.4.1)
```
tavy88: I get this change now but I think for the future we should keep our commits atomic and PRs focused on what they say.

app/views/home/index.html.erb
```diff
@@ -3,18 +3,17 @@
...
+    <p class="govuk-body">Use this service to find out what other jobs you can do with your current skills. You can also find out how to get new skills that could help you find a better job.</p>
     <%= link_to 'Start now', task_list_path, class: 'govuk-button govuk-button--start govuk-!-margin-top-2 govuk-!-margin-bottom-8', role: 'button' %>
     <h2 class="govuk-heading-m">Before you start</h2>
     <p class="govuk-body">You’ll get the most out of this service if all of the following apply to you:</p>
     <ul class="govuk-list govuk-list--bullet">
       <li>you're employed</li>
-      <li>you do not have a degree</li>
-      <li>you're based in the Liverpool area</li>
+      <li>you don't have a degree</li>
+      <li>you’re based in the Manchester area</li>
```
tavy88: This symbol is not consistent with the one at line#11 or line#12 for instance?

Gemfile.lock
cwrw: Not sure that applies here, bundle has to run on the branch, and while running bundle it will produce this change, so we *have* to include it

tavy88: Yes but this gem has nothing to do with the actual feature `Landing page content changes` does it?

cwrw: no but this was fallout from the last pr we missed, and when running bundle on the branch because it was required to it included it. So I think there is no harm in including this

PR #88, opened 2019-07-17
Sample code - WIP

tavy88:
### Context

This is just a sample code implementation for https://dfedigital.atlassian.net/browse/GET-172 just to see what is possible with Smart Survey.


tavy88:
The Spike is completed and the code is no longer needed at this level.

PR #89, opened 2019-07-18
Add Javascript RSpec Feature Support and missing tests

cwrw:
* Set default Capybara driver to selenium headless chrome
* Add container for selenium standalone chrome to run as a service in CI
* When running in CI, set Capybara chrome driver to read from remote selenium chrome url running in container
* Pass in docker bridge ip to allow selenium chrome container access to the running instance
* Add missing JS tests for check your skills and explore occupations

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,6 +20,12 @@ resources:
       POSTGRES_DB: 'get-help-to-retrain'
       POSTGRES_USER: 'test'
       POSTGRES_PASSWORD: 'test'
+  - container: 'selenium'
+    image: 'selenium/standalone-chrome:3'
+    ports:
+      - 4444:4444/tcp
+    volumes:
+      - /dev/shm:/dev/shm
 
 stages:
 - stage: Build
@@ -33,13 +39,18 @@ stages:
 
     services:
       postgres: postgres
+      selenium: selenium
 
     steps:
       - script: |
           git_sha=$(git rev-parse --short HEAD)
           echo "##vso[build.updatebuildnumber]$git_sha"
-          echo "##vso[task.setvariable variable=ImageTag;]$git_sha"
         displayName: 'Set Image Tags'
+
+      - script: |
+          gateway_ip=$(docker network inspect bridge --format='{{range .IPAM.Config}}{{.Gateway}}{{end}}')
+          echo "##vso[task.setvariable variable=SeleniumRemoteAddress;]$gateway_ip"
+        displayName: 'Set Task Variables'
 
       - task: Docker@1
         displayName: Docker Hub login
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -33,14 +39,19 @@ stages:
 
     services:
       postgres: postgres
+      selenium: selenium
 
     steps:
       - script: |
           git_sha=$(git rev-parse --short HEAD)
           echo "##vso[build.updatebuildnumber]$git_sha"
-          echo "##vso[task.setvariable variable=ImageTag;]$git_sha"
```
cwrw: this isn't used anywhere (leftover from previous overhaul) so removed

tavy88: Should we wait for @mafonso?

cwrw: if this is relating to the pr review, I've pinged him to have a look. Regarding this change you can grep for `ImageTag` variable across this file and see that it isn't used. I added in an initial release a while back, but in @mafonso overhaul he is only using the `BuildNumber` instead

azure-pipelines.yml
```diff
@@ -33,14 +39,19 @@ stages:
...
         displayName: 'Set Image Tags'
 
+      - script: |
+          gateway_ip=$(docker network inspect bridge --format='{{range .IPAM.Config}}{{.Gateway}}{{end}}')
```
tavy88: Not sure I understand what this is trying to do?

spec/features/explore_occupations_spec.rb
```diff
@@ -65,4 +65,21 @@
 
     expect(page).to have_selector('ul.govuk-list li', count: 2)
   end
+
+  scenario 'cannot send search form with no input', js: true do
```
tavy88: Surely this test passes without the js tag as well. How do you know JS is actually running?

cwrw: Unfortunately it does not without the `js: true` setting as rack test does not include JS support. If you remove this setting the test will fail.  
Since we fallback to the non js solution, if we do disable the search form javascript, the test will pass, *but* we still need the `js: true` regardless

spec/features/check_your_skills_spec.rb
```diff
@@ -53,4 +53,33 @@
 
     expect(page).to have_selector('ul.govuk-list li', count: 2)
   end
+
+  scenario 'search form required field available when no js running' do
```
smoothcontract: Cool 👍

spec/features/check_your_skills_spec.rb
```diff
@@ -53,4 +53,33 @@
...
+    visit(check_your_skills_path)
+
+    expect(page).to have_selector('#search[required]')
+  end
+
+  scenario 'search form required field disabled by default', js: true do
```
smoothcontract: Do these `js` tags work if you just write them as `scenario 'search form required field disabled by default', :js do` - if so should we prefer that syntax?

cwrw:
This is a good step towards not including test dependencies in our main docker image. But the current status of our docker image is a bit more complex though as it has the gem dependencies and needs different type of work as well as discussion to move those out

PR #90, opened 2019-07-18
Create infrastructure for courses

tavy88:
### Context

Create the infrastructure to support the courses data.

### JIRA

https://dfedigital.atlassian.net/browse/GET-224

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,0 +1,5 @@
+class Course < ApplicationRecord
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+
+  validates_presence_of :title, :url, :provider, :topic
+end
--- a/db/migrate/20190718095913_create_courses.rb
+++ b/db/migrate/20190718095913_create_courses.rb
@@ -1,0 +1,18 @@
+class CreateCourses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :courses do |t|
+      t.string :title, null: false
+      t.string :provider, null: false
+      t.string :url, null: false
+      t.string :address_line_1
+      t.string :address_line_2
+      t.string :town
+      t.string :county
+      t.string :postcode
+      t.string :email
+      t.string :topic, index: true, null: false
+      t.string :phone_number
+      t.boolean :active, default: false
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_07_07_143458) do
+ActiveRecord::Schema.define(version: 2019_07_18_095913) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -22,6 +22,22 @@ ActiveRecord::Schema.define(version: 2019_07_07_143458) do
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.index ["slug"], name: "index_categories_on_slug", unique: true
+  end
+
+  create_table "courses", force: :cascade do |t|
+    t.string "title", null: false
+    t.string "provider", null: false
+    t.string "url", null: false
+    t.string "address_line_1"
+    t.string "address_line_2"
+    t.string "town"
+    t.string "county"
+    t.string "postcode"
+    t.string "email"
+    t.string "topic", null: false
+    t.string "phone_number"
+    t.boolean "active", default: false
+    t.index ["topic"], name: "index_courses_on_topic"
   end
 
   create_table "job_profile_categories", force: :cascade do |t|
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -1,0 +1,28 @@
+FactoryBot.define do
+  factory :course do
+    title { Faker::Educator.course_name }
+    provider { Faker::Educator.university }
+    url { Faker::Internet.url }
+    address_line_1 { Faker::Address.street_name }
+    address_line_2 { Faker::Address.street_address }
+    town { Faker::Address.city }
+    county { Faker::Address.state }
+    postcode { Faker::Address.postcode }
+    email { Faker::Internet.email }
+    phone_number { Faker::PhoneNumber.phone_number }
+    active { false }
+    topic { Faker::Educator.subject }
+
+    trait :enabled do
+      active { true }
+    end
+
+    trait :maths do
+      topic { 'maths' }
+    end
+
+    trait :english do
+      topic { 'english' }
+    end
+  end
+end
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -2,13 +2,11 @@ require 'rails_helper'
 
 RSpec.describe Course, type: :model do
   describe 'validations' do
-    let(:course) { build(:course, email: 'weirdemail@') }
-
     describe '#email' do
-      context 'when the email is invalid' do
-        it 'resource is not valid' do
-          expect(course).not_to be_valid
-        end
+      let(:course) { build(:course, email: 'weirdemail@') }
+
+      it 'is not valid' do
+        expect(course).not_to be_valid
       end
     end
   end
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -1,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe Course, type: :model do
+  describe 'validations' do
+    let(:course) { build(:course, email: 'weirdemail@') }
+
+    describe '#email' do
+      context 'when the email is invalid' do
+        it 'resource is not valid' do
+          expect(course).not_to be_valid
+        end
+      end
+    end
+  end
+end


app/models/course.rb
```diff
@@ -0,0 +1,3 @@
+class Course < ApplicationRecord
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+end
```
tavy88: I did not add validation for phone number yet as I want to see how they come in inside that doc.

db/migrate/20190718095913_create_courses.rb
```diff
@@ -0,0 +1,18 @@
+class CreateCourses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :courses do |t|
+      t.string :title
```
cwrw: can we add the required, null false etc. validations for all fields we expect to be required?

tavy88: I wouldn't go there yet as we don't know yet which will be optional and which not. We just assumed in the original discussion.

app/models/course.rb
```diff
@@ -0,0 +1,3 @@
+class Course < ApplicationRecord
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
```
smoothcontract: Don't we want mandatory validation for other attributes too (except for address_line_2)? Also email should be optional. Could omit validation entirely for the time being though to be honest, and address this later when we think about importing the courses data.

db/migrate/20190718095913_create_courses.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string :provider
+      t.string :url
+      t.string :address_line_1
+      t.string :address_line_2
+      t.string :town
+      t.string :county
+      t.string :postcode
+      t.string :email
+      t.string :subject
```
smoothcontract: Probably worth including an index on subject as we'll be selecting on those. Or perhaps consider making this an enum (with permitted values of `maths` and `english` for now)?

db/schema.rb
```diff
@@ -24,6 +24,21 @@
...
+    t.string "url"
+    t.string "address_line_1"
+    t.string "address_line_2"
+    t.string "town"
+    t.string "county"
+    t.string "postcode"
+    t.string "email"
+    t.string "subject"
+    t.string "phone_number"
+    t.boolean "active", default: true
```
smoothcontract: Maybe make the default `false` here - then we would need to explicitly enable any new courses before they appear to the user. That might be a better fit when we start adding courses in advance for new regions before fully rolling out to that region. Just a thought.

spec/factories/courses.rb
```diff
@@ -0,0 +1,24 @@
...
+    address_line_2 { Faker::Address.street_address }
+    town { Faker::Address.city }
+    county { Faker::Address.state }
+    postcode { Faker::Address.postcode }
+    email { Faker::Internet.email }
+    subject { Faker::Educator.subject }
+    phone_number { Faker::PhoneNumber.phone_number }
+    active { true }
+
+    trait :disabled do
```
smoothcontract: If the default is swapped to `false` then invert this trait also

spec/factories/courses.rb
```diff
@@ -0,0 +1,24 @@
+FactoryBot.define do
+  factory :course do
+    title { Faker::Educator.course_name }
```
smoothcontract: Love Faker! ❤️

spec/factories/courses.rb
```diff
@@ -0,0 +1,24 @@
...
+    email { Faker::Internet.email }
+    subject { Faker::Educator.subject }
+    phone_number { Faker::PhoneNumber.phone_number }
+    active { true }
+
+    trait :disabled do
+      active { false }
+    end
+
+    trait :with_invalid_email do
```
cwrw: a trait is usually used when we expect to reuse a factory, I don't believe this will be the case for invalid emails

app/models/course.rb
tavy88: Yeah, I did not add yet validations on the rest of the fields as we don't know yet which will be mandatory and which not. We just assumed but we haven't seen the data yet?

db/migrate/20190718095913_create_courses.rb
tavy88: We could add an index yes.

db/schema.rb
tavy88: But if we do a mass import we want to make all those active as a separate process? I would that most will be active? Hence the amount of courses we need to mark as inactive is significantly less?

spec/factories/courses.rb
```diff
@@ -0,0 +1,24 @@
...
+      email { 'weirdemail@' }
+    end
+  end
+end
```
smoothcontract: Could you add traits for `:maths` and `:english` too please? I suspect we'll be using those a fair bit in our tests as those are the two initial course types.

db/schema.rb
tavy88: If we know for sure they send us that information then we can use a default inactive, but again not 100% sure that will be available.

db/migrate/20190718095913_create_courses.rb
cwrw: the bare minimum like title, url, provider, subject surely should required?

spec/factories/courses.rb
tavy88: Yes, but I thought it is out of scope for this ticket as we won't be doing anything with them yet. Those would be added as part of the tickets work. Remember this is a Spike.

db/schema.rb
cwrw: and null false as well :D

smoothcontract:
Another thought, could you add some example courses to the seed service too please? Something like this (assuming extra traits for each course type):

```ruby
  def seed_courses
    10.times { FactoryBot.create :course, :maths }
    10.times { FactoryBot.create :course, :english }
  end
```

spec/factories/courses.rb
tavy88: Surely we will need to reuse this when we do the import functionality right?

app/models/course.rb
smoothcontract: I think we can make some reasonable assumptions based on how the UI will look with missing data. Obviously the course title is required or nothing is displayed. That said I'm happy enough for all of the validation to be added later when we think about importing the data if that's easier to think about.

db/schema.rb
smoothcontract: All good things to consider - we don't know how the data will be imported yet, but I'd prefer to err on the side of safety.

spec/factories/courses.rb
cwrw: sure that's fine, but generally its good to add traits when we start seeing reuse, and avoid doing that in the beginning

spec/models/course_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+RSpec.describe Course, type: :model do
+  describe 'validations' do
+    let(:course) { build(:course, :with_invalid_email) }
+
+    context 'when the email is invalid' do
+      it 'does not create the resource' do
+        course.save
+
+        expect(Course.count).to be_zero
```
smoothcontract: Could combine these two lines with `expect(course.save).to be_false`

spec/factories/courses.rb
```diff
@@ -0,0 +1,28 @@
...
+    email { Faker::Internet.email }
+    phone_number { Faker::PhoneNumber.phone_number }
+    active { false }
+    topic { Faker::Educator.subject }
+
+    trait :enabled do
+      active { true }
+    end
+
+    trait :math do
```
smoothcontract: maths instead of math please 🇬🇧

spec/models/course_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  describe 'validations' do
+    let(:course) { build(:course, email: 'weirdemail@') }
+
+    context 'when the email is invalid' do
+      it 'does not create the resource' do
+        expect(course.save).to be false
+      end
+    end
+
+    it { is_expected.to validate_presence_of(:title) }
```
smoothcontract: A bit lukewarm about these matchers to be honest. Do we need to add specs for these and associated extra gem?

db/schema.rb
```diff
@@ -24,6 +24,29 @@
...
+    t.string "county"
+    t.string "postcode"
+    t.string "email"
+    t.string "topic", null: false
+    t.string "phone_number"
+    t.boolean "active", default: false
+    t.index ["topic"], name: "index_courses_on_topic"
+  end
+
+  create_table "flipflop_features", force: :cascade do |t|
```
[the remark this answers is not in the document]
tavy88: oops

spec/models/course_spec.rb
tavy88: I think so yes. Especially if we are doing TDD.

cwrw: agreed, I think generally I don't test validations as its testing the framework unless custom, also RSpec provides appropriate matchers as well https://www.rubydoc.info/gems/rspec-rails/RSpec/Rails/Matchers#be_valid-instance_method

tavy88: I do feel this is a personal preference though.

tavy88: > agreed, I think generally I don't test validations as its testing the framework unless custom, also RSpec provides appropriate matchers as well https://www.rubydoc.info/gems/rspec-rails/RSpec/Rails/Matchers#be_valid-instance_method

I don't think it's the same case for want I do though. I am testing that the validation is in place rather than testing the actual functionality.

smoothcontract: I think these specs for every validation rule are overly explicit. I’ve not seen a Rails project where this level of detail is needed. A single case to check that a course is valid with all attributes populated is sufficient here and removes the need for extra shoulda gem.

tavy88: We were actually doing these a lot at my previous place. Anyway moving forward I feel we're wasting too much time arguing about this. I can take them out if it makes you guys happy.

PR #91, opened 2019-07-19
Show courses page

tavy88:
### Context

Show courses results page by course topic.

### Screenshots

Screen Shot 2019-07-22 at 11 25 32

Screen Shot 2019-07-21 at 14 41 03

### JIRA

https://dfedigital.atlassian.net/browse/GET-250

### How to test

1. Run CoursesService.seed
2. Visit http://localhost:3000/courses/maths and http://localhost:3000/courses/english


app/helpers/application_helper.rb
```diff
@@ -4,15 +4,16 @@ def page_title(key)
...
-    )
+    links = [
+      previous_pages.map { |title, link|
+        content_tag(:li, nil, class: 'govuk-breadcrumbs__list-item') {
+          link_to(title, link, class: 'govuk-breadcrumbs__link')
+        }
+      }
+    ]
+
+    links << content_tag(:li, current_page, class: 'govuk-breadcrumbs__list-item') if current_page
```
cwrw: Im not sure I understand this change, we should always have a current page, rather than a link

app/views/courses/index.html.erb
```diff
@@ -0,0 +1,24 @@
+<% page_title :courses_index %>
+
+<% content_for :breadcrumb do
+    generate_breadcrumbs(
+      nil,
```
cwrw: I'm having a look at the prototype, and this seems like a mistake: https://get-help-to-retrain-prototype.herokuapp.com/main/release2_e2e_v1/maths_courses_near_me

Can you confirm with Steve what the wording/breadcrumb should be

app/views/shared/_contact_us.html.erb
```diff
@@ -1,9 +1,13 @@
 <div class="govuk-grid-column-one-third">
-  <div class="govuk-related-items govuk-!-padding-bottom-4 top-border__container">
+  <div class="govuk-related-items top-border__container">
     <h2 class="govuk-heading-m"><%= t('contact_us.title') %></h2>
-    <p class="govuk-body-m govuk-!-margin-bottom-0">
-      <%= t('contact_us.body') %> <%= link_to t('contact_us.telephone'), t('contact_us.telephone_link'), class: 'govuk-!-padding-left-1 govuk-link' %>
-    </p>
-    <p class="govuk-body-m govuk-!-margin-bottom-0"><%= t('contact_us.telephone_times') %></p>
+    <ul class="govuk-list">
```
cwrw: why is this added to a list?

spec/factories/courses.rb
```diff
@@ -2,12 +2,12 @@
...
     title { Faker::Educator.course_name }
     provider { Faker::Educator.university }
-    url { Faker::Internet.url }
+    url { 'http://example.com/' }
     address_line_1 { Faker::Address.street_name }
     address_line_2 { Faker::Address.street_address }
     town { Faker::Address.city }
     county { Faker::Address.state }
-    postcode { Faker::Address.postcode }
+    postcode { ['EC1A 1AA', 'EC1A 1JJ', 'EC1A 1JN', 'EC1A 2AH', 'EC1P 1AB', 'EC1P 1EW', 'EC1P 1NY'].sample }
```
cwrw: can you not set Faker to generate UK postcodes rather than this defined set?

spec/controllers/courses_controller_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.describe CoursesController, type: :controller do
```
cwrw: can we add feature specs for the courses pages as well?

spec/factories/courses.rb
tavy88: I wanted to be specific so you can seed your data accordingly and get the desired distance. Will keep it like this for now.

app/views/shared/_contact_us.html.erb
tavy88: Ask Steve :)

app/views/courses/index.html.erb
tavy88: I followed the design :) I will have a chat with him.

app/helpers/application_helper.rb
tavy88: Not according to the prototype?

spec/controllers/courses_controller_spec.rb
tavy88: What would you be testing exactly?

cwrw: We should be testing  all journeys in feature tests as an integration test, I would test that the page loads with a list of Courses (can count that list)

app/views/courses/index.html.erb
smoothcontract: Does seem like inconsistent wording 🤔

spec/controllers/courses_controller_spec.rb
smoothcontract: I thought we agreed to add feature specs rather than controller tests? Controller tests are basically deprecated. Could you change this to a feature spec instead please - and make sure that we see at least some course content on the page.

spec/factories/courses.rb
smoothcontract: This seems okay for now.

app/helpers/application_helper.rb
tavy88: It's not unrelated because it breaks the breadcrumbs on the prototype as it is now because we don't have a current page?

spec/controllers/courses_controller_spec.rb
tavy88: This was part of TDD.

config/locales/en.yml
```diff
@@ -5,12 +5,13 @@ en-GB:
     task_list_home: 'Home: Get help to retrain'
     check_your_skills: Check your existing skills
     explore_occupations: Explore occupations
-    find_training_courses: Find and apply to training courses near you
+    find_training_courses: Training courses near you
```
tavy88: I didn't change this key as it will mess up our tracked page titles right now. I could change this so it looks less duplicated than find training key at line#14.

app/views/pages/find_training_courses.html.erb
```diff
@@ -1,7 +1,7 @@
 <% page_title :pages_find_training_courses %>
 <% content_for :breadcrumb do
     generate_breadcrumbs(
-      t('breadcrumb.find_training_courses'),
+      t('breadcrumb.find_training'),
```
cwrw: :sc

app/views/shared/_contact_us.html.erb
```diff
@@ -1,9 +1,9 @@
 <div class="govuk-grid-column-one-third">
-  <div class="govuk-related-items govuk-!-padding-bottom-4 top-border__container">
```
cwrw: are we sure we don't need this anymore? (padding and margins) as I recall this was needed for other pages?

tavy88: Hmm, I can add it back.

app/views/shared/_contact_us.html.erb
```diff
@@ -4,6 +4,6 @@
     <p class="govuk-body-m govuk-!-margin-bottom-0">
       <%= t('contact_us.body') %> <%= link_to t('contact_us.telephone'), t('contact_us.telephone_link'), class: 'govuk-!-padding-left-1 govuk-link' %>
     </p>
-    <p class="govuk-body-m govuk-!-margin-bottom-0"><%= t('contact_us.telephone_times') %></p>
+    <p class="govuk-body-s work-hours"><%= t('contact_us.telephone_times') %></p>
```
cwrw: Can we keep the margin, so this is `govuk-body-s govuk-!-margin-bottom-0 muted-text` instead

PR #92, opened 2019-07-19
GET-123 Add Geocoding to Courses

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-123

* Add latitude longitude columns to courses table to allow geocoding of postcodes of courses
* Validate postcode on course model against uk postcodes
* Format postcode on course model against uk postcodes to be in capital letter format, and a space in the middle
* Add callback to geocode the postcodes after validation on courses model. This means latitude and longitude values will be populated by the Geocoder gem
* Add method to return courses ordered by distance to an address
* Distance can be defined and defaults to 5 miles
* Validate address before geocoding and performing a geospatial
  query to get all courses within the distance defined
* Return custom errors for invalid address and Geocoder API errors  so we can create validations on the FE
* Allow Geocoder to return all errors rather than hide them behind nil values so we have the correct messaging on the FE

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,6 +1,15 @@
 class Course < ApplicationRecord
   validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+  validates :postcode, presence: true, postcode: true
+  validates_presence_of :title, :url, :provider, :topic
 
-  validates_presence_of :title, :url, :provider, :topic, :postcode
+  geocoded_by :postcode
+  after_validation :format_postcode
+  after_validation :geocode, if: -> { postcode_changed? && !latitude_changed? && !latitude_changed? }
 
+  def format_postcode
+    return unless postcode
+
+    self.postcode = UKPostcode.parse(postcode).to_s
+  end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,4 +1,7 @@
 class Course < ApplicationRecord
+  InvalidAddressError = Class.new(StandardError)
+  GeocoderAPIError = Class.new(StandardError)
+
   validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
   validates :postcode, presence: true, postcode: true
   validates_presence_of :title, :url, :provider, :topic
@@ -6,6 +9,23 @@ class Course < ApplicationRecord
   geocoded_by :postcode
   after_validation :format_postcode
   after_validation :geocode, if: -> { postcode_changed? && !latitude_changed? && !latitude_changed? }
+
+  def self.find_courses_near(address:, distance: 5)
+    return none unless address.present?
+
+    coordinates = Geocoder.coordinates(parse_postcode(address))
+    near(coordinates, distance, units: :mi, order: :distance)
+  rescue SocketError, Timeout::Error, Geocoder::Error => e
+    Rails.logger.error("Geocoder API error: #{e.message}")
+    raise GeocoderAPIError
+  end
+
+  def self.parse_postcode(address)
+    uk_postcode = UKPostcode.parse(address)
+    raise InvalidAddressError unless uk_postcode.full_valid?
+
+    uk_postcode.to_s
+  end
 
   def format_postcode
     return unless postcode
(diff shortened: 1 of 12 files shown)


app/models/course.rb
```diff
@@ -1,5 +1,35 @@
...
+  InvalidAddressError = Class.new(StandardError)
+  GeocoderAPIError = Class.new(StandardError)
 
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+  validates :postcode, presence: true, postcode: true
   validates_presence_of :title, :url, :provider, :topic
+
+  geocoded_by :postcode
+  after_validation :format_postcode
+  after_validation :geocode, if: -> { postcode_changed? && !latitude_changed? && !latitude_changed? }
```
smoothcontract: Last term should be `!longitude_changed?` ?

app/models/course.rb
```diff
@@ -1,5 +1,35 @@
...
+
+  def self.find_courses_near(address:, distance: 5)
```
smoothcontract: We'll also need to scope courses by topic - can that be included as another search term?

smoothcontract: `address` isn't perhaps the best variable name here - is it because `postcode` is a reserved word? If so maybe something more explicit like `search_postcode`.

spec/rails_helper.rb
```diff
@@ -20,6 +21,7 @@
   vcr.hook_into :webmock
   vcr.ignore_localhost = true
   vcr.configure_rspec_metadata!
+  vcr.allow_http_connections_when_no_cassette = true
```
smoothcontract: Perhaps you can talk through why this change is needed? Can we not use VCR to capture the API traffic to postcodes.io for unit tests?

cwrw: Ah this is kind of breaking locally because of chrome/js tests and not this pr. So basically its unrelated to Geocoder. We are setting the location to `test` in RSpec so all is good there and its being mocked see `Geocoder::Lookup::Test.add_stub`

smoothcontract: okay 👍

app/models/course.rb
```diff
@@ -8,21 +8,24 @@ class Course < ApplicationRecord
...
-    return none unless address.present?
+  def self.find_courses_near(search_postcode:, topic: nil, distance: 5)
+    return none unless search_postcode.present?
 
-    coordinates = Geocoder.coordinates(parse_postcode(address))
-    near(coordinates, distance, units: :mi, order: :distance)
+    coordinates = Geocoder.coordinates(parse_postcode(search_postcode))
+
+    courses = near(coordinates, distance, units: :mi, order: :distance)
+    courses = courses.where(topic: topic) if topic.present?
```
smoothcontract: This would be nice to split out into a scope, that accepts blank topic param something like `courses = courses.by_topic(topic)` here

```ruby
  scope :by_topic, ->(topic = nil) { topic.present? ? where(topic: topic) : all }
```

smoothcontract: alternatively, if you are just returning an ActiveRecord result set here, you could externalise the `by_topic` part and chain it to the results returned here in the calling controller.

smoothcontract: i.e. from courses controller...

```ruby
@courses = Course.find_courses_near(search_postcode: 'FOO').by_topic(:maths)
```

app/models/course.rb
```diff
@@ -1,5 +1,38 @@
...
 
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+  validates :postcode, presence: true, postcode: true
   validates_presence_of :title, :url, :provider, :topic
+
+  geocoded_by :postcode
+  after_validation :format_postcode
+  after_validation :geocode, if: -> { postcode_changed? && !latitude_changed? && !longitude_changed? }
+
+  def self.find_courses_near(search_postcode:, topic: nil, distance: 5)
```
smoothcontract: Looking at this again, `postcode` should work fine as an argument name as this is a class method - `postcode` is only mapped to the model attribute within an instance method.

PR #93, opened 2019-07-22
Pipeline v2

mafonso:

This PR makes use of the new Azure multi-stage pipelines so we can manage deployment from the azure-pipelines.yaml.
https://devblogs.microsoft.com/devops/whats-new-with-azure-pipelines/

This refactoring also changed the way deployment variables are passed when deploying to environments. We moved from inline overriding of params to a per environment config file which is versioned.
This was possible because secrets were moved into Azure KeyVault and they are now just a reference.

### Guidance to review

ToDo:   While testing this change I could see some value in having a common parameter file that applies to *all* envs, so that things are a bit more DRY

Config files for upper environments will be added on separate PRs after migration takes place.

I appreciate that the first stage is a bit redundant now and it's making the whole process a couple of minutes longer.  This is to be optimized later (cache bundler?) but the placeholder has value now and still provides a quicker feedback cycle if static code tests fail.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,6 +22,41 @@ resources:
       POSTGRES_PASSWORD: 'test'
 
 stages:
+- stage: Checks
+  displayName: Preflight checks
+
+  jobs:
+    - job: Tests
+      displayName: Rubocop tests
+      pool:
+        vmImage: 'ubuntu-16.04'
+
+      steps:
+        - task: UseRubyVersion@0
+          inputs:
+            versionSpec: '>= 2.6.2'
+            addToPath: true # Optional
+
+        # Todo: Look for support in the build box for the same ruby minor version.
+        - script: |
+            echo "2.6.2" > .ruby-version
+            sudo apt-get install libpq-dev
+            gem install bundler
+            bundle install --retry=3 --jobs=4
+          displayName: 'bundle install'
+
+        - script: |
+            bundle exec rubocop --format clang
+          displayName: 'rubocop'
+
+        - script: |
+            bundle exec brakeman
+          displayName: 'brakeman'
+
+        - script: |
+            bundle exec govuk-lint-sass app/webpacker/styles
+          displayName: 'GovUK lint'
+
 - stage: Build
   displayName: Build image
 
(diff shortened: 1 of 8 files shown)


azure-pipelines.yml
```diff
@@ -22,6 +22,40 @@ resources:
...
+        vmImage: 'ubuntu-16.04'
+
+      steps:
+        - task: UseRubyVersion@0
+          inputs:
+            versionSpec: '>= 2.6.2'
+            addToPath: true # Optional
+
+        - script: |
+            echo "2.6.2" > .ruby-version
```
mafonso: Ruby 2.6.3 is not available on the box atm.
These tests are still done on the built image later as they only consume 6s extra, but this should give a quicker feedback cycle for now.

cwrw: If ruby isn't available Id rather we keep these as is for now until we speed up the container

smoothcontract: This seems like a good compromise - but it might be worth annotating this with a FIXME or TODO so that if 2.6.3 becomes available at some point this can be unpicked and we don't forget.

mafonso: We still get some value from this earlier feedback. 
Also we don't have an ETA to have a newer version of ruby on the box, so this will be the setup we will have, apart from maybe some more caching.  
But small steps, it's better that this get is in and improve it later that aim for the perfect thing in one go.
I've discussed the ruby version with @smoothcontract and he was happy with the lower minor ruby version for this static tests.  Rspec still runs later with the proper version.

azure-pipelines.yml
```diff
@@ -22,6 +22,40 @@ resources:
...
+            sudo apt-get install libpq-dev
+            gem install bundler
+            bundle install --retry=3 --jobs=4
```
smoothcontract: I think you can get away with only installing stuff in the development group here - brakeman, linter and rubocop are all in there. No need to install the other test stuff or anything for production. You can add `--with=development` flag for this.

azure-pipelines.yml
```diff
@@ -140,3 +174,69 @@ stages:
...
+    - job: CI
+      displayName: Deploy to CI
+      pool:
+        vmImage: 'vs2017-win2016'
+
+      variables:
+        - name: resourceGroupName
+          value: s108d01-ci
+        - name: azureSubscription
+          value: azdo.pipelines.cip.S108D.arm10a1d6de-2cee-4f62-b5fa-2cf08d05a1dc
```
cwrw: can these values be variables?

azure/config/ci.parameters.json
```diff
@@ -0,0 +1,54 @@
...
+        },
+        "secretKeyBase": {
+          "reference": {
+            "keyVault": {
+              "id": "/subscriptions/f35bd249-3028-48bd-9aab-d00e9d1a56b7/resourceGroups/s108d01-shared/providers/Microsoft.KeyVault/vaults/s108d01-shared-kv-01"},
+            "secretName": "ci-secretKeyBase"
+          }
+        },
+        "securityAlertEmail": {
+            "value": "miguel.afonso@digital.education.gov.uk"
```
smoothcontract: Are you happy to make this public? Maybe better moved to the vault 🤔

azure/config/dev2.parameters.json
```diff
@@ -0,0 +1,54 @@
...
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "serviceName": {
+            "value": "dev"
+        },
+        "databaseName": {
+            "value": "ghtr"
+        },
+        "databaseUsername": {
+            "value": "app"
```
smoothcontract: I think would be better to move the username to the vault too, otherwise an attacker already knows a valid user account and we rely on password alone for security.

azure-pipelines.yml
mafonso: Those are service connection references. 
They are specific to the the environment hence being defined as variables in the job level.   
I tried to use references to these on the following tasks but variable interpolation seems not to be consistent when the pipeline is evaluated, so there is still a bit of repetition. 
(I'm following discussions on microsoft github looking for improvements to this)

azure-pipelines.yml
smoothcontract: @cwrw I'm pretty happy that 2.6.2 is close enough to 2.6.3 to get run linting here without any excessive risk. And it seems useful to get some early feedback on the builds. However if we want to extend this approach and also run specs then I think we'd need a better solution so that the ruby versions are aligned.

azure-pipelines.yml
mafonso: I'll add the TODO comment.
Yes, I see room for improvment on this first stage. I just wanted to get a stage placeholder without interference to the current workflow. 
I see the end goal being build the docker image for "production" only and this stage "test" only, but I leave that for you ruby devs :)

azure/config/ci.parameters.json
mafonso: I'd feel better if we had a team address. 
But fair point, I can move it to the vault to avoid spam.

azure/config/dev2.parameters.json
mafonso: Fair point.

azure/config/ci.parameters.json
mafonso: I might leave some of this common values in a variable group in the Azure Devops portal. The vaults are per environment family and like to avoid repeating common config on all of them for now.

azure-pipelines.yml
cwrw: From my side, I'm strongly against running our tests against the wrong Ruby version. Moving those out doesn't really help us much anyway, and I'm happy to wait for improving our docker build time to actually use the cache (I've got something that might work) so I'd rather not do this and risk having surprising errors/behaviour

mafonso: It wont be causing surprising errors/behaviour  because it's just a first pass at those tests. In the unlikely event one minor version down could be problematic the very same test also runs after the build with the proper version.
This change is zero risk while introducing this stage in the pipeline which I really want to bring in at this stage. 
There is obviously lots of room to improve the docker build and this very same stage, but let's do in small incremental steps rather then aiming to do it an in a PR that is meant to just change the process.

azure/config/ci.parameters.json
mafonso: Addressed.

azure/config/dev2.parameters.json
mafonso: Addressed

azure-pipelines.yml
mafonso: ToDo registered

mafonso:
> If I'm understanding this correctly, you're building a different image now just for preflight checks? If so then a few things could be removed from the bundle. Also I think would be good practice to move any user names (as well as passwords) to the vault to reduce the attack surface. Otherwise looks promising.

Not building a new image, just running it on the CI box directly prior to the actual build. 
Sensitive data was offloaded to KeyVault, leaving just references behind.

PR #95, opened 2019-07-22
GET-123 Add postcode search on courses page

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-123

* return no courses if postcode not given
* When searching for postcode return courses sorted by distance
* Return correct wording if postcode not valid, or api errors
* Return 0 results if user has a postcode too far or if they entered
wrong course

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -8,20 +8,16 @@ class CoursesController < ApplicationController
     ).map { |c| CourseDecorator.new(c) }
   rescue CourseGeospatialSearch::GeocoderAPIError
     flash.now[:error] = t('courses.api_down_error')
-    @courses = all_courses
+    @courses = Course.none
   rescue CourseGeospatialSearch::InvalidPostcodeError
     flash.now[:error] = t('courses.invalid_postcode_error')
-    @courses = all_courses
+    @courses = Course.none
   end
 
   private
 
   def handle_missing_courses
     redirect_to task_list_path if Course.count.zero?
-  end
-
-  def all_courses
-    Course.all.map { |c| CourseDecorator.new(c) }
   end
 
   def courses_params
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -1,10 +1,17 @@
 class CoursesController < ApplicationController
   before_action :handle_missing_courses
 
-  def index
-    @courses = Course.where(topic: params[:topic_id]).map { |c| CourseDecorator.new(c) }
-
-    redirect_to '/404' unless @courses.any?
+  def index # rubocop:disable Metrics/AbcSize
+    @courses = Course.find_courses_near(
+      postcode: params[:postcode],
+      topic: params[:topic_id]
+    ).map { |c| CourseDecorator.new(c) }
+  rescue CourseGeospatialSearch::GeocoderAPIError
+    flash.now[:error] = t('courses.api_down_error')
+    @courses = Course.all.map { |c| CourseDecorator.new(c) }
+  rescue CourseGeospatialSearch::InvalidPostcodeError
+    flash.now[:error] = t('courses.invalid_postcode_error')
+    @courses = Course.all.map { |c| CourseDecorator.new(c) }
   end
 
   private
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -3,20 +3,28 @@ class CoursesController < ApplicationController
 
   def index # rubocop:disable Metrics/AbcSize
     @courses = Course.find_courses_near(
-      postcode: params[:postcode],
-      topic: params[:topic_id]
+      postcode: courses_params[:postcode],
+      topic: courses_params[:topic_id]
     ).map { |c| CourseDecorator.new(c) }
   rescue CourseGeospatialSearch::GeocoderAPIError
     flash.now[:error] = t('courses.api_down_error')
-    @courses = Course.all.map { |c| CourseDecorator.new(c) }
+    @courses = all_courses
   rescue CourseGeospatialSearch::InvalidPostcodeError
     flash.now[:error] = t('courses.invalid_postcode_error')
-    @courses = Course.all.map { |c| CourseDecorator.new(c) }
+    @courses = all_courses
   end
 
   private
 
   def handle_missing_courses
     redirect_to task_list_path if Course.count.zero?
+  end
+
+  def all_courses
+    Course.all.map { |c| CourseDecorator.new(c) }
+  end
+
+  def courses_params
+    params.permit(:postcode, :topic_id)
   end
 end
--- a/app/models/course_geospatial_search.rb
+++ b/app/models/course_geospatial_search.rb
@@ -11,7 +11,7 @@ class CourseGeospatialSearch
   end
 
   def find_courses
-    return Course.all unless postcode.present?
+    return Course.none unless postcode.present?
 
     courses.near(coordinates, distance, units: :mi, order: :distance)
   end
--- a/app/models/course_geospatial_search.rb
+++ b/app/models/course_geospatial_search.rb
@@ -11,7 +11,7 @@ class CourseGeospatialSearch
   end
 
   def find_courses
-    return Course.none unless postcode.present?
+    return Course.all unless postcode.present?
 
     courses.near(coordinates, distance, units: :mi, order: :distance)
   end
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -25,7 +25,7 @@
               <span class="govuk-visually-hidden">Error:</span> <%= flash.now[:error] %>
             </span>
           <% end %>
-          <%= text_field(nil, :postcode, value: params[:postcode], class: "search-input govuk-input govuk-!-width-one-half #{'govuk-input--error' if flash.now[:error]}", required: true) %>
+          <%= text_field_tag(:postcode, params[:postcode], class: "search-input govuk-input govuk-!-width-one-half #{'govuk-input--error' if flash.now[:error]}", required: true) %>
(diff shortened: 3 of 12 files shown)


app/controllers/courses_controller.rb
```diff
@@ -1,10 +1,17 @@
 class CoursesController < ApplicationController
   before_action :handle_missing_courses
 
-  def index
-    @courses = Course.where(topic: params[:topic_id]).map { |c| CourseDecorator.new(c) }
-
-    redirect_to '/404' unless @courses.any?
+  def index # rubocop:disable Metrics/AbcSize
+    @courses = Course.find_courses_near(
```
cwrw: this is not great, improve and look into active model for address validation

spec/controllers/courses_controller_spec.rb
```diff
@@ -12,16 +12,6 @@
       end
     end
 
-    context 'when a non-existing course topic is accessed' do
```
cwrw: add this back and guard against this behaviour

app/views/courses/index.html.erb
```diff
@@ -10,11 +10,40 @@
...
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl"><%= params[:topic_id].titleize %> courses near me</h1>
     <div class="govuk-grid-column-full govuk-!-padding-0">
-      <%= render partial: 'course', collection: @courses %>   
+      <%= form_tag courses_path, method: 'get', id: 'courses-search' do %>
+        <div class="govuk-form-group <%='govuk-form-group--error' if flash.now[:error] %>">
+          <label class="govuk-label" for="postcode">
+            Postcode
+          </label>
+          <% if flash.now[:error] %>
```
cwrw: improve how this integrated into page

app/views/courses/index.html.erb
```diff
@@ -10,11 +10,40 @@
     )
   end
 %>
+
```
cwrw: ✂️

spec/features/find_training_courses_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    find('.search-button-results').click
+
+    expect(page).to have_selector('ul.govuk-list li', count: 1)
+  end
+
+  scenario 'User cannot find training courses with non existing topics' do
+    create(:course, topic: 'maths')
+
+    visit(courses_path(topic_id: 'history'))
+    fill_in('postcode', with: 'NW6 8ET')
```
cwrw: remove the next 2 lines so test fails

app/controllers/courses_controller.rb
```diff
@@ -1,10 +1,17 @@
...
+      postcode: params[:postcode],
+      topic: params[:topic_id]
+    ).map { |c| CourseDecorator.new(c) }
+  rescue CourseGeospatialSearch::GeocoderAPIError
+    flash.now[:error] = t('courses.api_down_error')
+    @courses = Course.all.map { |c| CourseDecorator.new(c) }
```
cwrw: check if this is the wanted behaviour

app/models/course_geospatial_search.rb
```diff
@@ -11,7 +11,7 @@ def initialize(postcode:, topic:, distance:)
   end
 
   def find_courses
-    return Course.none unless postcode.present?
+    return Course.all unless postcode.present?
```
smoothcontract: This does seem like slightly odd behaviour - but we didn't cover off this case in the ticket requirements. so lets review with stakeholders and decide how to move forward.

app/webpacker/packs/search-form.js
```diff
@@ -1,9 +1,9 @@
 function SearchForm () {
   this.start = function () {
-    const $jobProfileName = document.querySelector('#search');
-    const $jobProfileSearchForm = document.querySelector('#job-profile-search');
+    const $searchField = document.querySelector('#search, #postcode');
+    const $searchForm = document.querySelector('#job-profile-search, #courses-search');
```
smoothcontract: Nice reuse 👍

config/routes.rb
```diff
@@ -11,7 +11,7 @@
   get 'find-training-courses', to: 'pages#find_training_courses'
   get 'next-steps', to: 'pages#next_steps'
 
-  get 'courses/:topic_id', to: 'courses#index'
+  resources :courses, path: 'courses/:topic_id', only: %[index], constraints: { topic_id: /maths|english/ }
```
smoothcontract: 👍

spec/controllers/courses_controller_spec.rb
smoothcontract: this spec should pass unchanged now that the route constraint is in place

spec/features/find_training_courses_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    create_list(:course, 2, topic: 'maths')
+    visit(courses_path(topic_id: 'maths'))
+
+    expect(page).to have_selector('ul.govuk-list li', count: 2)
+  end
+
+  scenario 'User can find training courses near them' do
+    Geocoder::Lookup::Test.add_stub(
+      'NW6 8ET', [{ 'coordinates' => [0.1, 1] }]
+    )
```
smoothcontract: Neat!

app/views/courses/index.html.erb
cwrw: had a look and it's too early to move to form helpers or use the gov uk form builder gem. If we encounter validation errors again we should move this out to a form helper

spec/controllers/courses_controller_spec.rb
```diff
@@ -13,12 +13,8 @@
...
 
     context 'when a non-existing course topic is accessed' do
-      it 'redirects /404' do
-        create(:course, :maths)
-
-        get :index, params: { topic_id: 'history' }
-
-        expect(response).to redirect_to '/404'
+      it 'is not routable' do
+        expect(get: :index, params: { topic_id: 'history' }).not_to be_routable # rubocop:disable RSpec/ExpectActual
```
cwrw: false positive regarding rubocop. If we move those to `spec/routes` this no longer will be a problem

app/controllers/courses_controller.rb
cwrw: I've had a look at active model validation and it doesn't fit here as we are returning a collection, open to suggestions on improving this, otherwise will keep as is for now

app/controllers/courses_controller.rb
```diff
@@ -1,15 +1,26 @@
...
-
-    redirect_to '/404' unless @courses.any?
+  def index # rubocop:disable Metrics/AbcSize
+    @courses = Course.find_courses_near(
+      postcode: courses_params[:postcode],
+      topic: courses_params[:topic_id]
+    ).map { |c| CourseDecorator.new(c) }
+  rescue CourseGeospatialSearch::GeocoderAPIError
+    flash.now[:error] = t('courses.api_down_error')
+    @courses = Course.none
```
tavy88: I would assume @courses is nil at this stage. Why do you need this?

app/views/courses/index.html.erb
```diff
@@ -10,11 +10,40 @@
...
+        <div class="govuk-form-group <%='govuk-form-group--error' if flash.now[:error] %>">
+          <label class="govuk-label" for="postcode">
+            Postcode
+          </label>
+          <% if flash.now[:error] %>
+            <span id="national-insurance-number-error" class="govuk-error-message">
+              <span class="govuk-visually-hidden">Error:</span> <%= flash.now[:error] %>
+            </span>
+          <% end %>
+          <%= text_field(nil, :postcode, value: params[:postcode], class: "search-input govuk-input govuk-!-width-one-half #{'govuk-input--error' if flash.now[:error]}", required: true) %>
```
tavy88: I think you might want `text_field_tag` in this case maybe? Not sure why you passing the nil here.

app/views/courses/index.html.erb
```diff
@@ -10,11 +10,40 @@
...
+          <%= button_tag('', class: 'search-button-results', name: nil) %>
```
tavy88: Again why do you need `name: nil` here?

spec/factories/courses.rb
```diff
@@ -7,7 +7,7 @@
     address_line_2 { Faker::Address.street_address }
     town { Faker::Address.city }
     county { Faker::Address.state }
-    postcode { ['EC1A 1AA', 'EC1A 1JJ', 'EC1A 1JN', 'EC1A 2AH', 'EC1P 1AB', 'EC1P 1EW', 'EC1P 1NY'].sample }
+    postcode { ['NW11 8QE', 'NW11 7PT', 'NW11 7HB', 'NW10 7SF', 'NW10 7NZ', 'NW10 6HJ'].sample }
```
tavy88: You didn't like my postcodes I see :(

spec/controllers/courses_controller_spec.rb
smoothcontract: Add TODO here to move this test to spec/routes

app/controllers/courses_controller.rb
cwrw: `Course.none` returns an empty active record association. We should maintain the type of `@courses`which is an array across our views

spec/factories/courses.rb
cwrw: I needed a full list of valid postcodes, the previous list had some which were, and some which weren't. This means we will not get `0.0` coordinates in our table for testing

app/views/courses/index.html.erb
cwrw: if we don't add `nil` here text on the button is rendered

app/views/courses/index.html.erb
cwrw: passing `nil` as to ignore the object name field needed by `text_field`. Will change to `text_field_tag` as it's not needed there :+1:

PR #100, opened 2019-07-24
Add maths/english overview static pages

tavy88:
### Context

Course Type Overview for Maths & English

### Sreenshots

Screen Shot 2019-07-25 at 09 53 38
Screen Shot 2019-07-25 at 09 53 50

### JIRA

https://dfedigital.atlassian.net/browse/GET-120

### Test

/pages/english_overview
/pages/maths_overview

### DIFF
--- a/app/views/pages/maths_overview.html.erb
+++ b/app/views/pages/maths_overview.html.erb
@@ -1,0 +1,28 @@
+<% page_title :maths_overview %>
+<% content_for :breadcrumb do
+    generate_breadcrumbs(
+      t('breadcrumb.maths_overview'),
+      [
+        [t('breadcrumb.task_list_home'), task_list_path],
+        [t('breadcrumb.find_training_courses'), find_training_courses_path]
+      ]
+    )
+  end
+%>
+
+<div class="govuk-grid-row govuk-!-margin-top-7 govuk-body">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t('.title') %></h1>
+    <h2 class="govuk-heading-m">How will a maths course help my career ?</h2>
+    <p class="govuk-body">Employers often say they want candidates with maths skills – and having a qualification is one way to show you have those skills.</p>
+    <p class="govuk-body">Doing a maths course as an adult isn’t the same as learning it at school. You can find a course that focuses on using maths in your everyday job. </p>
+    <p class="govuk-body">Doing a maths course can improve your confidence and help show employers you’re able to use numbers at work. This can really boost your job prospects.</p>
+    <h2 class="govuk-heading-m">The right course for you</h2>
+    <p class="govuk-body">Lots of different maths courses are taught in your area, so you’ll be able to find a course at the right level for your needs. The course provider will help assess what level of course is right for you. You can also find courses that offer flexible learning and part time study to fit around your life.</p>
+    <p class="govuk-body">Many courses are available for free in your area.</p>
+    <%= link_to 'Find a maths course', courses_path('maths'), class: 'govuk-button govuk-!-margin-top-5'%>
+  </div>
+  <%= render "shared/contact_us" %>
+  <%= render 'shared/what_you_can_do_next' %>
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -27,6 +29,8 @@ en-GB:
     pages_find_training_courses: Get help to retrain - Find training
     courses_index: Get help to retrain - Courses near me
     pages_next_steps: Get help to retrain - Next steps
+    maths_overview: Get help to retrain - Benefits of maths
+    english_overview: Get help to retrain - Benefits of English
```
cwrw: could you prefix those with `pages_` as that's what we're doing with any files under the `pages/` directory

tavy88: What about `courses_index`?

config/routes.rb
```diff
@@ -10,6 +10,8 @@
   get 'task-list', to: 'pages#task_list'
   get 'find-training-courses', to: 'pages#find_training_courses'
   get 'next-steps', to: 'pages#next_steps'
+  get 'maths-overview', to: 'pages#maths_overview'
```
cwrw: maybe those fit better under courses 🤔 as so `/courses/maths_overview`, thoughts?

app/views/pages/maths_overview.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    <p class="govuk-body">Employers often say they want candidates with maths skills – and having a qualification is one way to show you have those skills.</p>
+    <p class="govuk-body">Doing a maths course as an adult isn’t the same as learning it at school. You can find a course that focuses on using maths in your everyday job. </p>
+    <p class="govuk-body">Doing a maths course can improve your confidence and help show employers you’re able to use numbers at work. This can really boost your job prospects.</p>
+    <h2 class="govuk-heading-m">The right course for you</h2>
+    <p class="govuk-body">Lots of different maths courses are taught in your area, so you’ll be able to find a course at the right level for your needs. The course provider will help assess what level of course is right for you. You can also find courses that offer flexible learning and part time study to fit around your life.</p>
+    <p class="govuk-body">Many courses are available for free in your area.</p>
+    <%= link_to 'Find a maths course', courses_path('maths'), class: 'govuk-button govuk-!-margin-top-5', target: '_blank' %>
```
cwrw: are those supposed to open in a new tab? as I think they probably shouldn't. Also the url should be `courses_path(topic_id: 'maths')` for it to work, same below

config/routes.rb
tavy88: I don't know...

tavy88: We could.

spec/features/course_overview_pages_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  scenario 'User continues journey to explore maths courses' do
+    create(:course, :maths)
+
+    visit(maths_overview_path)
+
+    click_on('Find a maths course')
+
+    expect(page).to have_text('Maths courses near me')
```
cwrw: this test should be failing as we are seeing `0 courses found` because of the wrong path, maybe you could count the courses shown instead? same below

config/routes.rb
tavy88: Do you mean the route or the location of the static pages?

spec/features/course_overview_pages_spec.rb
tavy88: This path will change. We need a post code param anyway but not for this story. Why this should fail? Please clarify? This page still has a title.

app/views/pages/maths_overview.html.erb
tavy88: They are if you check the story desc. This again will change. The path is correct.

config/routes.rb
cwrw: talked in person, have a think later

spec/features/course_overview_pages_spec.rb
cwrw: had a talk in person and this is fine

PR #101, opened 2019-07-25
Get 256 feature flags

smoothcontract:
### Context
Adds support for feature flags using Flipflop library (explored previously in a spike), but integrated with split.io service so that feature flag configuration can be externalised from the app in production. I did explore using Flipper too but Flipflop was much easier to integrate with split.io

Split.io Ruby SDK works by polling split.io API for config changes on a separate thread - this is every 5 seconds by default but we can configure that if we think it should be less frequent (but wait longer for feature changes to be reflected in the app).

* Add Flipflop and split.io gems - https://github.com/voormedia/flipflop https://help.split.io/hc/en-us/articles/360020673251
* Add new environment variable `SPLIT_API_KEY` (stored in vault)
* Add various "strategies" for Flipflop to control feature flags:
  - Query strings can be used to manually enable/disable features as required
  - Test strategy (in test mode) to support unit and feature tests
  - Custom strategy using split ruby SDK
* Add `/features` dashboard in development mode to review current feature configuration
* Add feature flag control to courses controller - this replaces previous temporary check of course models being present
* Add some helper methods for Rspec to allow programatic control of enabled features within tests

### Guidance to review
* New API key is only setup for dev1 at present. We can have multiple API keys to represent different environments. I'm envisaging one for dev/test and another for prod at present.
* Other environments will need to be configured to support API key before this is merged.
* Need to login to split.io to update feature flags for the deployed application. Ask me for details of that.


Gemfile
```diff
@@ -48,6 +45,13 @@ gem 'geocoder', '~> 1.5', '>= 1.5.1'
 # Postcode validation
 gem 'uk_postcode', '~> 2.1', '>= 2.1.4'
 
+# Web scraper
```
smoothcontract: I moved this up here so it's grouped with the other custom gems rather than being lost below

Gemfile
```diff
@@ -114,6 +115,3 @@ group :test do
   gem 'vcr', '~> 5.0', require: false
   gem 'webmock', '~> 3.6', require: false
 end
-
```
smoothcontract: Not part of this PR but it's been bugging me for ages 😛

app/controllers/courses_controller.rb
```diff
@@ -17,7 +17,7 @@ def index # rubocop:disable Metrics/AbcSize
   private
 
   def handle_missing_courses
-    redirect_to task_list_path if Course.count.zero?
+    protect_feature(:course_directory)
```
smoothcontract: We can use this approach within controllers to completely prevent access to stuff.

azure/config/dev1.parameters.json
```diff
@@ -32,6 +32,13 @@
             "secretName": "dev-secretKeyBase"
           }
         },
+        "splitApiKey": {
+          "reference": {
+            "keyVault": {
+              "id": "/subscriptions/f35bd249-3028-48bd-9aab-d00e9d1a56b7/resourceGroups/s108d01-shared/providers/Microsoft.KeyVault/vaults/s108d01-shared-kv-01"},
+            "secretName": "dev-splitApiKey"
+          }
+        },
```
smoothcontract: This might not be working perfectly yet - seems to be causing an issue on deployment.

config/features.rb
```diff
@@ -0,0 +1,10 @@
+require 'flipflop/split_client_strategy'
+
+Flipflop.configure do
+  strategy :test if Rails.env.test?
+  strategy :query_string
+  strategy :split_client, api_key: ENV['SPLIT_API_KEY'] if ENV['SPLIT_API_KEY'].present?
```
smoothcontract: We wouldn't normally enable the split client strategy for local development, but can if required so I've made this part optional based on the env var.

config/features.rb
```diff
@@ -0,0 +1,10 @@
+require 'flipflop/split_client_strategy'
+
+Flipflop.configure do
+  strategy :test if Rails.env.test?
+  strategy :query_string
```
smoothcontract: Query string strategy is really handy - can just append e.g. `?course_directory` to enable the feature. We should discuss whether it's okay to all this in production too.

config/features.rb
```diff
@@ -0,0 +1,10 @@
...
+  strategy :split_client, api_key: ENV['SPLIT_API_KEY'] if ENV['SPLIT_API_KEY'].present?
+
+  feature :foo, description: 'Example feature flag', default: true
+  feature :course_directory, description: 'Training courses feature'
```
smoothcontract: Any new features we add need to be defined here and added to split.io with the same name.

config/locales/en.yml
```diff
@@ -79,10 +79,19 @@ en-GB:
   help_improve_this_service:
     title: Help improve this service
     body: Take a quick survey and tell us about your experience
+  courses:
```
smoothcontract: I moved this so that the third party gem stuff is all grouped together at the bottom.

azure/config/dev1.parameters.json
smoothcontract: Update - so now it deploys okay, but still isn't exposing the `SPLIT_API_KEY` environment variable correctly :(

smoothcontract: Working now 👍

spec/lib/flipflop/strategies/split_client_strategy_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  # file rather than making API calls which is faster for testing
+  let(:api_key) { 'localhost' }
+  let(:path) { Rails.root.join('spec', 'fixtures', 'files', 'split.yml') }
+
+  describe '#initialize' do
+    context 'with invalid configuration' do
+      let(:api_key) { 'eXaMpLeApIkEy' }
+
+      it 'raises exception with invalid path configuration' do
+        expect { strategy }.to raise_error(RuntimeError)
```
tavy88: I usually like to see what the error message is, especially if there are different messages for different exceptions.

app/lib/flipflop/split_client_strategy.rb
```diff
@@ -0,0 +1,31 @@
...
+      def enabled?(feature)
+        @client.get_treatment('user', feature.to_s) == 'on'
+      end
+
+      private
+
+      def factory(options)
+        api_key = options.delete(:api_key)
+        path = options.delete(:path)
+        raise "#{self} path option is only permitted in localhost mode" if path.present? && api_key != 'localhost'
```
tavy88: I would extract these in 

```ruby
def api_key
  @api_key ||= options.delete(:api_key)
end

def path
...
end

```

And make options `attr_accessor` so you don't have to pass it as arg every time.

smoothcontract: Seems like overkill to me. It’s already a very short method and I have no other need to make options an instance variable.

azure/config/dev1.parameters.json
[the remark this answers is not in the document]
smoothcontract: It’s been prototyped in dev1. I’ll work with Miguel to get this added to other envs but that can be in a follow up PR as this blocks other stories in the sprint.

spec/lib/flipflop/strategies/split_client_strategy_spec.rb
smoothcontract: It’s not that important here. The error exists just to flag if someone passes path option mistakenly when not expecting to test split.io in localhost mode. There’s only one error that gets potentially raised too.

spec/support/feature_flags.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def switch_feature!(feature_name, enabled)
+    test_strategy = Flipflop::FeatureSet.current.test!
+    test_strategy.switch!(feature_name, enabled)
+  end
+end
+
+RSpec.configure do |config|
+  config.include FeatureFlags
+end
```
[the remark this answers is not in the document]
smoothcontract: It’s a self contained approach to keep all of the feature flag spec stuff in one file. Same approach has been used elsewhere in the project.

app/lib/flipflop/split_client_strategy.rb
tavy88: Not sure why you feel it's overkill as this approach would lead more towards the single responsibility principle but up to you.

spec/support/feature_flags.rb
```diff
@@ -0,0 +1,18 @@
...
+    switch_feature!(feature_name, false)
+  end
+
+  def switch_feature!(feature_name, enabled)
+    test_strategy = Flipflop::FeatureSet.current.test!
+    test_strategy.switch!(feature_name, enabled)
+  end
+end
+
+RSpec.configure do |config|
```
cwrw: I'm not sure you need this block if you required this in the rails helper instead

azure/config/dev1.parameters.json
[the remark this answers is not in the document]
mafonso: We could add the config for the other environments, because on merge it will bubble up to "test" and potentially production.
I don't believe it will fail, because it would just default to "disabled", but would not be consistent.  
I can help you with this.

spec/support/feature_flags.rb
smoothcontract: See my reply below to @tavy88 - the module needs to be included in the Rspec config somewhere.

PR #103, opened 2019-07-26
Get 233 enable skills admin

smoothcontract:
### Context
Adds `enabled` attribute for skills, improves admin pages functionality and courses data setup.

* New `enabled` attribute for skills will be used during the forthcoming spike to explore use of skills for matching job profiles. We know that not all of the skills will be used for this, so this adds the ability to disable some of those. This capability won't be used programatically yet, but adding the column allows the data to be setup in advance and exposed via the admin interface.
* Improved data export (CSV) for job profile and skills in admin pages. Now includes a list of job profiles for each skill, and skills for each job profile. Some additional tweaks to achieve that without N+1 query issues.
* Added a rake task and some docs around setting up sample course data
* Added courses to the admin pages

### Guidance to review
Mostly admin stuff, so best to pull locally and explore; the admin pages only work with `RAILS_ENV=development`.

### DIFF
--- a/app/admin/job_profiles.rb
+++ b/app/admin/job_profiles.rb
@@ -67,6 +67,25 @@ if defined?(ActiveAdmin)
       end
       f.actions
     end
+
+    controller do
+      def scoped_collection
+        super.includes :skills
+      end
+    end
+
+    csv do
+      column :id
+      column :slug
+      column :name
+      column :recommended
+      column :salary_min
+      column :salary_max
+      column :alternative_titles
+      column :skills do |job_profile|
+        job_profile.skills.map(&:name)
+      end
+    end
   end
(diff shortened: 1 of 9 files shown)


app/admin/job_profiles.rb
```diff
@@ -67,6 +67,25 @@
       end
       f.actions
     end
+
+    controller do
+      def scoped_collection
+        super.includes :skills
```
tavy88: Nice so any relation needed further down in the DSL would need to be declared under the `scoped_collection`?

smoothcontract: Yep, that's the convention within ActiveAdmin. Normally you'd use that to restrict data by account or something, but it's also useful here to prevent N+1 query on each job profile.

PR #105, opened 2019-07-26
Add redesigned Find Training Courses page

tavy88:
Considerations:

- this feature is flagged and will go live
when courses feature goes live
- this page has a new URL so the feature is
not backwards compatible
- keeping it as a separate page seems cleaner

### Screenshots

Screen Shot 2019-07-29 at 10 42 00

Screen Shot 2019-07-29 at 11 26 10

### JIRA

https://dfedigital.atlassian.net/browse/GET-118


config/routes.rb
```diff
@@ -9,10 +9,14 @@
   get '/500', to: 'errors#internal_server_error', via: :all
 
   get 'task-list', to: 'pages#task_list'
-  get 'find-training-courses', to: 'pages#find_training_courses'
   get 'next-steps', to: 'pages#next_steps'
-  get 'maths-course-overview', to: 'pages#maths_overview'
-  get 'english-course-overview', to: 'pages#english_overview'
+  get 'find-training-courses', to: 'pages#find_training_courses'
+
+  constraints(->(_req) { Flipflop.course_directory? }) do
```
smoothcontract: ❤️

spec/features/training_hub_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'rails_helper'
+
+RSpec.feature 'Training Hub page', type: :feature do
```
smoothcontract: @cwrw Thoughts on explicitly stating the type of test with specs? It seems superfluous to me given that Rspec can deduce this from the file location.

spec/routing/courses_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'rails_helper'
+
+RSpec.describe 'routes for Courses', type: :routing do
```
smoothcontract: 👍

spec/features/training_hub_spec.rb
cwrw: so I think it should be ok to remove, however we might need to add some specific logic for features later where this might return, but for now should be ok

spec/features/training_hub_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  background do
+    enable_feature! :course_directory
+  end
+
+  scenario 'User navigates to the training hub page' do
+    visit(training_hub_path)
```
cwrw: can we link to this from the task list page?

spec/features/task_list_spec.rb
```diff
@@ -61,4 +63,12 @@
       ]
     )
   end
+
+  scenario 'User navigates to task list page when course directory feature is enabled' do
```
cwrw: I think we've got this covered in the further tests (if we navigate to the training hub in the `training_hub_spec` instead

spec/features/training_hub_spec.rb
tavy88: Will be done on another PR as discussed.

PR #106, opened 2019-07-29
GET-36 Check location eligibility 

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-36

* Add location eligibility behind feature flag
* Refactor Geolocation search for cleanup and validations
* Add Form helper, taken from: https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/app/helpers/govuk_form_helper.rb

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -3,7 +3,7 @@ class CoursesController < ApplicationController
 
   def index
     @search = course_geospatial_search
-    @courses = course_geospatial_search.courses.map { |c| CourseDecorator.new(c) }
+    @courses = @search.valid? ? course_geospatial_search.find_courses.map { |c| CourseDecorator.new(c) } : Course.none
   end
 
   private
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -2,8 +2,11 @@ class CoursesController < ApplicationController
   before_action :handle_missing_courses
 
   def index
-    @search = course_geospatial_search
-    @courses = @search.valid? ? course_geospatial_search.find_courses.map { |c| CourseDecorator.new(c) } : Course.none
+    @search = CourseGeospatialSearch.new(
+      postcode: courses_params[:postcode],
+      topic: courses_params[:topic_id]
+    )
+    @courses = @search.find_courses.map { |c| CourseDecorator.new(c) }
   end
 
   private
@@ -14,12 +17,5 @@ class CoursesController < ApplicationController
 
   def courses_params
     params.permit(:postcode, :topic_id)
-  end
-
-  def course_geospatial_search
-    @course_geospatial_search ||= CourseGeospatialSearch.new(
-      postcode: courses_params[:postcode],
-      topic: courses_params[:topic_id]
-    )
   end
 end
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -1,17 +1,9 @@
 class CoursesController < ApplicationController
   before_action :handle_missing_courses
 
-  def index # rubocop:disable Metrics/AbcSize
-    @courses = Course.find_courses_near(
-      postcode: courses_params[:postcode],
-      topic: courses_params[:topic_id]
-    ).map { |c| CourseDecorator.new(c) }
-  rescue CourseGeospatialSearch::GeocoderAPIError
-    flash.now[:error] = t('courses.api_down_error')
-    @courses = Course.none
-  rescue CourseGeospatialSearch::InvalidPostcodeError
-    flash.now[:error] = t('courses.invalid_postcode_error')
-    @courses = Course.none
+  def index
+    @search = course_geospatial_search
+    @courses = course_geospatial_search.courses.map { |c| CourseDecorator.new(c) }
   end
 
   private
@@ -22,5 +14,12 @@ class CoursesController < ApplicationController
 
   def courses_params
     params.permit(:postcode, :topic_id)
+  end
+
+  def course_geospatial_search
+    @course_geospatial_search ||= CourseGeospatialSearch.new(
+      postcode: courses_params[:postcode],
+      topic: courses_params[:topic_id]
+    )
   end
 end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,11 +4,12 @@ class PagesController < ApplicationController
   end
 
   def location_eligibility
-    protect_feature(:location_eligibility)
-    @search = course_geospatial_search
+    @search = CourseGeospatialSearch.new(
+      postcode: eligibility_params[:postcode]
+    )
 
-    if eligibility_params[:postcode] && @search.valid?
-      course_geospatial_search.find_courses.any? ? (redirect_to task_list_path) : (redirect_to root_path)
+    if eligibility_params[:postcode].present? && @search.valid?
+      @search.find_courses.any? ? redirect_to(task_list_path) : redirect_to(root_path)
     end
   end
 
@@ -16,11 +17,5 @@ class PagesController < ApplicationController
 
   def eligibility_params
     params.permit(:postcode)
-  end
-
-  def course_geospatial_search
-    @course_geospatial_search ||= CourseGeospatialSearch.new(
-      postcode: eligibility_params[:postcode]
-    )
   end
 end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -3,23 +3,24 @@ class PagesController < ApplicationController
     render template: "pages/#{params[:page]}"
   end
 
-  def location_eligibility # rubocop:disable Metrics/AbcSize
+  def location_eligibility
     protect_feature(:location_eligibility)
-    return unless eligibility_params[:postcode]
-
-    courses = Course.find_courses_near(
-      postcode: eligibility_params[:postcode]
-    )
-    courses.any? ? (redirect_to task_list_path) : (redirect_to root_path)
-  rescue CourseGeospatialSearch::GeocoderAPIError
-    flash.now[:error] = t('courses.api_down_error')
-  rescue CourseGeospatialSearch::InvalidPostcodeError
-    flash.now[:error] = t('courses.invalid_postcode_error')
+    if course_geospatial_search.errors.any?
+      @search = course_geospatial_search
+    else
+      course_geospatial_search.courses.any? ? (redirect_to task_list_path) : (redirect_to root_path)
+    end
   end
 
   private
 
   def eligibility_params
     params.permit(:postcode)
+  end
+
+  def course_geospatial_search
+    @course_geospatial_search ||= CourseGeospatialSearch.new(
+      postcode: eligibility_params[:postcode]
+    )
   end
 end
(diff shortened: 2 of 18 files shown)


app/models/course_geospatial_search.rb
```diff
@@ -1,38 +1,46 @@
...
   def coordinates
-    Geocoder.coordinates(uk_postcode)
+    return unless valid_uk_postcode?
+
+    @coordinates ||= Geocoder.coordinates(uk_postcode.to_s)
   rescue SocketError, Timeout::Error, Geocoder::Error => e
     Rails.logger.error("Geocoder API error: #{e.message}")
-    raise GeocoderAPIError
+    errors.add(:postcode, I18n.t('courses.api_down_error'))
+    return # rubocop:disable Style/RedundantReturn
```
cwrw: this is needed for memoization as there is an exception being rescued

spec/features/location_eligibility_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+    visit(location_eligibility_path)
+    fill_in('postcode', with: 'NW6 8ET')
+    find('.govuk-button').click
+
+    expect(page).to have_text(/Enter a real postcode/)
+  end
+
+  xscenario 'User gets relevant messaging if their address coordinates could not be retrieved' do
```
cwrw: keeping this test until we add a page for api down error next ticket

app/models/course_geospatial_search.rb
```diff
@@ -1,38 +1,50 @@
...
+    @coordinates ||= Geocoder.coordinates(uk_postcode.to_s)
   rescue SocketError, Timeout::Error, Geocoder::Error => e
     Rails.logger.error("Geocoder API error: #{e.message}")
     raise GeocoderAPIError
   end
 
-  def courses
-    topic.present? ? Course.where(topic: topic) : Course.all
+  def postcode_exists
+    return unless errors.blank?
```
cwrw: this is so we don't try to run this validation when there is no postcode or it's invalid

app/controllers/courses_controller.rb
```diff
@@ -1,17 +1,9 @@
...
-    ).map { |c| CourseDecorator.new(c) }
-  rescue CourseGeospatialSearch::GeocoderAPIError
-    flash.now[:error] = t('courses.api_down_error')
-    @courses = Course.none
-  rescue CourseGeospatialSearch::InvalidPostcodeError
-    flash.now[:error] = t('courses.invalid_postcode_error')
-    @courses = Course.none
+  def index
+    @search = course_geospatial_search
+    @courses = @search.valid? ? course_geospatial_search.find_courses.map { |c| CourseDecorator.new(c) } : Course.none
```
tavy88: Could we put 

```ruby
course_geospatial_search.find_courses.map { |c| CourseDecorator.new(c) }
```

on a separate private method maybe? It will prevent us from having a horribly long line of code. What do you think?

spec/features/find_training_courses_spec.rb
```diff
@@ -35,10 +35,23 @@
...
+    create(:course, topic: 'maths')
+
+    visit(courses_path(topic_id: 'maths'))
+    fill_in('postcode', with: 'NW6 8ET')
+    find('.search-button-results').click
+
+    expect(page).to have_text(/Enter a real postcode/)
+  end
+
+  xscenario 'User gets relevant messaging if their address coordinates could not be retrieved' do
```
cwrw: kept this test for next ticket which will add a page

app/controllers/pages_controller.rb
```diff
@@ -2,4 +2,25 @@ class PagesController < ApplicationController
   def show
     render template: "pages/#{params[:page]}"
   end
+
+  def location_eligibility
+    protect_feature(:location_eligibility)
```
tavy88: Could this be a `before_action`?

E.g:

```ruby
before_action -> { protect_feature(:location_eligibility) }, only: [ :location_eligibility ]
```

app/controllers/pages_controller.rb
```diff
@@ -2,4 +2,25 @@ class PagesController < ApplicationController
...
+    @search = course_geospatial_search
```
tavy88: I wonder if we can ever have a search that is not a `course_geospatial_search` within this context? Otherwise, we can just use `course_geospatial_search` everywhere, instead? It would spare us from having to assign it to a second instance var.

app/controllers/pages_controller.rb
```diff
@@ -2,4 +2,25 @@ class PagesController < ApplicationController
...
+
+    if eligibility_params[:postcode] && @search.valid?
```
tavy88: Why not use `course_geospatial_search.valid?` here? Why do you need a second instance var `@search`? Also, I think you might want to use 

```ruby
eligibility_params[:postcode].present? 
```
here? You could have `?postcode` in your URL but you are expecting a value I assume.

app/models/course_geospatial_search.rb
```diff
@@ -1,38 +1,50 @@
 class CourseGeospatialSearch
-  InvalidPostcodeError = Class.new(StandardError)
+  include ActiveModel::Validations
```
smoothcontract: I think perhaps a key reason why you were struggling with the validations is that you've just included the validation stuff rather than inheriting from the ActiveModel class. There's a bit more to the error handling with callbacks and things - perhaps that might have made adding validations a little simpler. It's fine to leave it as is now if everything is working as you wish.

app/controllers/pages_controller.rb
smoothcontract: I think I prefer the approach Octavian has used in this PR - maybe the `protect_feature` method was a bad (my) idea. In most cases it's probably better to return a 404 for an unfinished feature rather than arbitrary redirecting. That also means that protecting access is all handled in the routes file rather than spread around controllers i.e. easier to remove later. See: https://github.com/DFE-Digital/get-help-to-retrain/pull/105/files#diff-21497849d8f00507c9c8dcaf6288b136R16

app/controllers/pages_controller.rb
smoothcontract: Does this test still work if `postcode` param is not passed? I'm wondering if it should check `.valid?` or something slightly more thorough?

app/helpers/form_helper.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  private
+
+  def error_id(object_name, attribute)
+    "#{object_name}_#{attribute}-error"
+  end
+
+  def messages(object, attribute)
+    object.errors.messages[attribute].map { |message|
+      content_tag(:span, 'Error:', class: 'govuk-visually-hidden') + ' ' + message
```
tavy88: I think you can use `concat` here to just mutate.

app/controllers/pages_controller.rb
```diff
@@ -2,4 +2,25 @@ class PagesController < ApplicationController
...
+      course_geospatial_search.find_courses.any? ? (redirect_to task_list_path) : (redirect_to root_path)
```
smoothcontract: I find this bracket arrangement slightly strange - is rubocop happy with this? Would be more conventional to write as:

```ruby
course_geospatial_search.find_courses.any? ? redirect_to(task_list_path) : redirect_to(root_path)
```

smoothcontract: Maybe it's just a preference thing...

app/helpers/form_helper.rb
```diff
@@ -0,0 +1,39 @@
+module FormHelper
```
smoothcontract: Nice, thanks for adding this - will be good to reuse for other forms 👍

app/views/courses/index.html.erb
```diff
@@ -16,18 +16,14 @@
...
             Postcode
           </label>
-          <% if flash.now[:error] %>
-            <span id="national-insurance-number-error" class="govuk-error-message">
-              <span class="govuk-visually-hidden">Error:</span> <%= flash.now[:error] %>
-            </span>
-          <% end %>
-          <%= text_field_tag(:postcode, params[:postcode], class: "search-input govuk-input govuk-!-width-one-half #{'govuk-input--error' if flash.now[:error]}", required: true) %>
+          <%= errors_tag @search, :postcode %>
+          <%= text_field_tag(:postcode, params[:postcode], class: css_classes_for_input(@search, :postcode, "search-input govuk-input govuk-!-width-one-half"), required: true) %>
```
smoothcontract: This is fine for now, perhaps in time the new helpers could be refactored into a custom form builder instead.

app/controllers/pages_controller.rb
cwrw: I think its ok to keep it in here as we're only using it for one method, when we have more methods we can move it out

app/views/home/index.html.erb
```diff
@@ -4,7 +4,7 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl"><%= t('.title') %></h1>
     <p class="govuk-body">Use this service to find out what other jobs you can do with your current skills. You can also find out how to get new skills that could help you find a better job.</p>
-    <%= link_to 'Start now', task_list_path, class: 'govuk-button govuk-button--start govuk-!-margin-top-2 govuk-!-margin-bottom-8', role: 'button' %>
+    <%= link_to 'Start now', location_eligibility_path, class: 'govuk-button govuk-button--start govuk-!-margin-top-2 govuk-!-margin-bottom-8', role: 'button' %>
```
smoothcontract: I think this should be feature flagged, otherwise users will go via the new page through a redirect to the task list page (I mentioned previously that the new page should be feature flagged in the routes file too). I guess either approach works, but the extra redirect via an incomplete feature (and the link url not representing the final destination) is not ideal. e.g. https://github.com/DFE-Digital/get-help-to-retrain/pull/105/files#diff-aac4c77d3ee690c9b3d57146c170ae22R56

app/helpers/form_helper.rb
cwrw: I think this is ok as both are interchangeable, we're not doing any large processes so performance should be the same

app/models/course_geospatial_search.rb
cwrw: I'm not sure `ActiveModel` would have helped in this case, as there would have been the same issues as before regarding validations, especially when they are called etc. But happy to hear some about improvements it could add and will amend otherwise

app/views/courses/index.html.erb
cwrw: There is a gov uk form builder gem https://github.com/ministryofjustice/govuk_elements_form_builder and similar like it, felt overkill for now but we could add it later

app/controllers/pages_controller.rb
cwrw: Let's discuss tomorrow, as not really sold on this one.

app/models/course_geospatial_search.rb
smoothcontract: That's fine, nothing needs to change but I was thinking the validation behaviour might have been more consistent with other models if inheriting from the entire class rather than mixing in just one module. It's working now so can stay as is.

app/views/courses/index.html.erb
smoothcontract: cool, something to consider later anyway

PR #107, opened 2019-07-29
Fix intermittent spec failures

tavy88:
Introducing feature flagging capability seems to
have made some of our specs intermittently fail.

We need to enforce the feature is ON for those specs.


smoothcontract:
Good catch, thanks 👍

PR #108, opened 2019-07-30
Add on-page user feedback survey

tavy88:
### Context

We are trying to bring in the in-page user feedback survey using Smart Survey.

Behaviour:

- once the user clicks on YES we send a custom event to AppInsights to record the answer
- once the user clicks no, a SMART SURVEY form is displayed under and the data they submit goes straight to smart survey.
- the styling of the form is controlled from Smart Survey dashboard rather from our codebase.

### JIRA

https://dfedigital.atlassian.net/browse/GET-133

### Screenshots

Screen Shot 2019-07-30 at 12 23 57
Screen Shot 2019-07-30 at 12 24 06
Screen Shot 2019-07-30 at 12 24 15
Screen Shot 2019-07-30 at 14 16 10

### TESTING ###

Add inside .env.development:
USER_FEEDBACK_SMART_SURVEY_LINK=https://www.smartsurvey.co.uk/s/1BBQK/


app/views/shared/feedback/_survey.html.erb
```diff
@@ -0,0 +1,27 @@
+<div class="feedback-prompt-questions hidden">
+  <iframe id="ss-embed-frame" onload="window.parent.parent.scrollTo(0,0)" src="http://www.smartsurvey.co.uk/s/8IBQ1/" style="width:100%;height:296px;border:0px;padding-bottom:4px;" frameborder="0"><a href="http://www.smartsurvey.co.uk/s/8IBQ1/">Please take our survey</a></iframe>
```
smoothcontract: Could the smartsurvey URL be externalised please? It doesn't really need to be configurable, but it is repeated twice in this view. Maybe split out as a constant or instance variable somewhere to DRY this up?

app/views/shared/feedback/_survey.html.erb
```diff
@@ -0,0 +1,27 @@
...
+</div>
+
+<div class="feedback-prompt feedback-footer" >
+  <div class="feedback-prompt-full-question">
+    <h2 class="feedback-prompt-question">Is this page useful?</h2>
+    <%= link_to '#', {
+      class: 'feedback-prompt-link govuk-link answer-yes',
+      data: {
+        tracked_event: 'Is this page useful in-page survey - Clicked Yes'
```
smoothcontract: I think perhaps the list of custom events should be reviewed to establish some naming conventions. This doesn't really seem to fit with the others already defined e.g. `"Courses - Clicked course link: #{course.url}"`. Perhaps it's just because it's a long name. Also is it possible to pass extra data to the custom event so that "Yes" could be split out as metadata rather than included in the event name. If so I think this would make ongoing use of the custom event data more useful.

smoothcontract: Also.. do we want a custom event for the opposite case where user clicks "No" ?

smoothcontract: Looks like you can pass custom `properties` to the Azure API (at least via the Ruby SDK, presumably the javascript implementation is the same).

tavy88: @smoothcontract Good point. My initial thought was not to track the NO with custom events as NOs are tracked by Smart Survey when they click Submit. In terms of naming convention for that custom event, I am open to suggestions.

app/views/shared/feedback/_survey.html.erb
tavy88: Actually the href on the bottom is not needed. I will remove that. I think this will need to stay as an env var actually as the real survey will have a different ID. Not sure if we plan to switch them for now though.

smoothcontract: Okay, if it's just one place where we link to Smartsurvey then I'm more relaxed about it. It doesn't necessarily need to be configurable at this point either. However as this isn't behind a feature flag then the correct survey will need to be published and referenced here before merging this branch so that we are able to push to production. Also worth double checking with PO that this feature can go live immediately if required. If not, then it can be feature flagged so there's less urgency to get the correct survey link in place.

app/views/shared/feedback/_survey.html.erb
smoothcontract: Maybe just something a bit shorter e.g. "In-page survey" and then pass properties like `response: "Yes"`.  I definitely would like to see the properties split out as it will make it cleaner to filter the results in Azure. Thoughts?

app/views/shared/feedback/_survey.html.erb
```diff
@@ -0,0 +1,27 @@
+<div class="feedback-prompt-questions hidden">
+  <iframe id="ss-embed-frame" onload="window.parent.parent.scrollTo(0,0)" src="http://www.smartsurvey.co.uk/s/8IBQ1/" style="width:100%;height:296px;border:0px;padding-bottom:4px;" frameborder="0"></iframe>
```
cwrw: talked in person but for posterity: could we move the url for smart survey to some form of ENV variable so that we have a local test one and a production survey to avoid mixup?

tavy88: Yep.

app/views/shared/feedback/_survey.html.erb
tavy88: No worries I will not render the survey form at all if it's not set as an ENV var.

app/views/layouts/application.html.erb
```diff
@@ -62,9 +61,13 @@
         </div>
         <%= yield %>
       </main>
+      <% if ENV.fetch('USER_FEEDBACK_SMART_SURVEY_LINK', nil).present? %>
```
smoothcontract: You can simplify this slightly to `if ENV['USER_FEEDBACK_SMART_SURVEY_LINK'].present?` - that also returns `nil` if the value is not populated and is a bit easier to read.

azure/template.json
```diff
@@ -146,6 +146,13 @@
     "appServicePlanInstances": {
       "type": "int",
       "defaultValue": 1
+    },
+    "userFeedbackSmartSurveyLink": {
+      "type": "string",
+      "defaultValue": "", 
+      "metadata": {
+        "description:": "User feedback in page Smart Survey"
```
smoothcontract: Maybe just add the world `url` to this description please?

app/webpacker/packs/track-events.js
```diff
@@ -9,9 +9,16 @@ function TrackEvents () {
...
   function sendEvent() {
     var eventLabel = this.dataset.trackedEvent;
-
+    var eventProps = this.dataset.trackedEventProps;
+    
+    if (typeof(eventProps) !== 'undefined') {
+      eventProps = JSON.parse(eventProps);
+    } else {
+      eventProps = {};
+    }
```
smoothcontract: Nice 👍

app/views/shared/feedback/_survey.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+<div id="feedback-prompt-container" class="feedback-prompt-container" >
+  <div id="feedback-prompt-question">
+    <h2 class="feedback-prompt-content">Is this page useful?</h2>
+    <%= link_to '#', {
+      class: 'feedback-prompt-link govuk-link',
+      id: 'answer-yes',
+      data: {
+        tracked_event: 'User Feedback On Page Survey',
+        tracked_event_props: { 'answer': 'yes' }
```
smoothcontract: 👍

azure/template.json
```diff
@@ -146,6 +146,13 @@
     "appServicePlanInstances": {
       "type": "int",
       "defaultValue": 1
+    },
+    "userFeedbackSmartSurveyLink": {
+      "type": "string",
+      "defaultValue": "",
```
cwrw: if we add the link here as a default value, do you need to add it on every environment?

cwrw: maybe @mafonso knows the answer to this?

tavy88: No this is to make it optional which we want. I consulted with him before @cwrw

.env.test
```diff
@@ -0,0 +1 @@
+USER_FEEDBACK_SMART_SURVEY_LINK=https://www.smartsurvey.co.uk/s/8IBQ1/
```
cwrw: can you add `.env.*.local` to `.gitignore`?

spec/features/user_feedback_survey_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+RSpec.feature 'User Feedback In Page Survey' do
+  scenario 'Feedback form is not present unless enabled from ENV' do
+    allow(ENV).to receive(:[]).with('USER_FEEDBACK_SMART_SURVEY_LINK').and_return(nil)
+
+    visit(root_path)
+
+    expect(page).not_to have_text('Is this page useful?')
+  end
+
+  scenario 'User gets a thank you message when clicking yes' do
```
cwrw: can you enable `js` on both the tests here and below as these will always pass regardless of `js` being enabled. 
example: `scenario 'User gets a thank you message when clicking yes', :js do`

PR #109, opened 2019-07-30
Get 233 admin updates

smoothcontract:
### Context
Some fixes and improvements to admin pages

* Fixed default sort order for all admin pages
* Fixed ability to edit courses (was throwing an error on save)
* Added ability to edit skills (to change name and enabled)

### Guidance to review
Admin pages are only available in development mode, so will need to pull branch and test locally if required.

### DIFF
--- a/app/admin/courses.rb
+++ b/app/admin/courses.rb
@@ -1,4 +1,5 @@
 if defined?(ActiveAdmin)
+  # rubocop:disable Metrics/BlockLength
   ActiveAdmin.register Course do
     permit_params :title, :provider, :url, :address_line_1, :address_line_2, :town, :county, :postcode,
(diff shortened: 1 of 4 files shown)


app/admin/courses.rb
```diff
@@ -1 +1,8 @@
-ActiveAdmin.register Course if defined?(ActiveAdmin)
+if defined?(ActiveAdmin)
+  ActiveAdmin.register Course do
+    permit_params :title, :provider, :url, :address_line1, :address_line2, :town, :county, :postcode,
+                  :email, :topic, :phone_number, :active, :latitude, :longitude
```
cwrw: discussed in person, we can remove lat/long from being edited as they are generated automatically from the postcode

PR #111, opened 2019-07-31
GET-39 Location Ineligible page

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-39

Add location ineligible page, when there are no courses close to user, which allows user to go to task list page anyway. based on https://get-help-to-retrain-prototype.herokuapp.com/main/release2_e2e_v2/location_ineligible
Also add error page incase API is down on the location eligibility path, we are still holding off on that one on the courses path as designs/journey not ready

ready to test on: https://dev2.nrs-ghtr.org.uk/

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -23,6 +23,47 @@
     {
       "warning_type": "Cross-Site Scripting",
       "warning_code": 4,
+      "fingerprint": "365fd9ea8cee0c69534a037daf9ffb9a53443fce909bcbf94046dc52347fc842",
+      "check_name": "LinkToHref",
+      "message": "Potentially unsafe model attribute in `link_to` href",
+      "file": "app/views/courses/_course.html.erb",
+      "line": 7,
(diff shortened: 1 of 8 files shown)


config/brakeman.ignore
```diff
@@ -51,6 +92,6 @@
       "note": "This is a known false positive from brakeman. We trust our data."
     }
   ],
-  "updated": "2019-07-29 17:42:49 +0100",
+  "updated": "2019-07-31 14:59:22 +0100",
```
cwrw: keeping both ignores so we don't have to keep doing this

config/locales/en.yml
```diff
@@ -96,7 +101,7 @@ en-GB:
   courses:
     invalid_postcode_error: Enter a valid postcode.
     nonexisting_postcode_error: Enter a real postcode.
-    api_down_error: We cannot verify your postcode at the current time
+    postcode_search_error: Sorry, there is a problem with this service. Postcode search isn't working right now.
```
cwrw: real content to be added tomorrow

PR #115, opened 2019-08-02
Add custom track events for postcode and job searches

smoothcontract:
### Context
This uses asynchronous telemetry client channel so doesn't block main thread - takes a few seconds for events to show up in Azure Application Insights but works well. Refactored relevant controllers very slightly to make extracting the key search param a little neater / DRY.

### Guidance to review
Easiest to test once deployed to a dev environment.
Search for job titles in check your skills and explore careers pages.
Search for postcode in courses page.
Ensure that custom events are recorded in Azure Application Insights

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,7 +2,9 @@ class ApplicationController < ActionController::Base
   private
 
   def track_event(event_key, properties = {})
-    TELEMETRY_CLIENT.track_event I18n.t(event_key), properties: properties if defined?(TELEMETRY_CLIENT)
+    return unless defined?(TELEMETRY_CLIENT)
+
+    TELEMETRY_CLIENT.track_event(I18n.t(event_key, scope: :events), properties: properties)
   end
 
   def protect_feature(feature)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,10 @@
 class ApplicationController < ActionController::Base
   private
 
+  def track_event(event_key, properties = {})
+    TELEMETRY_CLIENT.track_event I18n.t(event_key), properties: properties if defined?(TELEMETRY_CLIENT)
+  end
+
   def protect_feature(feature)
     redirect_to task_list_path unless Flipflop.enabled?(feature)
   end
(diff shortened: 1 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,6 +1,10 @@
 class ApplicationController < ActionController::Base
   private
 
+  def track_event(event_key, properties = {})
+    TELEMETRY_CLIENT.track_event I18n.t(event_key), properties: properties if defined?(TELEMETRY_CLIENT)
```
cwrw: can we add brackets here and everywhere? `TELEMETRY_CLIENT.track_event(I18n.t(event_key), properties: properties)` as well as `track_event(:check_your_skills_index_search, search: search)` etc as it makes it easier to read?

app/controllers/check_your_skills_controller.rb
```diff
@@ -1,10 +1,15 @@
 class CheckYourSkillsController < ApplicationController
   def results
-    @job_profiles = JobProfile.search(job_profile_params[:search]).page(params[:page])
+    track_event :check_your_skills_index_search, search: search
```
cwrw: do you not need `t()` here? or are those read automatically?

app/controllers/application_controller.rb
smoothcontract: I'd agree with brackets improving readability here so have added. Less swayed by the calls to `track_event` though - seems readable already to me.

app/controllers/pages_controller.rb
```diff
@@ -4,17 +4,21 @@ def show
   end
 
   def location_eligibility
-    @search = CourseGeospatialSearch.new(
-      postcode: eligibility_params[:postcode]
-    )
+    track_event :pages_location_eligibility_search, search: postcode
```
cwrw: you could add this line into the method location_eligibility_through_courses to make sure we always a valid postcode?

smoothcontract: I think I might leave it where it is for now - it would be good to capture some stats on how many invalid postcodes we get, maybe collate any common user errors etc. If that data proves not to be helpful then this can be tweaked to capture fewer postcodes. It's easier to restrict the data moving forwards once we know we're not missing anything useful.

cwrw: ok that's fair then maybe we could wrap this with a `postcode.present?` check to make sure not to fire an event on just landing on the page

smoothcontract:
> Is there a risk of the tracking event slowing the page down or breaking if it can't talk to the Client?

Events are queued and sent asynchronously in batches so we shouldn't see any impact on the running service. Probably easiest to monitor this on test environment for any adverse affects.

PR #117, opened 2019-08-05
GET-281 Imrpove CI build time

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-281

* Download RVM to use our own ruby version 2.6.3
* Add path to other tasks with: `[task.prependpath]` as profile/bashrc loaded in other tasks reads from readonly system profile/bashrc
* Move tests out of docker build to run on box
* Add caching for bundle step. Unfortunately the caching step does not return the bin files as executables and loses the symlinks. So we have a step to allow binary file mods to be changed. Also note that bundle always runs as (there is no condition to say it doesn't run with caching is restored), this is to restore the correct symlinks. 
Bundle running again does not cause any slowness if caching is restored so it is ok until caching is fixed.

Build time now is ~9 minutes as opposed to ~15 minutes before

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN apk add --update $BUILD_PACKAGES && \
 # Generate Assets
 WORKDIR /app
 COPY Gemfile Gemfile.lock .ruby-version ./
-RUN bundle install --clean --force --without "development" \
+RUN bundle install --clean --force --without "development test" \
      && rm -rf /usr/local/bundle/cache/*.gem \
      && find /usr/local/bundle/gems/ -name "*.c" -delete \
      && find /usr/local/bundle/gems/ -name "*.o" -delete
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,6 +9,7 @@ pr: none
 variables:
   IMAGE_NAME: '$(dockerHubUserName)/get-help-to-retrain'
   postgresUrl: postgres://test:test@localhost:5432/get-help-to-retrain
+  BUNDLE_CACHE_FOLDER: $(Pipeline.Workspace)/vendor/bundle
 
 resources:
   containers:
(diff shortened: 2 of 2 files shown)


azure-pipelines.yml
```diff
@@ -29,39 +30,87 @@ resources:
...
 
-      steps:
-        - task: UseRubyVersion@0
-          inputs:
-            versionSpec: '>= 2.6.2'
-            addToPath: true # Optional
+    steps:
+      - script: |
+          gateway_ip=$(docker network inspect bridge --format='{{range .IPAM.Config}}{{.Gateway}}{{end}}')
+          echo "##vso[task.setvariable variable=GatewayIP;]$gateway_ip"
```
mafonso: Do we still need to resort to the docker bridge address?  I thought it did work with localhost.

cwrw: This is for the selenium build rather than database, I can try again now, just to double check if it works with localhost, but we can keep as is otherwise

azure-pipelines.yml
```diff
@@ -29,39 +30,87 @@ resources:
...
+        env:
+          RAILS_ENV: test
+          DATABASE_URL: $(postgresUrl)
+
+      - script: |
+          bundle exec rspec --format RspecJunitFormatter --out /build/TEST-rspec.xml
+        displayName: 'Run Rspec tests'
+        env:
+          DATABASE_URL: $(postgresUrl)
+          SELENIUM_REMOTE_ADDRESS: localhost
```
cwrw: testing this now

Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apk add --update $BUILD_PACKAGES && \
 # Generate Assets
 WORKDIR /app
 COPY Gemfile Gemfile.lock .ruby-version ./
-RUN bundle install --clean --force --without "development" \
+RUN bundle install --clean --force --without "development test" \
```
smoothcontract: Splendid 👍

azure-pipelines.yml
```diff
@@ -29,39 +30,87 @@ resources:
...
-        # Todo: Look for support in the build box for the same ruby minor version.
-        - script: |
-            echo "2.6.2" > .ruby-version
-            sudo apt-get install libpq-dev
-            gem install bundler
-            bundle install --retry=3 --jobs=4
-          displayName: 'bundle install'
+      - script: |
+          sudo apt-get install libpq-dev bison libgdbm-dev libsqlite3-dev libyaml-dev sqlite3 libreadline6-dev
+          gpg --keyserver hkp://pool.sks-keyservers.net:80 --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
```
smoothcontract: What is this line for?

smoothcontract: ah, nevermind it's part of rvm install https://rvm.io/rvm/install

azure-pipelines.yml
cwrw: Unfortunately we still need it, so will keep as is for now

azure-pipelines.yml
```diff
@@ -29,39 +30,87 @@ resources:
...
+          path: $(BUNDLE_CACHE_FOLDER)
+          cacheHitVar: CACHE_RESTORED_BUNDLE
+        displayName: Cache bundle
 
-        - script: |
-            bundle exec rubocop --format clang
-          displayName: 'rubocop'
+      - script: |
+          gem install bundler
+          bundle install --jobs=4 --retry=3 --path $(BUNDLE_CACHE_FOLDER)
```
smoothcontract: Is the bundle that will be used to run the test suite? If so can `development` gems be excluded here?

azure-pipelines.yml
```diff
@@ -29,39 +30,87 @@ resources:
...
+      - script: |
+          bundle exec govuk-lint-sass app/webpacker/styles
+        displayName: 'GovUK lint'
+
+      - script: |
+          yarn install
+        displayName: 'yarn install'
+
+      - script: |
+          bundle exec rake db:create db:migrate
```
smoothcontract: I think would be faster to run `db:schema:load` and more reliable?

azure-pipelines.yml
mafonso: 😔   Ok, let's keep it for now.

azure-pipelines.yml
mafonso: Maybe this looks a bit cleaner? 
```
curl -sSL https://rvm.io/mpapis.asc | gpg --import -
curl -sSL https://rvm.io/pkuczynski.asc | gpg --import -
```
Since we are curling stuff into a shell, we might as well get the keys the same way

azure-pipelines.yml
cwrw: When I have some time I can see why that is, as ideally it should work

azure-pipelines.yml
```diff
@@ -29,39 +30,87 @@ resources:
...
+          path: $(BUNDLE_CACHE_FOLDER)
+          cacheHitVar: CACHE_RESTORED_BUNDLE
+        displayName: Cache bundle
 
-        - script: |
-            bundle exec rubocop --format clang
-          displayName: 'rubocop'
+      - script: |
+          gem install bundler
+          bundle install --jobs=4 --retry=3 --path $(BUNDLE_CACHE_FOLDER) --without "development"
```
smoothcontract: Very minor nitpick - `development` doesn't need to be quoted here

azure-pipelines.yml
mafonso: Unless it's selenium initiating the call. That would make sense as why it won't work.

azure-pipelines.yml
cwrw: I'd like to keep this as is as it's one line, and according to https://rvm.io/rvm/security#alternatives it's a less secure way to do it as follows

azure-pipelines.yml
cwrw: yea, selenium is initiating the call. I went through alot of documentation/blogs/articles about this, and some were querying `ifconfig` to get the ip or using some other way, and some localhost. Im ok with this for now although I agree it's not great

PR #122, opened 2019-08-06
GET-333 Make sure SplitIO does not fall over if it cannot connect

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-333

Rescue from SplitIO connection errors and default to all features
being false if we fail to connect to the API.

### DIFF
--- a/app/lib/flipflop/split_client_strategy.rb
+++ b/app/lib/flipflop/split_client_strategy.rb
@@ -1,6 +1,7 @@
 module Flipflop
   module Strategies
     class SplitClientStrategy < AbstractStrategy
+      attr_reader :client
       class << self
         def default_description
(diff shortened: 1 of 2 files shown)


app/lib/flipflop/split_client_strategy.rb
```diff
@@ -1,20 +1,22 @@
...
       class << self
         def default_description
           'External configuration via split.io client'
         end
       end
 
       def initialize(**options)
-        @client = factory(options).client
-
+        @client = factory(options)
```
tavy88: Is this object the same type now?

tavy88: Discussed all good.

PR #124, opened 2019-08-07
Persist user input postcode on the session

tavy88:
### Context

We want to persist the user input postcode on the session so that when I go to find courses I don't get a broken journey.

### JIRA

https://dfedigital.atlassian.net/browse/GET-349

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -18,7 +18,7 @@ class CoursesController < ApplicationController
   end
 
   def postcode
-    courses_params[:postcode]
+    @postcode ||= courses_params[:postcode] || session[:postcode]
   end
 
   def courses_params
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -8,7 +8,11 @@ class PagesController < ApplicationController
 
     @search = CourseGeospatialSearch.new(postcode: postcode)
 
-    location_eligibility_through_courses if postcode.present? && @search.valid?
+    if postcode.present? && @search.valid?
+      session[:postcode] = postcode
+
+      location_eligibility_through_courses
+    end
   rescue CourseGeospatialSearch::GeocoderAPIError
     redirect_to postcode_search_error_path
   end
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -21,7 +21,7 @@
             Postcode
           </label>
           <%= errors_tag @search, :postcode %>
-          <%= text_field_tag(:postcode, params[:postcode], class: css_classes_for_input(@search, :postcode, "search-input govuk-input govuk-!-width-one-half"), required: true) %>
+          <%= text_field_tag(:postcode, @postcode, class: css_classes_for_input(@search, :postcode, "search-input govuk-input govuk-!-width-one-half"), required: true) %>
           <%= button_tag('', class: 'search-button-results', name: nil, aria: { label: 'Search button' }) %>
         <% end %>
       <% end %>
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -45,7 +45,7 @@ Rails.application.configure do
   config.active_storage.service = :local
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
--- a/spec/features/find_training_courses_spec.rb
+++ b/spec/features/find_training_courses_spec.rb
@@ -12,7 +12,47 @@ RSpec.feature 'Find training courses', type: :feature do
     expect(page).to have_text('0 courses found')
   end
 
-  scenario 'User can find training courses near them' do
+  scenario 'The postcode is persisted on Maths courses page when user comes from training hub page' do
+    capture_user_location('NW6 1JF')
+
+    visit(training_hub_path)
+
+    click_on('Find a maths course')
+
+    expect(find_field('postcode').value).to eq 'NW6 1JF'
+  end
+
+  scenario 'The postcode is persisted on English courses page when user comes from training hub page' do
+    capture_user_location('NW6 1JF')
+
+    visit(training_hub_path)
+
+    click_on('Find an English course')
+
+    expect(find_field('postcode').value).to eq 'NW6 1JF'
+  end
+
+  scenario 'The postcode is persisted on maths courses page when user comes from math course overview page' do
+    capture_user_location('NW6 1JF')
+
+    visit(maths_course_overview_path)
+
+    click_on('Find a maths course')
+
+    expect(find_field('postcode').value).to eq 'NW6 1JF'
+  end
+
+  scenario 'The postcode is persisted on english courses page when user comes from an english course overview page' do
+    capture_user_location('NW6 1JF')
+
+    visit(english_course_overview_path)
+
+    click_on('Find an English course')
+
+    expect(find_field('postcode').value).to eq 'NW6 1JF'
+  end
+
+  scenario 'Users can find training courses near them' do
     Geocoder::Lookup::Test.add_stub(
       'NW6 8ET', [{ 'coordinates' => [0.1, 1] }]
     )
@@ -74,5 +114,13 @@ RSpec.feature 'Find training courses', type: :feature do
     visit(courses_path(topic_id: 'maths'))
 
     expect(page).not_to have_selector('#postcode[required]')
+  end
+
+  def capture_user_location(postcode)
+    enable_feature! :course_directory, :location_eligibility
+
+    visit(location_eligibility_path)
+    fill_in('postcode', with: postcode)
+    find('.govuk-button').click
   end
 end
--- a/spec/support/feature_flags.rb
+++ b/spec/support/feature_flags.rb
@@ -1,15 +1,17 @@
 module FeatureFlags
-  def enable_feature!(feature_name)
-    switch_feature!(feature_name, true)
+  def enable_feature!(*feature_names)
+    switch_feature!(feature_names, true)
   end
 
-  def disable_feature!(feature_name)
-    switch_feature!(feature_name, false)
+  def disable_feature!(*feature_names)
+    switch_feature!(feature_names, false)
   end
 
-  def switch_feature!(feature_name, enabled)
+  def switch_feature!(feature_names, enabled)
     test_strategy = Flipflop::FeatureSet.current.test!
-    test_strategy.switch!(feature_name, enabled)
+    feature_names.each do |feature|
+      test_strategy.switch!(feature, enabled)
+    end
   end
 end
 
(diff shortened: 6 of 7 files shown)


spec/support/feature_flags.rb
```diff
@@ -1,15 +1,17 @@
 module FeatureFlags
-  def enable_feature!(feature_name)
-    switch_feature!(feature_name, true)
+  def enable_feature!(*features)
```
tavy88: I kept the same name as I don't want to update all the places that are using this. This has a high risk and it would assume we need to do a full regression testing.

I don't want to add a `enable_features` method either as that would create a lot of redundancy.

app/controllers/courses_controller.rb
```diff
@@ -5,7 +5,7 @@ def index
     track_event(:courses_index_search, search: postcode) if postcode.present?
 
     @search = CourseGeospatialSearch.new(
-      postcode: postcode,
+      postcode: postcode || session[:postcode],
```
cwrw: if we were to add this on line 21 instead as so we can also include tracking:
```
  def postcode
    courses_params[:postcode] || session[:postcode]
  end
```

spec/controllers/pages_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
```
cwrw: We agreed to not add controller specs and I think we should stick to that

spec/features/find_training_courses_spec.rb
```diff
@@ -12,7 +12,27 @@
...
-  scenario 'User can find training courses near them' do
+  scenario 'Users can find training courses near them after feeding back the postcode from location eligibility page' do
+    enable_feature! :course_directory, :location_eligibility
+
+    Geocoder::Lookup::Test.add_stub(
+      'NW6 8ET', [{ 'coordinates' => [0.1, 1] }]
+    )
+
+    create(:course, latitude: 0.1, longitude: 1.001, topic: 'maths')
+    create(:course, latitude: 0.1, longitude: 2, topic: 'maths')
```
cwrw: we don't need the next two courses creation as they are too far, if you amend the longitude to be closer we can count instead 3 courses in the results page

app/controllers/courses_controller.rb
tavy88: I'm not sure if we're supposed to track the postcode from the session. My understanding was that tracking at this page is what they actually type in?

spec/controllers/pages_controller_spec.rb
tavy88: There is a reason I added this and it's an important unit test that checks the session is persisted. I find the rule of never adding a controller spec too rigid tbh. I am ok with not adding them when they do not bring any particular value. In this case, it does bring value.

spec/features/find_training_courses_spec.rb
tavy88: What's the value in doing that?

app/controllers/courses_controller.rb
smoothcontract: I think it would be cleaner to make the change @cwrw suggests - it nicely encapsulates retrieval of the postcode either from params or session, so keeps the search code clearer.

app/controllers/pages_controller.rb
```diff
@@ -8,7 +8,11 @@ def location_eligibility
 
     @search = CourseGeospatialSearch.new(postcode: postcode)
 
-    location_eligibility_through_courses if postcode.present? && @search.valid?
+    if postcode.present? && @search.valid?
+      session[:postcode] = postcode
```
smoothcontract: 👍

spec/features/find_training_courses_spec.rb
tavy88: I still want courses that do not match my search as well.

config/environments/production.rb
```diff
@@ -45,7 +45,7 @@
   config.active_storage.service = :local
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
smoothcontract: Interesting, I thought we already had this enabled. Presumably this works okay when deployed to Azure? My understanding is that SSL termination happens within Azure and only non secure traffic is forwarded to the Rails app.

app/controllers/courses_controller.rb
cwrw: I think there is value in tracking all postcodes that come through to this page, from the session and entered. But if there is doubt double check with the product owners to confirm

config/environments/production.rb
smoothcontract: Just an aside - is the comment above this line misleading? The session store config seems to be ignoring this and just checking `Rails.env.production?`

config/initializers/session_store.rb
```diff
@@ -0,0 +1 @@
+Rails.application.config.session_store :cookie_store, key: '_ghtr_session', secure: Rails.env.production?
```
smoothcontract: I wonder if the `secure:` option here is breaking convention - does this not just default to whatever `Rails.application.config.force_ssl` is set to (as suggested by the comment in file above). I haven't checked the framework implementation to be sure.

app/controllers/courses_controller.rb
tavy88: @smoothcontract thoughts?

spec/controllers/pages_controller_spec.rb
smoothcontract: Request spec would be preferable I think. You can check all of the same things there with minimal changes.

spec/features/find_training_courses_spec.rb
smoothcontract: One or two maximum should be enough courses to demonstrate this I think. Three seems excessive for the purposes of demonstrating postcode propagation.

cwrw: ok that's fair if it's intentional

app/controllers/courses_controller.rb
cwrw: Having a look locally, we should populate the search bar with the postcode from the session, to avoid confusion when landing on teh page

tavy88: What is passed on the session should be tracked at the eligibility page level, when the user inputs that postcode imo.

spec/features/find_training_courses_spec.rb
smoothcontract: Actually it might be sufficient to just check that the postcode input is populated with the value from the session - presumably that does get populated? If that's the case then this feature spec arguably doesn't need to actually populate training courses 🤔

app/controllers/courses_controller.rb
cwrw: Also what happens when I update my postcode/search for a different one on this page, is it supposed to keep the old session or update it to the new postcode?

spec/support/feature_flags.rb
smoothcontract: I like the change to accept multiple feature names, but prefer the original `feature_names` rather than `feature` - that more explicitly defines what should be passed here.

app/controllers/courses_controller.rb
smoothcontract: We're tracking the fact that users searched for courses - it doesn't matter whether the postcode came from session or params really. Each time they search a customEvent is added to AAI and that's fine. So abstracting the postcode into the `postcode` method accomplishes that nicely whilst keeping the code easy to follow.

config/initializers/session_store.rb
cwrw: Ive been looking at other rails examples and seems about right but let me check again to confirm

spec/support/feature_flags.rb
tavy88: Ok

app/controllers/courses_controller.rb
smoothcontract: That behaviour didn't get defined in the requirements I believe. I'm happy that we leave this unchanged so that the params (from the updated postcode entered in the search input) would override the session value for the duration of the request and to populate the course results, but doesn't overwrite the value stored in the session (which, after all represents the postcode they provided in location eligibility page). We don't need to update the session when searching for courses. Therefore if the user visits the courses page, changes the postcode they would see results for the postcode entered, but if they then return later in the same session the page would revert to showing courses for the postcode originally entered on location eligibility page.

Hopefully that makes sense. This can then be user tested to see if users find it intuitive. No need to guess more complex requirements at this stage.

spec/features/find_training_courses_spec.rb
tavy88: You do need to check if you get results. Because if you get an error the count will be 2.

config/environments/production.rb
tavy88: This comment is generated by rails. I didn't add that.

spec/support/feature_flags.rb
tavy88: I don't see necessarily where the confusion is but will do the change.

smoothcontract: Thankyou 👍

config/environments/production.rb
tavy88: Will test this once deployed.

app/views/pages/english_overview.html.erb
```diff
@@ -20,7 +20,7 @@
     <h2 class="govuk-heading-m">The right course for you</h2>
     <p class="govuk-body">There are lots of English courses available for every level of ability. The course provider will help assess what level of course is right for you. You’ll also be able to choose a course that offers flexible learning or part time study to fit around your life.</p>
     <p class="govuk-body">Many courses are available for free in your area.</p>
-    <%= link_to 'Find an English course', courses_path('english'), class: 'govuk-button govuk-!-margin-top-5', target: '_blank' %>
+    <%= link_to 'Find an English course', courses_path('english', postcode: session[:postcode]), class: 'govuk-button govuk-!-margin-top-5', target: '_blank' %>
```
cwrw: instead of passing the session around which isn't needed, we can amend the postcode params in the controller to equal the session if it doesn't exist

tavy88: I did another change for that.

app/views/courses/index.html.erb
```diff
@@ -21,7 +21,7 @@
             Postcode
           </label>
           <%= errors_tag @search, :postcode %>
-          <%= text_field_tag(:postcode, params[:postcode], class: css_classes_for_input(@search, :postcode, "search-input govuk-input govuk-!-width-one-half"), required: true) %>
+          <%= text_field_tag(:postcode, params[:postcode] || session[:postcode], class: css_classes_for_input(@search, :postcode, "search-input govuk-input govuk-!-width-one-half"), required: true) %>
```
smoothcontract: one minor suggestion would be to add an instance variable in the controller instead to DRY this up a little.. something like

```
@postcode = postcode
```

and then reference that instance variable here. The `postcode` method already contains this logic.

tavy88: 👍

PR #126, opened 2019-08-08
Test -> CI -> QA

mafonso:
### Context

With the replacement of the 'test' environment with 'ci' we need to port the split.io key

- Move split key from test to ci
- Remove test environment


mafonso:
ℹ️ The key was adjusted in KeyVault accordingly

azure/config/ci.parameters.json
```diff
@@ -29,6 +29,13 @@
             "secretName": "ci-secretKeyBase"
           }
         },
+        "splitApiKey": {
```
cwrw: actually let's also add the feedback form here as well

cwrw:
```
"userFeedbackSmartSurveyLink": {
            "value": "https://www.smartsurvey.co.uk/s/8IBQ1/"
        }
```

mafonso: 👍 will add

azure/config/ci.parameters.json
```diff
@@ -41,6 +48,9 @@
         "keyVaultResourceGroup": {
             "value": "s108d01-shared"
         },
+        "userFeedbackSmartSurveyLink": {
+          "value": "https://www.smartsurvey.co.uk/s/8IBQ1/"
+        }
```
cwrw: 🦂

cwrw: `,`

mafonso:
Rename first stage environment as QA as per team agreement.
Secrets in keyvault ajusted to reflect this.

azure/config/dev1.parameters.json
```diff
@@ -1,55 +1,59 @@
...
-        "keyVaultName": {
-            "value": "s108d01-shared-kv-01"
-        },
-        "keyVaultResourceGroup": {
-            "value": "s108d01-shared"
-        },
-        "userFeedbackSmartSurveyLink": {
-            "value": "https://www.smartsurvey.co.uk/s/8IBQ1/"
-        }
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
```
smoothcontract: ARG! Whitespace changes make this really hard to diff :(

azure/config/dev1.parameters.json
```diff
@@ -1,55 +1,59 @@
...
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "databaseName": {
+      "value": "ghtr"
+    },
+    "databaseUsername": {
+      "reference": {
+        "keyVault": {
+          "id": "/subscriptions/f35bd249-3028-48bd-9aab-d00e9d1a56b7/resourceGroups/s108d01-shared/providers/Microsoft.KeyVault/vaults/s108d01-shared-kv-01"
```
smoothcontract: Is it possible to DRY up the keyvault so it's not repeated throughout the file?

mafonso: Unfortunately it does not seem possible at this level, this is the source parameter file which does not allow for interpolation on the parameters/variables as it is the literal source for the nested template to interpolate.

I even tried to use `subscription().subscriptionId` instead of the literal ID as we want to target the vault in the same subscription, it does not evaluate the expression at this stage. 

https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-keyvault-parameter

PR #127, opened 2019-08-08
Get 226 import job growth

smoothcontract:
### Context
We need to display job growth data for a couple of stories.

* Adds extra columns to persist job growth data (just a decimal) along with SOC codes originating from NCS to make future data matching easier.
* Adds a new service to update these attributes from a spreadsheet (uses Roo gem)
* Adds a rake task to invoke the new service
* Extended ActiveAdmin to allow editing growth and viewing/exporting new attributes
* Tweaked stoopid rubocop rule around specs/expectations 😛 

https://github.com/roo-rb/roo

### Guidance to review
To test locally you'll need the magic spreadsheet. That's been tweaked slightly (renamed second column to make names unique) from the original version. I can provide if required. The fixture file is a cut down version with only one sheet and a few rows.

Just run the rake task and observe output e.g.

```
rails data_import:import_job_growth['Job growth 05082019.xlsx']
```


app/admin/job_profiles.rb
```diff
@@ -10,6 +10,7 @@
     filter :categories
     filter :skills
     filter :related_job_profiles
+    filter :extended_soc
```
smoothcontract: I didn't add a filter for `soc` as it's a subset of `extended_soc` so that can be used instead.

app/services/job_profile_import_service.rb
```diff
@@ -0,0 +1,50 @@
+class JobProfileImportService
+  GROWTH_COLUMN_HEADINGS = {
+    soc: 'SOCCode',
+    extended_soc: 'NCS SOCCode',
```
smoothcontract: This is the column that was renamed. It's nicer to map to the column names rather than assume a specific ordering.

db/migrate/20190807131052_add_soc_and_growth_to_job_profiles.rb
```diff
@@ -0,0 +1,7 @@
+class AddSocAndGrowthToJobProfiles < ActiveRecord::Migration[5.2]
+  def change
+    add_column :job_profiles, :soc, :string
+    add_column :job_profiles, :extended_soc, :string
+    add_column :job_profiles, :growth, :decimal
```
smoothcontract: This will be mapped into a user friendly description as part of the presentation stories. Here we're just persisting the raw data.

spec/.rubocop.yml
```diff
@@ -16,6 +16,8 @@ RSpec/DescribeClass:
   - tasks/**/*.rb
 RSpec/ExampleLength:
   Enabled: false
+RSpec/MultipleExpectations:
+  Enabled: false
```
smoothcontract: 🙀

tavy88: FYI: https://github.com/abinoda/rspec-best-practices#only-one-expectation-per-example

smoothcontract: I disagree with that "best practice" for performance reasons 😛

app/models/job_profile.rb
```diff
@@ -10,6 +10,10 @@ class JobProfile < ApplicationRecord
                           foreign_key: :job_profile_id,
                           association_foreign_key: :related_job_profile_id
 
+  def self.find_by_name(name)
+    where('lower(name) = ?', name.downcase).first
+  end
+
```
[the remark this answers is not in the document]
smoothcontract: I don't think it needs explicit coverage - it's covered by the new service that uses this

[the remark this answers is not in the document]
tavy88: Ok.

app/models/job_profile.rb
```diff
@@ -10,6 +10,10 @@ class JobProfile < ApplicationRecord
                           foreign_key: :job_profile_id,
                           association_foreign_key: :related_job_profile_id
 
+  def self.find_by_name(name)
+    where('lower(name) = ?', name.downcase).first
```
cwrw: we should have probably save everything lowercase initially

app/services/job_profile_import_service.rb
```diff
@@ -0,0 +1,50 @@
...
+    growth: 'Recent growth (% change to March 2018)'
+  }.freeze
+
+  def initialize
+    not_production!
+    @errors = []
+  end
+
+  def import_growth(filename)
+    file = Roo::Spreadsheet.open(filename)
```
cwrw: if we used a normal csv file, we can use the native `CSV.parse(filename, headers: true)` instead, what do you think?

lib/tasks/data_import/import_job_growth.rake
```diff
@@ -0,0 +1,14 @@
+namespace :data_import do
+  # bin/rails data_import:import_job_growth[filename.xlsx]
+  desc 'Import job growth data from Excel spreadsheet'
+  task :import_job_growth, [:filename] => [:environment] do |_task, args|
+    filename = args[:filename]
+    abort 'No spreadsheet filename provided' unless filename.present?
+
+    print "Importing job growth data from #{filename}..."
```
cwrw: does puts not work here?

spec/tasks/import_job_growth_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  it 'raises error if invalid filename' do
+    expect { task.execute(filename: 'foo.xlsx') }.to raise_exception(IOError)
+  end
+
+  it 'updates matching job profiles' do
+    expect { task.execute(filename: path) }.not_to raise_exception
+
+    expect(ceo.soc).to eq '1115'
```
cwrw: instead of the rubocop setting change, I would prefer we use the `have_attributes` matcher instead
`expect(ceo.extended_soc).to have_attributes(...)`
https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/have-attributes-matcher

spec/services/job_profile_import_service_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  describe '#import_growth' do
+    it 'raises error if invalid filename' do
+      expect { importer.import_growth('foo.xlsx') }.to raise_exception(IOError)
+    end
+
+    it 'updates matching job profiles' do
+      expect { importer.import_growth(path) }.not_to raise_exception
+
+      expect(ceo.soc).to eq '1115'
```
cwrw: same here regarding `have_attributes` matcher (See comment below)

spec/.rubocop.yml
cwrw: 🔥 see comments below about not needing this anymore

app/services/job_profile_import_service.rb
cwrw: talked in person, keeping as is

app/models/job_profile.rb
smoothcontract: No, we shouldn't. These are the displayed job names so are as scraped from NCS so if we made them lower case we would then have the problem of displaying them correctly again.

app/services/job_profile_import_service.rb
smoothcontract: If the original file was a single sheet then yep, I'd agree. But it's a mammoth spreadsheet with loads of tabs. Easier to just roll with what we're given I think.

db/schema.rb
```diff
@@ -75,6 +75,9 @@
     t.integer "salary_min"
     t.integer "salary_max"
     t.string "alternative_titles"
+    t.string "soc"
+    t.string "extended_soc"
+    t.decimal "growth"
```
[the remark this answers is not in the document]
smoothcontract: Nope, we want the default to be nil so that we can determine which job profiles do not have growth information (and hence will not display it). `0` would actually be a valid value for growth - it can be positive or negative.

lib/tasks/data_import/import_job_growth.rake
smoothcontract: `puts` does but when I worked on the other rake tasks I discovered it also echos to the rspec output. `print` writes to the console when run interactively but is quiet in the spec output so works well.

spec/tasks/import_job_growth_spec.rb
smoothcontract: Ah, nice. Seems like a good alternative. I'll look into that 👍

app/models/job_profile.rb
cwrw: initially I thought they just capitalised first letter, but apparently its custom to what the occupation is, ah well

db/schema.rb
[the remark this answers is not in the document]
cwrw: ah interesting, so it's stable if it's zero. Im not a fan of having a column with multiple types but I guess we need it 👍

PR #129, opened 2019-08-09
GET-231 Skills Builder

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-231

Add Skills Builder as a feature flagged service. Users get to this page from the check your skills results page, after clicking on any job. 
To allow this new behaviour a service object: skills builder takes in skill ids selected and adds them to the user session. The skills builder validates if any skills are selected and if not returns the appropriate validation message. The skills builder also returns the list of ids of selected by user to be used in the checkboxes themselves to maintain what the user selected/didn't select

deployed for testing: https://dev3.nrs-ghtr.org.uk

### DIFF
--- a/app/controllers/job_profiles_skills_controller.rb
+++ b/app/controllers/job_profiles_skills_controller.rb
@@ -1,19 +1,12 @@
 class JobProfilesSkillsController < ApplicationController
   def index
+    skills_builder.build
+
     if skills_valid?
-      redirect skills_path
+      redirect_to(skills_path(job_profile_id: skills_params[:job_profile_id]))
     else
-      @skills = job_profile.skills
-      @skills.errors.add(:postcode, 'Select at least one skill')
+      render 'job_profiles/skills/index'
     end
-  end
-
-  def current_job_skills
-
-  end
-
-  def your_skills
-    @skills = job_profile.skills
   end
 
   private
@@ -24,11 +17,19 @@ class JobProfilesSkillsController < ApplicationController
     )
   end
 
-  def skills_params
-    params.permit(:job_profile_id, skill_ids: [])
+  def skills_builder
+    @skills_builder ||= SkillsBuilder.new(
+      skills_params: skills_params[:skill_ids],
+      job_profile_skills: job_profile.skills,
+      user_session: session
+    )
   end
 
   def skills_valid?
-    skills_params[:skill_ids].present? && skills_params[:skill_ids].count > 1
+    skills_params[:skill_ids].present? && skills_builder.valid?
+  end
+
+  def skills_params
+    params.permit(:job_profile_id, skill_ids: [])
   end
 end
--- a/app/controllers/job_profiles_skills_controller.rb
+++ b/app/controllers/job_profiles_skills_controller.rb
@@ -1,0 +1,34 @@
+class JobProfilesSkillsController < ApplicationController
+  def index
+    if skills_valid?
+      redirect skills_path
+    else
+      @skills = job_profile.skills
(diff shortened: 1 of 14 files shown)


app/controllers/skills_controller.rb
```diff
@@ -1,6 +1,12 @@
 class SkillsController < ApplicationController
   def index
-    @skills = job_profile.skills
+    if Flipflop.skills_builder?
+      job_profile
+      @skills = Skill.where(id: session[:skill_ids])
+      render 'index_v2'
```
cwrw: this will be changed to `index` once we remove the feature flag

app/controllers/job_profiles_skills_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  def skills_builder
+    @skills_builder ||= SkillsBuilder.new(
+      skills_params: skills_params[:skill_ids],
+      job_profile_skills: job_profile.skills,
+      user_session: session
+    )
+  end
+
+  def skills_valid?
+    skills_params[:skill_ids].present? && skills_builder.valid?
```
smoothcontract: I'm not sure why you would need to double check the params again here. Does just using `skills_builder.valid?` work just as well (as it already contains same validation logic). And if it does this probably doesn't need to be a separate method and could be added on line 5 instead.

app/views/skills/index_v2.html.erb
```diff
@@ -0,0 +1,42 @@
...
+        </thead>
+        <tbody class="govuk-table__body">
+          <% @skills.each do |skill| %>
+            <tr class="govuk-table__row">
+              <td class="govuk-table__cell govuk-!-padding-bottom-4 govuk-!-padding-top-4"><%= skill.name %></th>
+            </tr>
+          <% end %>
+        </tbody>
+      </table>
+      <%= link_to('Find out what you can do with these skills', '#', class: 'govuk-button') %>
```
smoothcontract: Maybe add a TODO comment here so this can be implemented in the skills matcher story.

spec/features/skills_builder_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  let!(:job_profile) do
+    create(
+      :job_profile,
+      name: 'Hitman',
+      skills: [
+        create(:skill, name: 'Chameleon-like blend in tactics'),
+        create(:skill, name: 'License to kill'),
+        create(:skill, name: 'Baldness')
```
smoothcontract: 😂

app/controllers/job_profiles_skills_controller.rb
cwrw: if we removed this check, we would always return an invalid form on first hitting the page, so in order to make sure we actually want to validate we first make sure skill ids are present then we validate against the builder

smoothcontract: So you're saying you explicitly want to avoid calling `.valid?` on the builder in that scenario (i.e. when user first lands on the page) ?

cwrw: yea

smoothcontract: Actually that description doesn't seem to match the implementation of the code. It looks like if no skills params are present then this method returns false and user gets redirected?

cwrw: if this method is false, we go to the skills builder page which is `render('job_profiles/skills/index')`, Im not sure which redirect your talking about?

cwrw: talked in person and all is good

PR #131, opened 2019-08-12
GET-363 Separate different job profiles in skill builder

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-363

When a user navigates to a job profile and selects a list of skills, the scope of those skills are only to the job profile. If a user navigates back to the search results and picks a different job profile, they should see all skills ticked unless they have visited that page before. Moving towards a skills scoped to a job profile we make it easier to move to the skill builder v2 which will now only require FE changes only.

Note that because sessions change the format of hashes to JSON, all ids are changed to strings to make things consistent throughout

ready to test on https://dev2.nrs-ghtr.org.uk

### DIFF
--- a/app/controllers/skills_controller.rb
+++ b/app/controllers/skills_controller.rb
@@ -2,7 +2,7 @@ class SkillsController < ApplicationController
   def index
     if Flipflop.skills_builder?
       job_profile
-      @skills = Skill.find(session[:skill_ids])
+      @skills = Skill.find(skill_ids)
       render 'index_v2'
     else
       @skills = job_profile.skills
@@ -15,6 +15,10 @@ class SkillsController < ApplicationController
     @job_profile ||= JobProfileDecorator.new(
       JobProfile.find_by(slug: skills_params[:job_profile_id])
     )
+  end
+
+  def skill_ids
+    session.fetch(:job_profile_skills, {})[job_profile.id.to_s] || []
   end
 
   def skills_params
--- a/app/models/skills_builder.rb
+++ b/app/models/skills_builder.rb
@@ -1,23 +1,26 @@
 class SkillsBuilder
   include ActiveModel::Validations
 
-  attr_reader :skills_params, :job_profile_skills, :user_session
+  attr_reader :skills_params, :job_profile, :user_session
   validate :skill_ids_presence
 
-  def initialize(skills_params:, job_profile_skills:, user_session:)
+  def initialize(skills_params:, job_profile:, user_session:)
     @skills_params = skills_params
-    @job_profile_skills = job_profile_skills
+    @job_profile = job_profile
     @user_session = user_session
+    @user_session[:job_profile_skills] = user_session[:job_profile_skills] || {}
   end
 
   def build
     return unless skills_params.present?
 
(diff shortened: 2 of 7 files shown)


app/controllers/skills_controller.rb
```diff
@@ -17,6 +17,10 @@ def job_profile
     )
   end
 
+  def skill_ids
+    session[:job_profile_skills] && session[:job_profile_skills][job_profile.id.to_s] || []
```
smoothcontract: Could you use `.dig` here or maybe `&` safe navigator?

spec/models/skills_builder_spec.rb
```diff
@@ -3,58 +3,104 @@
...
 
-      expect(builder.user_session).to eq(skill_ids: [1])
+      builder.build
+      expect(builder.user_session).to eq(job_profile_skills: { job_profile.id.to_s => [1] })
     end
 
     it 'overrides existing user session with new skill param ids' do
+      job_profile = create(:job_profile, skills: [create(:skill)])
+      builder = described_class.new(
+        skills_params: ['1', '', '5', '6'],
```
smoothcontract: How would blank skill id's end up in the user session? Is this a valid scenario to code/test for or just added complexity?

cwrw: spoke in person, the empty string comes from the hidden field in the form set to nil, and is needed to determine if we navigate to the page on first instance

app/controllers/skills_controller.rb
cwrw: Unfortunately none of those methods work on `ActionDispatch::Request::Session` :(

smoothcontract: I was thinking something like this...

```ruby
session[:job_profile_skills]&.[job_profile.id.to_s] || []
```

But that looks kinda funky, if indeed it is valid syntax. So maybe leave as is.

spec/models/skills_builder_spec.rb
smoothcontract: 👍

app/controllers/skills_controller.rb
tavy88: @cwrw Can you do maybe:

```ruby
session.fetch(: job_profile_skills, {})[:job_profile.id.to_s]
```
?

app/models/skills_builder.rb
```diff
@@ -1,23 +1,26 @@
...
 
-    user_session[:skill_ids] = formatted_skill_params
+    user_session[:job_profile_skills][job_profile.id.to_s] = formatted_skill_params
   end
 
   def skill_ids
-    user_session[:skill_ids] || job_profile_skills.pluck(:id)
+    @skill_ids ||=
+      user_session[:job_profile_skills][job_profile.id.to_s] ||
+      job_profile.skills.pluck(:id)
```
tavy88: This method is doing too many things. Wonder if you would consider getting close to a single responsibility principle?

cwrw: hmm Im not sure I agree, I think this method is returning the skill ids, either from the session or from the job profile. I'd rather keep as is for now until it becomes more complex

PR #134, opened 2019-08-13
Update the build name on first stage

mafonso:
Update the build name on first stage.
The pipeline name is not being updated on first stage, so it keeps the internal ID until stage 'build'

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,6 +44,8 @@ stages:
       - script: |
           gateway_ip=$(docker network inspect bridge --format='{{range .IPAM.Config}}{{.Gateway}}{{end}}')
           echo "##vso[task.setvariable variable=GatewayIP;]$gateway_ip"
+          git_sha=$(git rev-parse --short HEAD)
+          echo "##vso[build.updatebuildnumber]$git_sha"
         displayName: 'Set Task Variables'
 
       - script: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -125,10 +125,6 @@ stages:
       vmImage: 'ubuntu-16.04'
 
     steps:
-      - script: |
-          git_sha=$(git rev-parse --short HEAD)
-        displayName: 'Set Image Tags'
-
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -44,6 +44,8 @@ stages:
       - script: |
           gateway_ip=$(docker network inspect bridge --format='{{range .IPAM.Config}}{{.Gateway}}{{end}}')
           echo "##vso[task.setvariable variable=GatewayIP;]$gateway_ip"
+          git_sha=$(git rev-parse --short HEAD)
+          echo "##vso[build.updatebuildnumber]$git_sha"
```
cwrw: can we set this to a global variable, so we can reuse in second build stage?

azure-pipelines.yml
```diff
@@ -125,7 +127,6 @@ stages:
     steps:
       - script: |
           git_sha=$(git rev-parse --short HEAD)
```
cwrw: we probably don't need this whole step either

mafonso: We do, because the `git_sha` won't be available in the second stage (it's another job, so it's a new VM)

azure-pipelines.yml
mafonso: It can only be defined in runtime as we need to get it's value from a command output. I don't see a way of defining it outside a job.

mafonso:
> can you also ammend https://github.com/DFE-Digital/get-help-to-retrain/pull/134/files#diff-fec826feae04e51c0d94076385408bdcL203 to QA instead of CI?

Yeah, It's outside the scope of this PR, but I'll make a new one.

azure-pipelines.yml
mafonso: But we can use $(Build.BuildNumber) after this step in the following jobs

azure-pipelines.yml
mafonso: Yeah, since we were already using $(Build.BuildNumber), it can indeed be removed.

PR #135, opened 2019-08-14
Add support for importing courses spreadsheet

smoothcontract:
### Context
Phase 2 requires real training course data for various regions for the roll out.

Adds new rake task and corresponding service to replace existing course data with content from spreadsheet. Spreadsheet has multiple tabs, each with identical structure but courses from various regions.

### Guidance to review
The relevant spreadsheet is linked in the Jira ticket (it's in a Google Drive folder named "P2 Training Providers") https://dfedigital.atlassian.net/browse/GET-339

Download that file and then run rake task as per README. This will remove any existing courses.


app/models/course.rb
```diff
@@ -1,12 +1,15 @@
 class Course < ApplicationRecord
-  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }, allow_blank: true
```
smoothcontract: This was breaking with blank email addresses; decision has been made not to import email addresses from the spreadsheet for now as they're not going to be shown on the UI for Phase 2.

app/services/course_import_service.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def courses_without_geocoding
+    Course.not_geocoded
+  end
+
+  private
+
+  def create_course(data)
+    # Cleanup data format
+    data.each do |_key, value| value&.strip! end
```
tavy88: I think the convention is that one-liners should use `{}` ?

```ruby
data.each { |_key, value| value&.strip! }
```

smoothcontract: Well, yes it is. I had it as you suggest then ran rubocop autocorrect and it changed this 🤦‍♂

smoothcontract: Arrgh! rubocop now reports a cop violation for the preferred syntax.

tavy88: Really? What is the violation - this is weird...

app/services/course_import_service.rb
```diff
@@ -0,0 +1,64 @@
...
+    title: 'title',
+    provider: 'provider',
+    url: 'url (course url)',
+    address_line_1: 'address line 1',
+    address_line_2: 'address line 2 (optional)',
+    town: 'town',
+    county: 'county',
+    postcode: 'postcode',
+    topic: 'type (Maths / English)',
+    phone_number: 'tel',
```
cwrw: can we move this to mirror order in spreadsheet, as telephone comes after provider?

app/services/course_import_service.rb
```diff
@@ -0,0 +1,64 @@
...
+        create_course(data)
+      end
+    end
+  end
+
+  def import_stats
+    {
+      courses_total: Course.count,
+      courses_with_geocoding: Course.geocoded.count,
+      courses_without_geocoding: Course.not_geocoded.count,
```
cwrw: can we use `courses_without_geocoding.count` here?

lib/tasks/data_import/import_courses.rake
```diff
@@ -0,0 +1,23 @@
...
+    importer.import(filename)
+
+    print "\nResults:\n#{importer.import_stats}"
+    if importer.errors.any?
+      print "\nErrors:"
+      importer.errors.each { |error| print error.inspect }
+    end
+    if importer.courses_without_geocoding.any?
+      print "\nGeocoding failures:"
+      importer.courses_without_geocoding.each { |course| print course.inspect }
```
cwrw: 👍

spec/services/course_import_service_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect { importer.import('foo.xlsx') }.to raise_exception(IOError)
+    end
+
+    it 'does not raise error with valid filename' do
+      expect { importer.import(path) }.not_to raise_exception
+    end
+
+    it 'creates courses' do
+      importer.import(path)
+      expect(course).to have_attributes(title: 'Basic English', provider: 'Acme Training Inc.', url: 'english url',
```
cwrw: can we add telephone here as well?

spec/tasks/import_courses_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  it 'raises error if invalid filename' do
+    expect { task.execute(filename: 'foo.xlsx') }.to raise_exception(IOError)
+  end
+
+  it 'does not raise error with valid filename' do
+    expect { task.execute(filename: path) }.not_to raise_exception
+  end
+
+  it 'creates courses' do
+    expect { task.execute(filename: path) }.to change(Course, :count).by(2)
```
cwrw: I'm thinking that maybe we should fail the task if geocoding fails?
If not can we add a test for teh geocoding error message?

lib/tasks/data_import/import_courses.rake
```diff
@@ -0,0 +1,23 @@
+namespace :data_import do
+  # bin/rails data_import:import_courses[filename.xlsx]
+  desc 'Import courses from Excel spreadsheet'
+  task :import_courses, [:filename] => [:environment] do |_task, args|
+    filename = args[:filename]
+    abort 'No spreadsheet filename provided' unless filename.present?
+
+    print "Importing courses from #{filename}..."
+    Course.delete_all
```
cwrw: Im not so sure about this, if the import fails halfway through we would have deleted all courses on the instance

app/services/course_import_service.rb
smoothcontract: The order of the spreadsheet doesn't matter - this matches by column headers and follows the order that attributes are defined in the schema. Either way I don't think it's worth changing.

spec/tasks/import_courses_spec.rb
smoothcontract: The intent is that the rake task is run manually so we can review the error output. I'm not sure how easy it would be to test for specific text output.

lib/tasks/data_import/import_courses.rake
smoothcontract: Discussed offline - rake task will be run against dev or test environment and monitored. Once the results are satisfactory, entire database then gets exported and imported to downstream environments. So there's no risk of breaking production, only dev/test environments. This could be improved in future perhaps but would need some thought to make this more robust; I'm not sure a transaction is a great solution when the volume of courses increases. Keep in mind this is also temporary to be replaced with the new courses API.

cwrw: Due to time restrictions and as long as this is never run on the live instance, I can go with this for now

PR #136, opened 2019-08-14
GET-232 skill matcher

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-232

Add skill matcher, which accepts a session with a set of skills, and optionally a job profile id to scope to for version 1.

If there is no job profile id, the skills matcher finds all job profile ids matching those skills, and groups them by skills matched count of job profiles as a subquery.

Then by grouping on the count of skills matched we get all job profile ids that have the same
score, and order them in ascending order of name, as well as on frequency. Finally we select
the records scoped to the fields needed and expand the ids of the ordered group maintaining
ordinality.

For the score we create a new hash from the ids and calculate the score mapped to a job profile id.
The scoring system chosen counts the skills matched in the job profile / the total count of skills selected %

### DIFF
--- a/app/controllers/skills_matcher_controller.rb
+++ b/app/controllers/skills_matcher_controller.rb
@@ -1,6 +1,6 @@
 class SkillsMatcherController < ApplicationController
   def index
-    @results = skills_matcher.match.page(params[:page])
+    @results = Kaminari.paginate_array(skills_matcher.match).page(params[:page]).per(10)
     @scores = skills_matcher.job_profile_scores
     @job_profiles = @results.map { |job_profile|
       JobProfileDecorator.new(job_profile)
--- a/app/controllers/skills_matcher_controller.rb
+++ b/app/controllers/skills_matcher_controller.rb
@@ -1,0 +1,8 @@
+class SkillsMatcherController < ApplicationController
+  def index
+    @results = JobProfile.all.includes(:categories).page(params[:page])
+    @job_profiles = @results.map { |job_profile|
+      JobProfileDecorator.new(job_profile)
+    }
+  end
+end
--- a/app/controllers/skills_matcher_controller.rb
+++ b/app/controllers/skills_matcher_controller.rb
@@ -1,8 +1,15 @@
 class SkillsMatcherController < ApplicationController
   def index
-    @results = JobProfile.all.includes(:categories).page(params[:page])
+    @results = skills_matcher.match.page(params[:page])
+    @scores = skills_matcher.job_profile_scores
     @job_profiles = @results.map { |job_profile|
       JobProfileDecorator.new(job_profile)
     }
+  end
+
+  private
+
+  def skills_matcher
+    @skills_matcher ||= SkillsMatcher.new(user_session: session)
   end
 end
--- a/app/controllers/skills_matcher_controller.rb
+++ b/app/controllers/skills_matcher_controller.rb
@@ -1,6 +1,6 @@
 class SkillsMatcherController < ApplicationController
   def index
-    @results = Kaminari.paginate_array(skills_matcher.match).page(params[:page]).per(10)
+    @results = Kaminari.paginate_array(skills_matcher.match).page(params[:page])
     @scores = skills_matcher.job_profile_scores
     @job_profiles = @results.map { |job_profile|
       JobProfileDecorator.new(job_profile)
--- a/app/decorators/job_profile_decorator.rb
+++ b/app/decorators/job_profile_decorator.rb
@@ -89,13 +89,13 @@ class JobProfileDecorator < SimpleDelegator # rubocop:disable Metrics/ClassLengt
     return unless score
 
     if score <= 25
-      'low'
+      'Low'
     elsif score > 25 && score <= 50
-      'reasonable'
+      'Reasonable'
     elsif score > 50 && score <= 75
-      'good'
+      'Good'
     elsif score > 75
-      'excellent'
+      'Excellent'
     end
   end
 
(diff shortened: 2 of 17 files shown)


app/decorators/job_profile_decorator.rb
```diff
@@ -63,6 +63,34 @@ def section(xpath: nil)
     @doc.to_html.gsub(%r{<a.*?>(.+?)</a>}, '\1').concat(separator_line)
   end
 
+  def recent_job_growth # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+    return unless growth
+
+    if growth <= -5
```
cwrw: to be removed after @tavy88 pr

app/models/skills_matcher.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def match
+    return JobProfile.none unless user_session[:job_profile_skills].present?
+
+    build_query
+  end
+
+  def job_profile_scores
+    @job_profile_scores ||= build_query.each_with_object({}) do |job_profile, hash|
+      hash[job_profile.id] = job_profile.skills_matched.to_f / user_skill_ids.count * 100
```
smoothcontract: Potential optimisation here - assign a local variable `user_skill_id_count` outside this loop

app/controllers/skills_matcher_controller.rb
```diff
@@ -0,0 +1,15 @@
+class SkillsMatcherController < ApplicationController
+  def index
+    @results = Kaminari.paginate_array(skills_matcher.match).page(params[:page]).per(10)
```
tavy88: I think you can set the `per_page` at the model level rather than hardcode it here.

app/decorators/job_profile_decorator.rb
tavy88: Fighting to get that in atm.

app/models/skills_matcher.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def match
+    return JobProfile.none unless user_session[:job_profile_skills].present?
+
+    build_query
+  end
+
+  def job_profile_scores
+    user_skills_ids_count = user_skill_ids.count
```
tavy88: I recommend this: https://stackoverflow.com/questions/6083219/activerecord-size-vs-count

spec/features/skills_matcher_feature.rb
```diff
@@ -0,0 +1,96 @@
+require 'rails_helper'
+
+RSpec.feature 'Skills matcher', type: :feature do
+  let!(:skill1) { create(:skill, name: 'Chameleon-like blend in tactics') }
+  let!(:skill2) { create(:skill, name: 'License to kill') }
+  let!(:skill3) { create(:skill, name: 'Baldness') }
+
+  let!(:job_profile1) do
```
tavy88: I think you can use `:job_profile` here.

spec/models/skills_matcher_spec.rb
```diff
@@ -0,0 +1,262 @@
...
+    it 'arranges job profiles in alphabetical order as well as matching skills order' do
+      skill1 = create(:skill)
+      skill2 = create(:skill)
+      skill3 = create(:skill)
+      job_profile1 = create(:job_profile, skills: [skill1, skill2, skill3])
+      job_profile2 = create(:job_profile, name: 'Researcher', skills: [skill2])
+      job_profile3 = create(:job_profile, name: 'Beekeeper', skills: [skill1, skill2, skill3])
+      job_profile4 = create(:job_profile, name: 'Admin', skills: [skill1, skill2, skill3])
+      job_profile5 = create(:job_profile, skills: [skill1, skill3])
+      job_profile6 = create(:job_profile, name: 'Boat builder', skills: [skill3])
```
tavy88: Could we not have just 3 profiles here? 

2 with the same score and 1 with a lower score?

Does it feel like you might be generating more data than needed?

app/controllers/skills_matcher_controller.rb
smoothcontract: I don't think this is paginating over a model though - it's an array of things?

spec/models/skills_matcher_spec.rb
```diff
@@ -0,0 +1,262 @@
...
+          job_profile_skills: {
+            job_profile1.id.to_s => [skill.id]
+          }
+        }
+      )
+
+      expect(matcher.match).to contain_exactly(job_profile2)
+    end
+
+    it 'returns multiple jobs matching skill selected' do
```
smoothcontract: This is essentially repeating the coverage from the preceding test, so I'd suggest removing the previous one and updating the description here to `returns multiple jobs matching skill selected ignoring job profiles in the session`

spec/models/skills_matcher_spec.rb
```diff
@@ -0,0 +1,262 @@
...
+          job_profile_skills: {
+            job_profile1.id.to_s => [skill1.id, skill3.id]
+          }
+        }
+      )
+
+      expect(matcher.match).to contain_exactly(job_profile2, job_profile3)
+    end
+
+    it 'returns multiple jobs matching multiple job skills selected' do
```
smoothcontract: I think this test makes the previous one superfluous, so maybe remove `returns multiple jobs matching different skills selected`

spec/models/skills_matcher_spec.rb
smoothcontract: You can probably get away with 4 job profiles here - as suggested above plus the job profile in the session.

app/models/skills_matcher.rb
cwrw: We are counting an array of no more than maybe 60 integers, not sure this is relevant here

spec/models/skills_matcher_spec.rb
cwrw: I would like to keep both tests as they are different in the sense that one checks that all skills *within* a job profile are included in query, and one that checks that all skills *across* job profiles are queried

smoothcontract: Okay 👍

PR #137, opened 2019-08-14
Fix broken UX journey caused page redesign

tavy88:
The Skills page has been redesigned and the search form
was removed from the results page causing an issue when
the searched skill has no results, as the user has no way to
search again.

## Screenshots

Screen Shot 2019-08-14 at 15 07 23

https://dfedigital.atlassian.net/browse/GET-365

### DIFF
--- a/app/views/check_your_skills/results.html.erb
+++ b/app/views/check_your_skills/results.html.erb
@@ -13,14 +13,16 @@
 <div class="govuk-grid-row govuk-!-padding-top-7">
   <div class="govuk-grid-column-two-thirds" >
     <% if @job_profiles.blank? %>
-      <p class="govuk-body-m"><%= render '/shared/search/no_results' %></p>
+      <%= render '/shared/search/no_results' %>
     <% else %>
+      <h3 class="govuk-heading-xl govuk-!-margin-bottom-7"><%= t('.title') %></h3>
+      <%= render "shared/search/results_form", search_path: results_check_your_skills_path, scope: 'check_your_skills.results' %>
       <p class="govuk-body-l govuk-!-margin-bottom-2"><%= t('.body') %></p>
       <%= render "results_list", job_profiles: @job_profiles %>
+      <div class="govuk-grid-column-full govuk-!-padding-top-6 govuk-!-margin-bottom-4 govuk-!-margin-top-6 pagination__section">
+        <%= paginate @job_profiles %>
+      </div>
     <% end %>
-    <div class="govuk-grid-column-full govuk-!-padding-top-6 govuk-!-margin-bottom-4 govuk-!-margin-top-6">
-      <%= paginate @job_profiles %>
-    </div>
   </div>
   <%= render "shared/contact_us" %>
 </div>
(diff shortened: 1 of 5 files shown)


spec/features/check_your_skills_spec.rb
```diff
@@ -33,7 +33,17 @@
...
+  end
+
+  scenario 'User can return to the search form when no results are found' do
+    visit(check_your_skills_path)
+    fill_in('search', with: 'Embalmer')
+    find('.search-button').click
+
+    click_on('Search again')
+
+    expect(current_path).to eql(check_your_skills_path)
```
smoothcontract: Rubocop is suggesting a change to use `have_current_path` matcher here.

app/views/check_your_skills/results.html.erb
```diff
@@ -12,16 +12,14 @@
...
-    <h3 class="govuk-heading-l"><%= t('.title') %></h3>
-    <%= render "shared/search/results_form", search_path: results_check_your_skills_path, scope: 'check_your_skills.results' %>
     <% if @job_profiles.blank? %>
       <p class="govuk-body-m"><%= render '/shared/search/no_results' %></p>
     <% else %>
       <p class="govuk-body-m" ><%= t('.results', count: @job_profiles.total_count) %></p>
       <p class="govuk-inset-text govuk-!-margin-bottom-2"><%= t('.body') %></p>
       <%= render "results_list", job_profiles: @job_profiles %>
     <% end %>
-    <div class="govuk-grid-column-full govuk-!-padding-top-6 govuk-!-margin-bottom-4 govuk-!-margin-top-6 pagination__section">
```
cwrw: discussed in person to add class back

app/views/check_your_skills/results.html.erb
```diff
@@ -12,7 +12,6 @@
 
 <div class="govuk-grid-row govuk-!-padding-top-7">
   <div class="govuk-grid-column-two-thirds" >
-    <h3 class="govuk-heading-xl"><%= t('.title') %></h3>
```
cwrw: is this intentional as well?

tavy88: Ok, it was a bad merge conflict fix as the condition I added is missing.

app/views/check_your_skills/results.html.erb
tavy88: Ok, it was a bad merge conflict fix here as well. This should have moved actually inside the else.

tavy88:
@cwrw The things you reported happened due to a previously bad merge. I have updated the code now.

app/views/check_your_skills/results.html.erb
```diff
@@ -12,16 +12,17 @@
...
     <% if @job_profiles.blank? %>
       <p class="govuk-body-m"><%= render '/shared/search/no_results' %></p>
```
cwrw: can you remove `<p class="govuk-body-m"></p>` and keep only `<%= render '/shared/search/no_results' %>` as this should be wrapped in it, also it's pushing the section down

PR #138, opened 2019-08-15
Add rake task to correctly set recommended flag for all job profiles

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-226

Other than 54 job profiles that have been identified, all other job profiles are recommended. The 54 identified are excluded as they require a degree level qualification. Excluded jobs will still be included in the search results for "Check your skills" and skills builder so no code changes have been made there. Explore careers is about to be retired in favour of skill matcher so no changes have been made there either.

Excluded jobs don't appear in the skill matcher results, but remain elsewhere.

### Guidance to review
Run the rake task locally as described in the README - this will set `recommended` to true for all job profiles except the ones listed in the spreadsheet linked to this Jira task: https://dfedigital.atlassian.net/browse/GET-304

This can be checked by taking a look at http://localhost:3000/admin/job_profiles

### DIFF
--- a/lib/tasks/data_import/update_recommended_job_profiles.rake
+++ b/lib/tasks/data_import/update_recommended_job_profiles.rake
@@ -1,0 +1,74 @@
+# rubocop:disable Metrics/BlockLength, Metrics/LineLength
+namespace :data_import do
+  # bin/rails data_import:update_recommended_job_profiles
+  desc 'Sets correct recommended value for all job profiles'
+  task update_recommended_job_profiles: :environment do
+    print 'Updating recommended flag for all job profiles'
+    if JobProfile.any?
+      non_recommended_jobs = JobProfile.where(slug: %w[
+                                                astronaut
+                                                naturopath
+                                                psychiatrist
+                                                vet
+                                                art-therapist
+                                                sport-and-exercise-psychologist
+                                                geoscientist
+                                                naval-architect
+                                                botanist
+                                                dentist
+                                                data-analyst-statistician
+                                                audiologist
+                                                archaeologist
+                                                dance-movement-psychotherapist
+                                                ornithologist
+                                                forensic-psychologist
+                                                psychologist
+                                                play-therapist
+                                                consumer-scientist
+                                                astronomer
+                                                nutritionist
+                                                chiropractor
+                                                research-and-development-manager
+                                                cognitive-behavioural-therapist
+                                                environmental-consultant
+                                                medical-herbalist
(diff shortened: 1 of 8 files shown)


lib/tasks/data_import/update_recommended_job_profiles.rake
```diff
@@ -0,0 +1,74 @@
...
+                                                oceanographer
+                                                sports-scientist
+                                                pharmacist
+                                                speech-and-language-therapist
+                                                climate-scientist
+                                                archivist
+                                                ergonomist
+                                                clinical-psychologist
+                                                hospital-doctor
+                                                oil-and-gas-operations-manager
```
tavy88: Wonder if this list will change we will need another deploy?

tavy88: You clarified this in the standup already. So ignore my comment.

tavy88:
Just some failures on the build.

spec/tasks/update_recommended_job_profiles_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+require 'support/tasks'
+
+RSpec.describe 'data_import:update_recommended_job_profiles' do
+  let!(:astronaut) { create :job_profile, slug: 'astronaut' }
+  let!(:mad_titan) { create :job_profile, slug: 'mad-titan' }
```
cwrw: 🥊

spec/tasks/update_recommended_job_profiles_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  it 'excludes specific named jobs' do
+    task.execute
+    expect(astronaut.reload).to be_excluded
```
cwrw: ✂️  `expect(astronaut.reload).not_to be_recommended `

smoothcontract:
> so we are ok that all new jobs added will be flagged as false as default?

Yep, that's the intention. Any new jobs that we scrape will need to have growth data added plus potentially they might be degree roles. So the safest thing is to leave the default for these as `recommended = false` then they won't appear in the skills matcher.

spec/tasks/update_recommended_job_profiles_spec.rb
smoothcontract: That's why the build fails.. I added a scope for this but it's not present as an instance method 🤦‍♂

PR #144, opened 2019-08-16
Foundation for user completion journey

tavy88:
### Context

We want to have user unlock the sections on our platform
once they've went to a certain flow as follows:

'CAN’T START YET' status shown on tasklist page
on all tasks except section 1.

‘CAN’T START YET’ removed from section two once user
has clicked on ‘Find out what you can do with
these skills’ CTA on selected skills page.

‘CAN’T START YET’ removed from sections 3,4,5  after user has
clicked on ‘Find a training course’ or ‘Other ways to change jobs’
CTAs on the job descriptions page.

The state of user progress is stored on the session.

###  Screenshots

Screen Shot 2019-08-16 at 15 24 57
Screen Shot 2019-08-16 at 15 25 34
Screen Shot 2019-08-16 at 15 25 49
Screen Shot 2019-08-16 at 15 35 38

### JIRA

https://dfedigital.atlassian.net/browse/GET-291

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,5 +1,5 @@
 class PagesController < ApplicationController
-  # TO DO: This needs to be removed as it's never being hit
+  # TODO: This needs to be removed as it's never being hit
   def show
     render template: "pages/#{params[:page]}"
   end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,6 +1,15 @@
 class PagesController < ApplicationController
+  # TO DO: This needs to be removed as it's never being hit
   def show
     render template: "pages/#{params[:page]}"
+  end
+
+  def training_hub
+    user_session.track_page('training_hub')
+  end
+
+  def next_steps
+    user_session.track_page('next_steps')
   end
 
   def location_eligibility
--- a/app/controllers/skills_matcher_controller.rb
+++ b/app/controllers/skills_matcher_controller.rb
@@ -1,10 +1,12 @@
 class SkillsMatcherController < ApplicationController
   def index
+    return redirect_to task_list_path unless user_session.job_profile_skills?
+
+    user_session.track_page('skills_matcher_index')
+
     @results = Kaminari.paginate_array(skills_matcher.match).page(params[:page])
     @scores = skills_matcher.job_profile_scores
-    @job_profiles = @results.map { |job_profile|
-      JobProfileDecorator.new(job_profile)
-    }
+    @job_profiles = @results.map { |job_profile| JobProfileDecorator.new(job_profile) }
   end
 
   private
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -1,0 +1,20 @@
+class UserSession
+  attr_reader :session
+
+  def initialize(session)
+    @session = session
+    @session[:visited_pages] ||= []
+  end
+
+  def track_page(page_key)
+    session[:visited_pages] << page_key unless session[:visited_pages].include?(page_key)
+  end
+
+  def page_visited?(page_key)
+    session[:visited_pages].include?(page_key)
+  end
+
+  def job_profile_skills?
+    session.fetch(:job_profile_skills, {}).keys.present?
+  end
+end
--- a/app/views/pages/task_list.html.erb
+++ b/app/views/pages/task_list.html.erb
@@ -17,77 +17,19 @@
     </p>
     <ol class="app-task-list">
       <li>
-        <h2 class="app-task-list__section">
-          <span class="app-task-list__section-number">1. </span> Check what you are good at
-        </h2>
-        <ul class="app-task-list__items">
-          <li class="app-task-list__item">
-            <div class="app-task-list__task-name">
-              <%= link_to check_your_skills_path, class: 'govuk-link' do %>
-                Check your existing skills <span class="app-step-nav__context">5 minutes</span>
-              <% end %>
-              <p class="govuk-body govuk-!-margin-bottom-0">You'll be surprised at what you already know.</p>
-            </div>
-          </li>
-        </ul>
+        <%= render 'pages/task_list_items/item_1' %>
       </li>
       <li>
-        <h2 class="app-task-list__section">
-          <span class="app-task-list__section-number">2. </span> Explore the types of jobs you could do
-        </h2>
-        <ul class="app-task-list__items">
-          <li class="app-task-list__item">
-            <div class="app-task-list__task-name">
-              <%= link_to Flipflop.skills_builder? ? skills_matcher_index_path : explore_occupations_path, class: 'govuk-link' do %>
-                Search for the types of jobs you could retrain to do <span class="app-step-nav__context">5 to 20 minutes</span>
-              <% end %>
-              <p class="govuk-body govuk-!-margin-bottom-0">These are not live jobs you can apply to – they're the types of jobs you'll be able to do with your existing skills or once you do some training.</p>
-            </div>
-          </li>
-        </ul>
+        <%= render 'pages/task_list_items/item_2' %>
       </li>
       <li>
-        <h2 class="app-task-list__section">
-          <span class="app-task-list__section-number">3. </span> Find training that's right for you
-        </h2>
-        <ul class="app-task-list__items">
-          <li class="app-task-list__item">
-            <div class="app-task-list__task-name">
-              <%= link_to Flipflop.course_directory? ? training_hub_path : find_training_courses_path, class: 'govuk-link' do %>
-                Find a training course <span class="app-step-nav__context">10 to 20 minutes</span>
-              <% end %>
-              <p class="govuk-body govuk-!-margin-bottom-0">Training can help you access a broader range of jobs and stay in secure work.</p>
-            </div>
-          </li>
-        </ul>
+        <%= render 'pages/task_list_items/item_3' %>
       </li>
       <li>
-        <h2 class="app-task-list__section">
-          <span class="app-task-list__section-number">4. </span> Find out what you can do next
-        </h2>
-        <ul class="app-task-list__items">
-          <li class="app-task-list__item">
-            <div class="app-task-list__task-name">
-              <%= link_to 'Get more support to help you on your new career path', next_steps_path, class: 'govuk-link' %>
-              <p class="govuk-body govuk-!-margin-bottom-0">For example, advice on how to find and apply for jobs.</p>
-            </div>
-          </li>
-        </ul>
+        <%= render 'pages/task_list_items/item_4' %>
       </li>
       <li>
-        <h2 class="app-task-list__section">
-          <span class="app-task-list__section-number">5. </span> Help improve this service
-        </h2>
-        <ul class="app-task-list__items">
-          <li class="app-task-list__item">
-            <div class="app-task-list__task-name">
-              <%= link_to 'https://www.smartsurvey.co.uk/s/get-help-to-retrain/', class: 'govuk-link', data: { tracked_event: 'TaskList - User Exit Survey Link' }, target: '_blank' do %>
-                Take a quick survey and tell us about your experience <span class="app-step-nav__context">3 to 5 minutes</span>
-              <% end %>
-              <p class="govuk-body govuk-!-margin-bottom-0">This is a new service – your feedback will help us improve it.</p>
-            </div>
-          </li>
-        </ul>
+        <%= render 'pages/task_list_items/item_5' %>
       </li>
     </ol>
   </div>
--- a/app/views/pages/task_list_items/_body.html.erb
+++ b/app/views/pages/task_list_items/_body.html.erb
@@ -1,0 +1,18 @@
+<% if enabled %>
+  <%= link_to path, class: 'govuk-link' do %>
+    <%= cta %>
+    <% if local_assigns[:time_to_complete].present? %>
+      <span class="app-step-nav__context"><%= time_to_complete %></span>
+    <% end %>
+  <% end %>
+<% else %>
+  <div class="govuk-grid-row govuk-body">
+    <div class="govuk-grid-column-full">
+      <span id='section-<%= span_id %>-blocked' class="govuk-tag app-task-list__task-not-active">Can’t start yet</span>
+      <%= cta %>
+      <% if local_assigns[:time_to_complete].present? %>
+        <span class="app-step-nav__context"><%= time_to_complete %></span>
+      <% end %>
+    </div>
+  </div>
+<% end %>
--- a/app/views/pages/task_list_items/_item_2.html.erb
+++ b/app/views/pages/task_list_items/_item_2.html.erb
@@ -1,0 +1,11 @@
+<h2 class="app-task-list__section">
+  <span class="app-task-list__section-number">2. </span> <%= t('pages.job_matches.title') %>
+</h2>
+<ul class="app-task-list__items">
+  <li class="app-task-list__item">
+    <div class="app-task-list__task-name">
+      <%= render 'pages/task_list_items/body', enabled: user_session.page_visited?('skills_matcher_index'), path: skills_matcher_index_path, cta: t('pages.job_matches.cta'), time_to_complete: t('pages.job_matches.time'), span_id: '2' %>
+      <p class="govuk-body govuk-!-margin-bottom-0"><%= t('pages.job_matches.description') %></p>
+    </div>
+  </li>
+</ul>
--- a/spec/features/skills_matcher_spec.rb
+++ b/spec/features/skills_matcher_spec.rb
@@ -36,6 +36,20 @@ RSpec.feature 'Skills matcher', type: :feature do
     expect(page).to have_text('Chameleon-like blend in tactics')
   end
 
+  scenario 'Page gets tracked on the session' do
+    visit_skills_for_current_job_profile
+
+    current_session = Capybara.current_session.driver.request.session
+
+    expect(current_session[:visited_pages]).to eq(['skills_matcher_index'])
+  end
+
+  scenario 'Redirect to tasks-list page when session is missing the job_profile_skills key' do
+    visit '/job-matches'
+
+    expect(page).to have_current_path(task_list_path)
+  end
+
   scenario 'returns other profiles than one selected for skills' do
     visit_skills_for_current_job_profile
 
--- a/spec/features/task_list_spec.rb
+++ b/spec/features/task_list_spec.rb
@@ -9,19 +9,29 @@ RSpec.feature 'Tasks List', type: :feature do
     expect(page).to have_text('Here\'s what you need to do to retrain for another job.')
   end
 
+  scenario 'With a clean new session the user will see sections 2,3,4,5 locked' do
+    disable_feature! :location_eligibility
+    visit(task_list_path)
+
+    (2..5).each do |section_no|
+      expect(page).to have_css("span#section-#{section_no}-blocked")
+    end
+  end
+
+  scenario 'With a clean new session the user will not be able to click sections 2,3,4,5' do
+    disable_feature! :location_eligibility
+    visit(task_list_path)
+
+    ['See types of jobs that match your skills', 'Find a training course', 'Find other ways to change jobs', 'Take a quick survey'].each do |link|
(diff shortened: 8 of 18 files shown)


app/sessions/user_session.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def page_visited?(page_key)
+    session[:visited_pages].include?(page_key)
+  end
+
+  def job_profile_skills?
+    session.fetch(:job_profile_skills, {}).keys.present?
+  end
+end
```
tavy88: This is tested indirectly. Writing unit tests is a bit difficult as the session is not accessible at this level. Not sure if there is any value in doing that here.

app/controllers/pages_controller.rb
```diff
@@ -1,8 +1,15 @@
 class PagesController < ApplicationController
+  # TO DO: This needs to be removed as it's never being hit
   def show
     render template: "pages/#{params[:page]}"
   end
 
+  %w[training_hub next_steps].each do |attribute|
```
smoothcontract: I'm not sure this warrants meta programming TBH. As there are only two methods it would be less code and easier to follow with two similar methods. I'd prefer an implementation along these lines:

```ruby
def training_hub
  track_page_vist(:training_hub)
end

def next_steps
  track_page_visit(:next_steps)
end

private

def track_page_visit(page_key)
  user_session.track_page(page_key) unless user_session.page_visited?(page_key)
end
```

app/controllers/skills_matcher_controller.rb
```diff
@@ -1,11 +1,15 @@
 class SkillsMatcherController < ApplicationController
+  # rubocop:disable Metrics/AbcSize
   def index
+    return redirect_to task_list_path unless user_session.job_profile_skills?
+
+    user_session.track_page('skills_matcher_index') unless user_session.page_visited?('skills_matcher_index')
```
smoothcontract: Feels like this could be a method that gets moved to the `ApplicationController` - as the same line appears here and in the previous controller. That would drop the complexity a little too so might be sufficient to remove the rubocop override.

app/controllers/skills_matcher_controller.rb
```diff
@@ -1,11 +1,15 @@
 class SkillsMatcherController < ApplicationController
+  # rubocop:disable Metrics/AbcSize
   def index
+    return redirect_to task_list_path unless user_session.job_profile_skills?
```
smoothcontract: You could consider making this a before action if the suggestion below isn't sufficient to reduce the complexity enough for rubocop to be happy. I'd prefer that we didn't need to ignore the AbcSize metric.

app/sessions/user_session.rb
```diff
@@ -0,0 +1,20 @@
+class UserSession
```
smoothcontract: I like this class but not sure what else would end up in the `sessions` folder to accompany it? Maybe move this to models or lib instead?

app/sessions/user_session.rb
smoothcontract: You can simulate the session by injecting a HashWithIndifferentAccess in the constructor - that should be enough to allow testing this at unit level.

app/controllers/skills_matcher_controller.rb
smoothcontract: Alternatively, you could add a new method to `UserSession` that avoids having to call it with a guard clause e.g.

```ruby
def track_unique_page(page_key)
  track_page(page_key) unless page_visited?(page_key)
end
```

and then call that here as `user_session.track_unique_page(:skills_matcher_index)`

app/sessions/user_session.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize(session)
+    @session = session
+    @session[:visited_pages] ||= []
+  end
+
+  def track_page(page_key)
+    session[:visited_pages] << page_key
+  end
+
```
smoothcontract: Consider adding `track_unique_page` method here as mentioned previously (you can keep the other methods if they're useful .. or perhaps just adapt the `track_page` method to prevent duplicates being added. Is this checking even needed if it's just a hash?

app/views/pages/task_list_items/_item_2.html.erb
```diff
@@ -0,0 +1,23 @@
+<h2 class="app-task-list__section">
+  <span class="app-task-list__section-number">2. </span> Explore the types of jobs you could do
+</h2>
+<ul class="app-task-list__items">
+  <li class="app-task-list__item">
+    <div class="app-task-list__task-name">
+      <% if user_session.page_visited?('skills_matcher_index') %>
```
smoothcontract: As these have moved to partials it's a nice opportunity to move this logic out to the parent view and pass in an variable here e.g. `enabled` - that then lets you avoid duplicating the markup. Something more like this pattern:

```html
<%= link_to_if enabled, t(:see_types_of_jobs), skills_matcher_index_path, class: 'govuk-link' %>
<% unless enabled %>
  <span id='section-2-blocked' class="govuk-tag app-task-list__task-not-active">Can’t start yet</span>
<% end %>
<span class="app-step-nav__context">5 to 20 minutes </span>
```

Another option to consider would be moving the `blocked` markup to a helper to DRY it up across these 5 views.

app/views/pages/task_list.html.erb
```diff
@@ -17,77 +17,19 @@
...
-          <li class="app-task-list__item">
-            <div class="app-task-list__task-name">
-              <%= link_to Flipflop.skills_builder? ? skills_matcher_index_path : explore_occupations_path, class: 'govuk-link' do %>
-                Search for the types of jobs you could retrain to do <span class="app-step-nav__context">5 to 20 minutes</span>
-              <% end %>
-              <p class="govuk-body govuk-!-margin-bottom-0">These are not live jobs you can apply to – they're the types of jobs you'll be able to do with your existing skills or once you do some training.</p>
-            </div>
-          </li>
-        </ul>
+        <%= render 'pages/task_list_items/item_2' %>
```
smoothcontract: See comment below, consider passing `enabled` local here e.g.

```html
<%= render 'item_2', enabled: user_session.page_visited?(:skills_matcher_index) %>
```

(You don't need the fully qualified path to the partial either)

app/sessions/user_session.rb
tavy88: Yeah I did think about it but it's not exactly a hash and I think @cwrw also had some issues with some methods she was trying on one of her PRs that were hash methods but not  ActionDispatch::Request::Session. However, because the logic is simple I think the risk of hiding regressions is quite low.

app/controllers/pages_controller.rb
tavy88: I prefer the metaprogramming way here rather than declaring 3 methods. We can use just one generator.

app/controllers/skills_matcher_controller.rb
tavy88: True, but I wonder if it's a huge gain in doing this? I could do this change if you are very keen on it.

app/controllers/skills_matcher_controller.rb
tavy88: Fair

app/sessions/user_session.rb
tavy88: It's not an Active Record model nor a library. That's why I chose `sessions` namespace. What would be the benefit of moving it to lib?

app/sessions/user_session.rb
tavy88: Yeah I was thinking of doing

```ruby
session[:visited_pages] << page_key unless [page_key key present]
```

app/views/pages/task_list_items/_item_2.html.erb
tavy88: True but just aware of the time pressure atm and how far do we want to go in this iteration. I did consider making it prettier not gonna lie and I agree with you, but I am very aware of the time constraint.

app/sessions/user_session.rb
smoothcontract: We have some other stuff in `app/lib` that doesn't have a good home elsewhere. This is possibly closest to a model although it's not an ActiveRecord model; again we have other classes in there that are not mapped to the database. This would be orphaned all alone otherwise - unless we think there will be other similar things added over time.

app/controllers/skills_matcher_controller.rb
tavy88: Agreed.

app/sessions/user_session.rb
smoothcontract: You're right, session is not a hash but for the purposes of adding and retrieving keys a hash would suffice for testing. I suggest `HashWithIndifferentAccess` as it shouldn't matter if we call this with symbols or strings. It's just an idea for adding extra test coverage if you wish.

app/views/pages/task_list_items/_item_2.html.erb
smoothcontract: Yes, we don't have much time remaining but we do have until Monday lunchtime. I think this could be DRY'd up a little without spending too much more time; you've already done the bulk of the hard work.

app/controllers/pages_controller.rb
cwrw: I have to agree with @smoothcontract, I don't think we need meta-programming here. Maybe there is a middle ground with a `before_action` hook, something like this:
```
before_action -> { |value| user_session.track_page(value) },  only: [:training_hub, :next_step]
```

app/controllers/skills_matcher_controller.rb
```diff
@@ -1,15 +1,21 @@
...
   end
 
   private
 
   def skills_matcher
     @skills_matcher ||= SkillsMatcher.new(user_session: session)
   end
+
+  def handle_user_session
+    return redirect_to task_list_path unless user_session.job_profile_skills?
```
cwrw: I don't think this should be in a before action, because if we add more methods here this behaviour will be surprising. We should move this back to the method itself

app/sessions/user_session.rb
cwrw: I prefer this to be in the `models` folder, as its not a library, and represents a user model we will later move to

spec/features/next_steps_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.feature 'Next steps page', type: :feature do
+  scenario 'Page visit is persisted in the session' do
+    visit(next_steps_path)
```
cwrw: As these are integration tests, we should not have any knowledge of the session, but rather test these at the unit level instead

app/views/pages/task_list_items/_item_2.html.erb
```diff
@@ -0,0 +1,23 @@
...
+        <%= link_to skills_matcher_index_path, class: 'govuk-link' do %>
+          See types of jobs that match your skills
+        <% end %>
+        <span class="app-step-nav__context">5 to 20 minutes </span>
+      <% else %>
+        <div class="govuk-grid-row govuk-body">
+          <div class="govuk-grid-column-full">
+            <span id='section-2-blocked' class="govuk-tag app-task-list__task-not-active">Can’t start yet</span>
```
cwrw: are we using those ids anywhere?

app/views/pages/task_list_items/_item_2.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <span class="app-task-list__section-number">2. </span> Explore the types of jobs you could do
+</h2>
+<ul class="app-task-list__items">
+  <li class="app-task-list__item">
+    <div class="app-task-list__task-name">
+      <% if user_session.page_visited?('skills_matcher_index') %>
+        <%= link_to skills_matcher_index_path, class: 'govuk-link' do %>
+          See types of jobs that match your skills
+        <% end %>
+        <span class="app-step-nav__context">5 to 20 minutes </span>
```
cwrw: I think those should go back into the link body, it was good that you could click through from the time as well, here and everywhere

spec/features/skills_matcher_spec.rb
```diff
@@ -36,6 +36,20 @@ def visit_skills_for_current_job_profile
...
+
+    current_session = Capybara.current_session.driver.request.session
+
+    expect(current_session[:visited_pages]).to eq(['skills_matcher_index'])
+  end
+
+  scenario 'Redirect to tasks-list page when session is missing the job_profile_skills key' do
+    visit '/job-matches'
+
+    expect(page).to have_current_path('/task-list')
```
cwrw: let's use task_list_path instead here

app/controllers/pages_controller.rb
tavy88: I am still not convinced about what is the real issue against metaprogramming. I haven't been presented a strong argument on why my approach is wrong.

app/controllers/skills_matcher_controller.rb
tavy88: The reason why it was added there was to silence the rubocop method complexity violation.

app/views/pages/task_list_items/_item_2.html.erb
tavy88: It's not what the design says.

app/views/pages/task_list_items/_item_2.html.erb
tavy88: Yes in the testing I can easily and efficiently select the exact corresponding label.

spec/features/next_steps_spec.rb
tavy88: Sorry, what do you want to do here exactly?

spec/features/task_list_spec.rb
```diff
@@ -36,26 +46,109 @@
...
+      name: 'Assassin',
+      skills: [
+        create(:skill, name: 'Chameleon-like blend in tactics')
+      ]
+    )
+
+    visit(job_profile_skills_path(job_profile_id: job_profile.slug))
+    find('.govuk-button').click
+    click_on('Find out what you can do with these skills')
+    visit("job-profiles/#{job_profile.slug}")
```
cwrw: can we use path methods instead of the hardcoded paths themselves across this file, because if those change it will require us to manually clean those up

spec/features/next_steps_spec.rb
cwrw: Im proposing we remove session expectations in those tests, as these are integration tests and should not have knowledge of teh session, but rather test the user journey. We can test the session at a unit level, maybe in the request specs

app/controllers/skills_matcher_controller.rb
cwrw: sure adding an 'only'  is another way to do that, Im fine with that as well, as long as we're making sure this action is only tied to index 👍

app/controllers/pages_controller.rb
cwrw: > Meta-programming is extremely valuable, and can make solving complex problems easier. But remember, it’s only worth using when there’s a tradeoff — like readability and comprehension in exchange for solving complex problems with less code

I don't believe this is a complex problem to solve, but rather a simple one that only needs DRYing up, which can be achieved with being explicit and using methods we have access to easily, rather than exchange readability as well create methods in runtime. We should avoid using meta-programming unless we really have to, as its a known anti-pattern, and you do incur technical debt. It should only be used if it helps resolve our problems in a cleaner way when they are quite complex, which I don't think it is here unfortunately.

tavy88: If you need to add a new page in my case you change in only one place. If the case you suggest you need to create a new method. I feel in this case using metaprogramming scales better.

spec/features/next_steps_spec.rb
tavy88: I recommend you read this: https://thoughtbot.com/blog/how-we-test-rails-applications#feature-specs

app/controllers/pages_controller.rb
cwrw: in the before action you only change it in one place as well, unless I'm missing something?

app/controllers/skills_matcher_controller.rb
smoothcontract: @tavy88 the threshold for AbcSize is too low, so in this commit (PR not merged yet) I've bumped it up from 15 to 20. With the higher threshold is Rubocop happy with this reverting back to a line within `index` method?

app/views/pages/task_list_items/_item_2.html.erb
cwrw: I can see it in the design here? https://get-help-to-retrain-prototype.herokuapp.com/main/release2_e2e_v3/task_list

spec/features/next_steps_spec.rb
cwrw: Im not sure I see anything different to what Im proposing there, they are testing the session in the controller specs, which im suggesting we do as we use request specs instead of controller specs?

app/controllers/pages_controller.rb
smoothcontract: Meta programming is overkill for this scenario; we should aggressively pursue simple solutions to things. This specific example rings of cleverness for cleverness sake and is easier to read and understand in the long hand version. If we add another bunch of pages such that the same method needs calling for more than 2 pages (and 2 is probably the exact threshold beyond which this would be reviewed) then this can be refactored to support that. But let's not solve problems we don't have yet. I'd prefer to invest time addressing the other feedback in this PR.

app/controllers/skills_matcher_controller.rb
smoothcontract: @tavy88 the increase to `AbcSize` threshold is now in master branch if that helps.

app/views/pages/task_list_items/_body.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  <div class="govuk-grid-row govuk-body">
+    <div class="govuk-grid-column-full">
+      <span id='section-<%= span_id %>-blocked' class="govuk-tag app-task-list__task-not-active">Can’t start yet</span>
+      <%= cta %>
+      <% if local_assigns[:time_to_complete].present? %>
+        <span class="app-step-nav__context"><%= time_to_complete %></span>
+      <% end %>
+    </div>
+  </div>
+<% end %>
```
cwrw: 🦂

app/controllers/pages_controller.rb
```diff
@@ -1,8 +1,17 @@
 class PagesController < ApplicationController
+  # TO DO: This needs to be removed as it's never being hit
```
smoothcontract: I think rubocop will flag this - should be `TODO:`

app/controllers/pages_controller.rb
```diff
@@ -1,8 +1,17 @@
...
     render template: "pages/#{params[:page]}"
   end
 
+  def training_hub
+    user_session.track_page('training_hub')
+  end
+
+  def next_steps
+    user_session.track_page('next_steps')
+  end
```
smoothcontract: Thankyou 👍

app/models/user_session.rb
```diff
@@ -0,0 +1,20 @@
+class UserSession
+  attr_reader :session
+
+  def initialize(session)
+    @session = session
+    @session[:visited_pages] ||= []
+  end
+
+  def track_page(page_key)
+    session[:visited_pages] << page_key unless session[:visited_pages].include?(page_key)
```
smoothcontract: 👍

spec/features/skills_matcher_spec.rb
```diff
@@ -36,6 +36,20 @@ def visit_skills_for_current_job_profile
...
+    current_session = Capybara.current_session.driver.request.session
+
+    expect(current_session[:visited_pages]).to eq(['skills_matcher_index'])
+  end
+
+  scenario 'Redirect to tasks-list page when session is missing the job_profile_skills key' do
+    visit '/job-matches'
+
+    expect(page).to have_current_path(task_list_path)
+  end
```
tavy88: We can't manipulate the session inside the request spec so we are leaving this here as it does test the feature.

app/views/pages/task_list.html.erb
smoothcontract: I think this got covered off in the other changes within the partial. Could you perhaps try a shorter path to the partials though - I think this would work:

```ruby
<%= render 'task_list_items/item_2' %>
```

smoothcontract: Discussed offline, removing the `pages/` prefix breaks the page.

PR #147, opened 2019-08-20
GET-291 Fix unlocking bug and some refactoring

cwrw:
* Add helper method to show complete/incomplete sections
* Helper replaces body partial
* Move blocked section out of divs into p tag
* Move translations back into partials
* Add helper method to user_session to check all pages for unlocking
sections 3,4,5 rather than just next steps or training hub
* Reorder feature tests to reflect journey

deployed to https://dev2.nrs-ghtr.org.uk/task-list for testing


spec/features/task_list_spec.rb
```diff
@@ -9,6 +9,43 @@
     expect(page).to have_text('Here\'s what you need to do to retrain for another job.')
```
cwrw: the diff in this class looks weird, it's just rearranging tests and adding 2 new ones, nothing has been removed

cwrw: easier to review as a file rather than the diff

app/helpers/task_list_helper.rb
```diff
@@ -0,0 +1,23 @@
+module TaskListHelper
+  def link_or_blocked(path, enabled, span_id, **options, &block)
```
smoothcontract: Do you think it would be clearer to use named parameters for this method? There are quite a few here so need to refer back to the method definition to understand how it’s being called each time it appears in the view. If the params were named this would be clearer. Thoughts?

config/locales/en.yml
```diff
@@ -97,18 +97,8 @@ en-GB:
       title: Get help to retrain
     find_training_courses:
       title: Find training that's right for you
-      cta: Find a training course
```
smoothcontract: Why were these locale entries replaced with hardcoded content in the view? Just trying to understand rationale...

cwrw: talked in person, keeping as is

spec/features/task_list_spec.rb
```diff
@@ -220,77 +294,23 @@
...
     visit(task_list_path)
 
-    (3..5).each do |section_no|
+    (2..5).each do |section_no|
       expect(page).to have_css("span#section-#{section_no}-blocked")
     end
   end
 
-  scenario 'Unlocking skill matcher section does not allow clicking sections 3,4,5' do
+  scenario 'User does not unlock 3,4,5 by deep linking into next steps' do
```
tavy88: Not sure if the meaning here is the same as the old one?

cwrw: this test for 'Unlocking skill matcher section does not allow clicking sections 3,4,5' is still there, just moved up, the diff makes the changes hard to read. I would suggest opening the file on it's own to see the changes.

app/helpers/task_list_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+  private
+
+  def blocked_section(span_id, &_block)
+    content_tag(
+      :span,
+      "Can't start yet".html_safe,
+      class: 'govuk-tag app-task-list__task-not-active',
+      id: "section-#{span_id}-blocked"
+    ) + content_tag(:p, class: 'govuk-!-margin-bottom-4 govuk-!-margin-top-0') { yield }
+  end
```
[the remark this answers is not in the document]
cwrw: so far we aren't re-using this in any other context, so the block should be mandatory as this will change the whole method otherwise and we will need to guard against the whole `content_tag(:p)`. Until there is a use case for this I think it's fair to keep as is

app/models/user_session.rb
```diff
@@ -17,4 +17,9 @@ def page_visited?(page_key)
   def job_profile_skills?
     session.fetch(:job_profile_skills, {}).keys.present?
   end
+
+  def unblock_all_sections?
+    page_visited?('skills_matcher_index') &&
+      (page_visited?('training_hub') || page_visited?('next_steps'))
+  end
```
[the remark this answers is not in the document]
cwrw: Im pretty sure this is the intended behaviour, just an oversight from us in implementation. In terms of pages visited, I think I'd rather keep as is as this will obscure the `&&` or `||` logic and make it hard to follow if we group pages together. Here it's quite explicit which ones we care about

spec/features/task_list_spec.rb
tavy88: Cool.

PR #150, opened 2019-08-21
Skills Builder V2

tavy88:
### Context

Skills Builder v2

We are extending the capability of v1 meaning that we now allow
users to set their skills for multiple profiles.

Changes:

1. Code refactor to ensure backwards compatibility between
v1 and v2
2. Move most of the methods handling the session to the
UserSession class
3. v2 is feature flagged

### Screenshots

Screen Shot 2019-08-23 at 10 09 46

### Jira

https://dfedigital.atlassian.net/browse/GET-305


app/controllers/job_profiles_skills_controller.rb
```diff
@@ -31,6 +30,20 @@ def skills_valid?
   end
 
   def skills_params
-    params.permit(:job_profile_id, :search, skill_ids: [])
+    params.permit(:job_profile_id, :search, :editable, skill_ids: [])
+  end
+
+  def profile_skills_not_editable?
+    skills_params[:editable] != '1'
```
tavy88: This is needed to force the state of the skill page to become editable when we have max 5 profiles. Because we use different states of the same page, without this, when trying to edit the skills of a given job profile (out of the 5 ones) the user will keep being redirected to:

Screen Shot 2019-08-23 at 10 09 46

As they are not supposed to be able to access:

Screen Shot 2019-08-23 at 19 47 03

Since they have reached the max number of profiles.

app/controllers/job_profiles_skills_controller.rb
```diff
@@ -31,6 +30,20 @@ def skills_valid?
...
+  end
+
+  def job_profiles_cap_reached?
+    user_session.job_profile_ids.size > 4
```
cwrw: isn't the cap more than or equal to 5?

app/views/check_your_skills/index.html.erb
```diff
@@ -11,7 +11,11 @@
 
 <div class="govuk-grid-row govuk-!-margin-top-7">
   <div class="govuk-grid-column-two-thirds" >
-    <%= render "shared/search/form", search_path: check_your_skills_path, scope: 'check_your_skills.index' %>
+    <% if user_session.job_profile_ids.size > 0 %>
```
cwrw: If we are still on v1, will we see the new page as well?

app/controllers/job_profiles_skills_controller.rb
cwrw: instead of passing around an editable param which appears in the url and adds complexity, we can use this:
```
user_session.job_profile_ids.exclude?(job_profile.id)
```
This has to be part of the job profiles cap reached as well as we only want to check this *if* we reach the cap of 5.

spec/models/skills_matcher_spec.rb
```diff
@@ -16,13 +16,13 @@
       job_profile1 = create(:job_profile, skills: [skill])
       job_profile2 = create(:job_profile, skills: [skill])
       job_profile3 = create(:job_profile, skills: [skill])
-      matcher = described_class.new(
-        user_session: {
```
cwrw: instead of adding 2 new variables to each test, it's easier to just replace this line with `UserSession.new(` across the tests in this file

app/controllers/skills_controller.rb
```diff
@@ -1,36 +1,64 @@
...
+    @skills = Skill.find(skill_ids_for_job_profile)
+
+    track_event(:skills_index_selected, job_profile.slug => skill_ids_for_job_profile)
+
+    render 'skills/v2/index'
+  end
+
+  # Skills Builder v2 methods
+  def job_profiles
+    @job_profiles ||= JobProfile.where(id: user_session.job_profile_ids)
```
cwrw: can we move the `includes(:skills)` here?

app/controllers/skills_controller.rb
```diff
@@ -1,36 +1,64 @@
...
+  end
+
+  def show_skills_builder_v2
+    return redirect_to task_list_path unless job_profiles.any?
+
+    @job_profiles_with_skills ||= job_profiles.includes(:skills).map { |job_profile|
+      {
+        profile_id: job_profile.id,
+        profile_slug: job_profile.slug,
+        hero_copy: JobProfileDecorator.new(job_profile).hero_copy,
```
cwrw: do we need the decorator, we have access to the name from the job_profile? as in `job_profile.name`

app/controllers/skills_controller.rb
```diff
@@ -1,36 +1,64 @@
...
+        skills: job_profile.skills.each_with_object({}) { |skill, hash| hash[skill.id] = skill.name }
+      }
```
cwrw: This method feels like it should live in a helper/decorator rather than here, what do you think?

app/controllers/job_profiles_skills_controller.rb
tavy88: Cap is 5.

app/controllers/check_your_skills_controller.rb
```diff
@@ -1,13 +1,13 @@
 class CheckYourSkillsController < ApplicationController
   def index
-    @job_profile_search = JobProfileSearch.new(term: search)
+    @job_profile_search = JobProfileSearch.new(term: search, user_session: user_session)
```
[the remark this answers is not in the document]
tavy88: Yeah we could do that.

app/controllers/job_profiles_skills_controller.rb
tavy88: Let me digest this and make sure it still works.

app/controllers/job_profiles_skills_controller.rb
tavy88: 5 > 4 so we're good.

app/controllers/skills_controller.rb
tavy88: Why not.

app/controllers/skills_controller.rb
tavy88: Oh, I wasn't aware of that. That would be great!

app/controllers/skills_controller.rb
tavy88: Let's think about it as discussed.

app/views/check_your_skills/index.html.erb
tavy88: Maybe we need to use the feature flag here yeah.

app/controllers/job_profiles_skills_controller.rb
cwrw: So I am aware of the 5 limit, and this does pass that specification fine. 
But I'm thinking more of readability, especially for another developer first seeing this, I feel that >=5 is more explicit and highlights our specs that a profile has to be 5 or more for the cap, rather than more than 4. Not a blocker though

app/controllers/skills_controller.rb
```diff
@@ -1,36 +1,64 @@
...
+
+  def show_skills_builder_v2
+    return redirect_to task_list_path unless job_profiles.any?
+
+    @job_profiles_with_skills ||= job_profiles.includes(:skills).map { |job_profile|
+      {
+        profile_id: job_profile.id,
+        profile_slug: job_profile.slug,
+        hero_copy: JobProfileDecorator.new(job_profile).hero_copy,
+        skills: job_profile.skills.each_with_object({}) { |skill, hash| hash[skill.id] = skill.name }
```
cwrw: maybe we can move the logic for limiting skills here, example: 
`job_profile.skills.each_with_object({}) { |skill, hash| hash[skill.id] = skill.name }.slice(user_session.skill_ids_for_profile(job_profile.id))`
or something similar

app/models/user_session.rb
```diff
@@ -10,6 +10,10 @@ def track_page(page_key)
     session[:visited_pages] << page_key unless session[:visited_pages].include?(page_key)
   end
 
+  def store_at(key:, value:)
+    session[key.to_sym] = value
```
cwrw: why don't we just pass a symbol here instead?

app/models/user_session.rb
```diff
@@ -18,6 +22,30 @@ def job_profile_skills?
...
+  def skill_ids_for_profile(id)
+    session.fetch(:job_profile_skills, {})[id.to_s] || []
+  end
+
+  def current_job?
+    session[:current_job_id].present?
+  end
+
+  def current_job_id
+    @current_job_id ||= session[:current_job_id]
```
cwrw: There not a risk that we will always have the old `current_job_id` even if we update it as this is memoized? Also we aren't reusing this anywhere

app/models/user_session.rb
```diff
@@ -18,6 +22,30 @@ def job_profile_skills?
     session.fetch(:job_profile_skills, {}).keys.present?
   end
 
+  def skill_ids
+    return skill_ids_for_profile(current_job_id) if Flipflop.skills_builder? && current_job?
+
+    session.fetch(:job_profile_skills, {}).values.flatten.uniq
```
cwrw: instead of doing `session.fetch(:job_profile_skills, {})` in each method can we not have a method that does this: 
```
def job_profile_skills
  session.fetch(:job_profile_skills, {})
end
```

cwrw:
Also another thing, if I navigate to edit these skills, and then back through breadcrumbs to search, I lose the search, we should maintain the search params in this journey as well

app/controllers/job_profiles_skills_controller.rb
```diff
@@ -1,10 +1,9 @@
 class JobProfilesSkillsController < ApplicationController
   def index
     skills_builder.build
-    if skills_valid?
-      redirect_to(
-        skills_path(job_profile_id: skills_params[:job_profile_id], search: skills_params[:search])
-      )
+
+    if (skills_valid? || job_profiles_cap_reached?) && profile_skills_not_editable?
```
cwrw: check validation after all 5 selected

app/models/job_profile_search.rb
```diff
@@ -1,22 +1,26 @@
 class JobProfileSearch
   include ActiveModel::Validations
 
-  attr_reader :term
+  attr_reader :term, :user_session
```
cwrw: talked in person, investigate what to do with user with old sessions and more than 5 jobs in the session

app/models/user_session.rb
tavy88: Simply because I prefer to pass the key as a value when I call `store_at`.

To me:

```ruby
store_at(key: 'some key', value: 'some value')
```
reads better than

```ruby
store_at(key: :some_key, value: 'some value')
```

cwrw: This method is basically a setter for user session keys. If I pass a key as a string, I expect the key to be a string, and would be surprised that it had been turned into a symbol. We have a mix of keys as strings/symbols in the user session which we will be setting later using this method, and if someone uses this method will be surprised that it doesn't match the expected behaviour. Instead we can allow the user to define what the key should be and pass it in as is.

app/models/job_profile_search.rb
```diff
@@ -1,22 +1,23 @@
 class JobProfileSearch
   include ActiveModel::Validations
 
-  attr_reader :term
+  attr_reader :term, :profile_ids_to_exclude
```
cwrw: can you add a test for this?

app/controllers/job_profiles_skills_controller.rb
```diff
@@ -1,11 +1,12 @@
...
-    if skills_valid?
-      redirect_to(
-        skills_path(job_profile_id: skills_params[:job_profile_id], search: skills_params[:search])
-      )
+    if profile_skills_not_editable? && job_profiles_cap_reached? && Flipflop.skills_builder_v2?
+      redirect_to_profile_skills_page
     else
+      skills_builder.build
+
+      redirect_to_profile_skills_page && return if skills_valid?
```
cwrw: this can just be `return redirect_to_profile_skills_page if skills_valid?`

app/controllers/job_profiles_skills_controller.rb
```diff
@@ -31,6 +32,20 @@ def skills_valid?
   end
 
   def skills_params
-    params.permit(:job_profile_id, :search, skill_ids: [])
+    params.permit(:job_profile_id, :search, :editable, skill_ids: [])
```
cwrw: we no longer need `:editable` here

app/models/job_profile_search.rb
tavy88: Tests are there.

PR #155, opened 2019-08-22
GET-326 Revisit accessibility in all forms

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-326

* Remove javascript disabling of form and required fields as now we will return an error message if user does not enter postcode in location eligibility as well as courses search, and job name in check your skills and occupations
* Add form validations for location eligibility, courses search, check your skills and explore occupations (not fully supported as deprecated soon)
* Move Job Search into its own class and out of Job Profile model


app/controllers/check_your_skills_controller.rb
```diff
@@ -1,7 +1,14 @@
...
+    @job_profile_search = JobProfileSearch.new(term: search)
+    redirect_to(results_check_your_skills_path(search: search)) if search && @job_profile_search.valid?
+  end
+
   def results
-    track_event :check_your_skills_index_search, search: search
-    @job_profiles = JobProfile.search(search).page(params[:page])
+    track_event(:check_your_skills_index_search, search: search) if search.present?
+
+    @job_profile_search = JobProfileSearch.new(term: search)
```
tavy88: Because this is used in both index & results wonder if it's worth adding it on a before_action?

app/controllers/explore_occupations_controller.rb
```diff
@@ -1,11 +1,16 @@
...
+    redirect_to(results_explore_occupations_path(search: search)) if search && @job_profile_search.valid?
   end
 
   def results
-    track_event :explore_occupations_index_search, search: search
-    @search_results = JobProfile.search(search).includes(:categories).page(params[:page])
+    track_event(:explore_occupations_index_search, search: search) if search.present?
+    @job_profile_search = JobProfileSearch.new(term: search)
+
+    @search_results = @job_profile_search.search.includes(:categories).page(params[:page])
```
tavy88: We don't use this anymore as explore occupations page goes away? You've done the effort now to add the logic but could you add a Deprecation comment for this controller?

cwrw: Unfortunately the tests failed and it was easier to amend this than to completely remove explore occupations which is to be done in another ticket. Comment is in the commit message about deprecation

app/controllers/check_your_skills_controller.rb
cwrw: `before_action` is used as a callback and not to define/DRY up any duplication in instance variables, also to define conditions *before* methods, for instance to check if something is nil etc, so I'd like to keep it as is.

tavy88: I know that - it was just bothering me seeing the same line of code in both methods. But it's my OCD ha ha.

app/controllers/explore_occupations_controller.rb
tavy88: I do feel it's better visibility to have that in the controller so we can actually see it.

PR #156, opened 2019-08-27
GET-407 Force locale and collation settings to english GB

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-407

Failures in tests are happening sporadically on sorting order, specifically alphabetically in CI. I could not replicate this locally with docker, but setting the encoding/collation explicitly as a test for now to see if this will mitigate the issue for the future, as we are not sure what is forcing the collation to change on specific vms.


cwrw:
This is not an issue, closing for now

PR #159, opened 2019-08-30
GET-382 clean up feature flags

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-382

* Remove location eligibility flag and make default feature
* Remove training hub flag and make default feature
* Remove explore occupations and category pages
* Remove old skills page in favour of skills builder and remove skills builder v1 flag

deployed to https://dev1.nrs-ghtr.org.uk/ for testing


app/views/job_profiles/skills/index.html.erb
```diff
@@ -19,7 +19,7 @@
           <fieldset class="govuk-fieldset" aria-describedby="current-job-skills-hint">
             <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
               <h1 class="govuk-fieldset__heading govuk-!-margin-bottom-3">
-                <%= @job_profile.hero_copy %>
+                <%= @job_profile.name %>
```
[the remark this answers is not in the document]
cwrw: there's no need to include the job profile decorator, as the hero_copy is basically the name, so just switching that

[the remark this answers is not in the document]
cwrw: There is no change really, just cleanup, which is within the scope

tavy88:
What about categories scrape task? And table? Do we still want to keep them?

spec/features/skills_builder_v2_spec.rb
```diff
@@ -161,6 +161,34 @@ def build_max_job_profiles
...
+    find('.govuk-button').click
+
+    visit(job_profile_skills_path(job_profile_id: job_profile1.slug))
+    find('.govuk-button').click
+
+    expect(page.all('div.govuk-grid-column-two-thirds h2.govuk-heading-m').collect(&:text)).to eq(
+      [
+        "#{job_profile2.name} edit these skills",
+        "#{job_profile1.name} edit these skills"
+      ]
```
tavy88: I don't think you need to check for `edit these skills` link here as that's already covered in another test and it's not really in the scope of this. Up to you.

tavy88: Thanks for addressing the fix here.

cwrw: it comes out as part of the text as it's in a span unfortunately

app/views/job_profiles/skills/index.html.erb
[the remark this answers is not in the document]
smoothcontract: Okay, that's a nice cleanup - but would be good to completely remove the `hero_copy` method from the decorator anywhere else it's still used and replace with `name`.

app/views/skills/v2/index.html.erb
```diff
@@ -16,24 +16,34 @@
...
-            <th class="govuk-table__header govuk-section-break__xl govuk-!-padding-0" scope="col">
-              <p class="govuk-body">You gained these skills from working as a <%= @job_profile.hero_copy %></p>
-            </th>
-          </tr>
-        </thead>
-        <tbody class="govuk-table__body">
-          <% @skills.each do |skill| %>
+      <% @job_profiles_with_skills.each do |job_profile_hash| %>
+        <table class="govuk-table">
+          <h2 class="govuk-heading-m govuk-!-margin-bottom-1 govuk-!-margin-top-8"><%= job_profile_hash[:hero_copy] %>
```
smoothcontract: I think this needs to change to `:name` too

app/views/skills/v2/index.html.erb
```diff
@@ -16,24 +16,34 @@
...
+            <span class='edit-skills govuk-!-margin-bottom-2'>
+              <%= link_to('edit these skills', job_profile_skills_path(job_profile_id: job_profile_hash[:profile_slug], search: params[:search]), class:'govuk-link govuk-body') %>
+            </span>
+          </h2>
+          <thead class="govuk-table__head">
             <tr class="govuk-table__row">
-              <td class="govuk-table__cell govuk-!-padding-bottom-4 govuk-!-padding-top-4"><%= skill.name %></th>
+              <th class="govuk-table__header govuk-section-break__xl govuk-!-padding-0" scope="col">
+                <p class="govuk-body">You gained these skills from working as a <%= job_profile_hash[:hero_copy] %></p>
```
smoothcontract: This needs to change to `:name` as well

spec/decorators/job_profile_decorator_spec.rb
```diff
@@ -151,42 +151,6 @@
...
-          </div>
-        </div>
-      </header>'
-    end
-
-    it 'extracts the hero copy' do
-      expect(job_profile.hero_copy).to eq 'Archivist'
-    end
-
-    it 'extracts the additional copy' do
```
smoothcontract: I think this test can be reinstated along with the corresponding setup - it's only the `hero_copy` method that was deleted.

app/views/skills/v2/index.html.erb
cwrw: this is being manually set here: https://github.com/DFE-Digital/get-help-to-retrain/blob/master/app/models/job_profile.rb#L40 could change this also I guess

app/views/skills/v2/index.html.erb
cwrw: see comment above

spec/decorators/job_profile_decorator_spec.rb
cwrw: ah I didn't see this as it is under the method describe, will reinstate

PR #164, opened 2019-08-30
Upgrade Rails 5.2.3 -> 6.0.0

tavy88:
### Context

Upgrade our app from Rails 5.2.3 to Rails 6.0.0


config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 6.0 upgrade.
+#
+# Once upgraded flip defaults one by one to migrate to the new default.
```
smoothcontract: Could you add a TODO here so we remember to complete this please?

smoothcontract:
Looking really good so far - will approve once the deployment issues to review app / Azure are resolved as I presume there will be further changes. 👍

config/initializers/new_framework_defaults_6_0.rb
tavy88: We might not need this actually as I created a new rails 6 app file and this is missing. Still reading about it.

PR #165, opened 2019-09-02
GET-383 user session

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-383

*  Move postcodes session read/write into user session
* Encapsulate job_profile_skills logic in user session rather than skills builder
* Rearrange user session methods per logic
* Add versioning to user_session:
  -  If user has existing session it is reset pre version
  - If user has existing session after versioning is set it remains intact
  - If skills builder switched between v1 and v2 session is reset
  - Add helper class for sessions as destroy method does not work on hashes
  - Fix up all tests that include sessions to work with versions

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -11,7 +11,7 @@ class PagesController < ApplicationController
     track_event(:pages_location_eligibility_search, search: postcode) if postcode.present?
     @search = CourseGeospatialSearch.new(postcode: postcode)
     if postcode && @search.valid?
-      user_session.store_at(key: :postcode, value: postcode)
+      user_session.postcode = postcode
 
       location_eligibility_through_courses
     end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -11,7 +11,7 @@ class PagesController < ApplicationController
     track_event(:pages_location_eligibility_search, search: postcode) if postcode.present?
     @search = CourseGeospatialSearch.new(postcode: postcode)
     if postcode && @search.valid?
-      session[:postcode] = postcode
+      user_session.store_at(key: :postcode, value: postcode)
 
       location_eligibility_through_courses
     end
--- a/app/controllers/skills_controller.rb
+++ b/app/controllers/skills_controller.rb
@@ -16,7 +16,7 @@ class SkillsController < ApplicationController
   def show_skills_builder
     return redirect_to task_list_path unless job_profile_and_skills_present
 
-    user_session.store_at(key: :current_job_id, value: job_profile.id)
+    user_session.current_job_id = job_profile.id
     @skills = Skill.find(skill_ids_for_job_profile)
     track_event(:skills_index_selected, job_profile.slug => skill_ids_for_job_profile)
   end
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -2,9 +2,16 @@ class UserSession
   attr_reader :session
 
   def initialize(session)
+    session.destroy unless session[:version] == expected_version
+
     @session = session
     @session[:visited_pages] ||= []
     @session[:job_profile_skills] ||= {}
+    @session[:version] ||= expected_version
+  end
+
+  def version
+    session[:version]
   end
 
   def postcode
@@ -71,5 +78,9 @@ class UserSession
 
   def current_job_id
     session[:current_job_id]
+  end
+
+  def expected_version
+    Flipflop.skills_builder_v2? ? 2 : 1
   end
 end
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -6,6 +6,10 @@ class UserSession
     @session[:visited_pages] ||= []
   end
 
+  def postcode
+    session[:postcode]
+  end
+
   def track_page(page_key)
     session[:visited_pages] << page_key unless session[:visited_pages].include?(page_key)
   end
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -2,9 +2,9 @@ class UserSession
   attr_reader :session
 
   def initialize(session)
-    session.destroy unless session[:version] == expected_version
-
     @session = session
+    @session.destroy unless version == expected_version
+
     @session[:visited_pages] ||= []
     @session[:job_profile_skills] ||= {}
     @session[:version] ||= expected_version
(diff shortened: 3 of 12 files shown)


app/models/user_session.rb
```diff
@@ -22,6 +46,18 @@ def job_profile_skills?
     job_profile_skills.keys.present?
   end
 
+  def job_profiles_cap_reached?
```
cwrw: just moved a few methods around to tidy up class, same with tests. I suggest looking at this class separately

app/controllers/pages_controller.rb
```diff
@@ -11,7 +11,7 @@ def location_eligibility
     track_event(:pages_location_eligibility_search, search: postcode) if postcode.present?
     @search = CourseGeospatialSearch.new(postcode: postcode)
     if postcode && @search.valid?
-      session[:postcode] = postcode
+      user_session.store_at(key: :postcode, value: postcode)
```
tavy88: Nice.

smoothcontract: Can we not have a writer method for postcode instead? It seems slightly odd that the pages controller needs to know which key should be used to store the postcode. I'd prefer

```ruby
user_session.postcode = postcode
```

smoothcontract: Within `UserSession` that can be implemented with a call to `store_at` if needed.

app/models/user_session.rb
```diff
@@ -38,29 +74,13 @@ def skill_ids_for_profile(id)
...
-    page_visited?('skills_matcher_index') &&
-      (page_visited?('training_hub') || page_visited?('next_steps'))
-  end
-
-  private
-
-  def job_profile_skills
-    session.fetch(:job_profile_skills, {})
+  def expected_version
+    Flipflop.skills_builder_v2? ? 2 : 1
```
tavy88: What would happen if you had 3 versions?

app/models/user_session.rb
```diff
@@ -2,18 +2,42 @@ class UserSession
   attr_reader :session
 
   def initialize(session)
+    session.destroy unless session[:version] == expected_version
```
smoothcontract: Please use `version` method here rather than `session[:version]` - I know it's the same code and potentially marginally less performant but it's nicer to read.

app/models/user_session.rb
```diff
@@ -2,18 +2,42 @@ class UserSession
...
+
+  def postcode
+    session[:postcode]
+  end
+
+  def job_profile_skills
+    session[:job_profile_skills]
   end
 
   def store_at(key:, value:)
```
smoothcontract: I'm tempted to say this method should be removed - consumers of the `UserSession` class should not have any visibility of the session keys. It's better that they work with dedicated methods for individual data items.

app/models/user_session.rb
```diff
@@ -2,18 +2,42 @@ class UserSession
...
     session[key] = value
   end
 
+  def set_skills_ids_for_profile(job_profile_id, skill_ids)
+    job_profile_skills[job_profile_id.to_s] = skill_ids
+  end
+
+  def track_page(page_key)
+    session[:visited_pages] << page_key unless session[:visited_pages].include?(page_key)
```
smoothcontract: Please use `unless page_visited?(page_key)` here instead for clarity

app/models/user_session.rb
```diff
@@ -2,18 +2,42 @@ class UserSession
   attr_reader :session
 
   def initialize(session)
+    session.destroy unless session[:version] == expected_version
+
     @session = session
     @session[:visited_pages] ||= []
+    @session[:job_profile_skills] ||= {}
+    @session[:version] ||= expected_version
```
tavy88: I'm not sure how one `version` key would work for multiple versioned features?

Example Scenario:

- SkillsBuilder has 2 versions (1, 2).
- Courses have 2 versions (1, 2).

How would one `version` key allow you to downgrade Courses feature from 2 -> 1 and upgrade SkillsBuilder from 1 -> 2

spec/support/session_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def destroy
+    clear
+  end
+end
+
+module SessionHelper
+  def create_fake_session(values, version: true)
+    expected_version = Flipflop.skills_builder_v2? ? 2 : 1
+    values = version ? values.merge(version: expected_version) : values
```
smoothcontract: I think this is cleaner:

```ruby
values.merge(version: expected_version) if version
```

spec/support/session_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+    super
+  end
+
+  def destroy
+    clear
+  end
+end
+
+module SessionHelper
+  def create_fake_session(values, version: true)
```
smoothcontract: Maybe rename this to `versioned` ? Consider it a nice to have :)

app/models/user_session.rb
smoothcontract: That can all be abstracted within `expected_version` - if our requirements get more complex in future then we can consider multiple flag combinations, but I would hope that's a rare scenario - and only matters when we make breaking changes to the session structure.

The scenario where we've given a different treatment to the same session structure across v1 and v2 skills builder is hopefully a rarity. I don't expect this will happen for every feature flag.

app/models/user_session.rb
cwrw: we can't use `version` here as this is before the initialisation where we set `@session = session`, so it will always be nil

cwrw: I'll try moving it down a line, see what happens

app/models/user_session.rb
tavy88: Sure @cwrw and I we were just discussing this and clarified it.

app/controllers/pages_controller.rb
```diff
@@ -11,7 +11,7 @@ def location_eligibility
     track_event(:pages_location_eligibility_search, search: postcode) if postcode.present?
     @search = CourseGeospatialSearch.new(postcode: postcode)
     if postcode && @search.valid?
-      session[:postcode] = postcode
+      user_session.postcode = postcode
```
tavy88: Great!

app/controllers/skills_controller.rb
```diff
@@ -16,7 +16,7 @@ def job_profile
   def show_skills_builder
     return redirect_to task_list_path unless job_profile_and_skills_present
 
-    user_session.store_at(key: :current_job_id, value: job_profile.id)
+    user_session.current_job_id = job_profile.id
```
tavy88: Great!

app/controllers/pages_controller.rb
smoothcontract: 👍

PR #167, opened 2019-09-05
GET-445 Add new Restricted database, and change session store

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-445
* Enable database switching in config
* Add restricted database with migrations path
* Add truncation to new db in tests
* Specify connection in models to which db to connect to
* Point restricted database credentials to primary db for now, this is until we pick our stategy for multiple db connections in azure
* Use active record session store gem to setup new table for sessions in restricted db
* Add config and specifically point to class as the different db connection is not automatically read
* Add new database setup to heroku review apps and azure
* Add new key to user_session to preserve order of job profile ids, as that is lost when written to db as hashmap

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -37,7 +37,7 @@ class UserSession
 
   def set_skills_ids_for_profile(job_profile_id, skill_ids)
     job_profile_skills[job_profile_id.to_s] = skill_ids
-    session[:job_profile_ids] << job_profile_id
+    session[:job_profile_ids] << job_profile_id unless job_profile_ids.include?(job_profile_id)
   end
 
   def track_page(page_key)
@@ -77,7 +77,7 @@ class UserSession
   def job_profile_ids
     return [current_job_id] if current_job?
 
-    session[:job_profile_ids].uniq
+    session[:job_profile_ids]
   end
 
   def skill_ids_for_profile(id)
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -7,6 +7,7 @@ class UserSession
 
     @session[:visited_pages] ||= []
     @session[:job_profile_skills] ||= {}
+    @session[:job_profile_ids] ||= []
     @session[:version] ||= expected_version
   end
 
@@ -36,6 +37,7 @@ class UserSession
 
   def set_skills_ids_for_profile(job_profile_id, skill_ids)
     job_profile_skills[job_profile_id.to_s] = skill_ids
+    session[:job_profile_ids] << job_profile_id
   end
 
   def track_page(page_key)
@@ -51,7 +53,7 @@ class UserSession
   end
 
   def job_profile_skills?
-    job_profile_skills.keys.present?
+    job_profile_ids.present?
   end
 
   def job_profiles_cap_reached?
@@ -75,7 +77,7 @@ class UserSession
   def job_profile_ids
     return [current_job_id] if current_job?
 
-    job_profile_skills.keys.flatten.map(&:to_i)
+    session[:job_profile_ids].uniq
   end
 
   def skill_ids_for_profile(id)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -108,7 +108,7 @@ stages:
         displayName: 'Primary Database Setup'
         env:
           RAILS_ENV: test
-          PRIMARY_DATABASE_URL: $(postgresUrlPrimary)
+          DATABASE_URL: $(postgresUrlPrimary)
 
       - script: |
           bundle exec rake db:migrate:restricted
@@ -121,7 +121,7 @@ stages:
           bundle exec rspec --format RspecJunitFormatter --out /build/TEST-rspec.xml
         displayName: 'Run Rspec tests'
         env:
-          PRIMARY_DATABASE_URL: $(postgresUrlPrimary)
+          DATABASE_URL: $(postgresUrlPrimary)
           RESTRICTED_DATABASE_URL: $(postgresUrlRestricted)
           SELENIUM_REMOTE_ADDRESS: $(GatewayIP)
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,17 +8,26 @@ pr: none
 
 variables:
   IMAGE_NAME: dfedigital/get-help-to-retrain
-  postgresUrl: postgres://test:test@localhost:5432/get-help-to-retrain
+  postgresUrlPrimary: postgres://test:test@localhost:5432/get-help-to-retrain-primary
+  postgresUrlRestricted: postgres://test:test@localhost:5433/get-help-to-retrain-restricted
   BUNDLE_CACHE_FOLDER: $(Pipeline.Workspace)/vendor/bundle
 
 resources:
   containers:
-  - container: 'postgres'
+  - container: 'postgres-primary'
     image: 'postgres:11-alpine'
     ports:
       - 5432:5432/tcp
     env:
-      POSTGRES_DB: 'get-help-to-retrain'
+      POSTGRES_DB: 'get-help-to-retrain-primary'
+      POSTGRES_USER: 'test'
+      POSTGRES_PASSWORD: 'test'
+  - container: 'postgres-restricted'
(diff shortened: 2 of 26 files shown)


app/models/user_session.rb
```diff
@@ -75,7 +77,7 @@ def skill_ids
   def job_profile_ids
     return [current_job_id] if current_job?
 
-    job_profile_skills.keys.flatten.map(&:to_i)
+    session[:job_profile_ids].uniq
```
smoothcontract: Could you change this to `.uniq!` please? I'm thinking that if someone continually updates the same job profile skills over and over then this array could grow in size. Adding the bang version updates the persisted session job id's too so the array would only ever have one duplicate prior to calling the method here.

smoothcontract: Alternatively, although I think it's slightly more complex, check that the job profile id is not already present before adding it on line 40.

azure-pipelines.yml
```diff
@@ -8,17 +8,26 @@ pr: none
...
       - 5432:5432/tcp
     env:
-      POSTGRES_DB: 'get-help-to-retrain'
+      POSTGRES_DB: 'get-help-to-retrain-primary'
+      POSTGRES_USER: 'test'
+      POSTGRES_PASSWORD: 'test'
+  - container: 'postgres-restricted'
+    image: 'postgres:11-alpine'
+    ports:
+      - 5433:5432/tcp
```
smoothcontract: This is mapping the default postgres port to be exposed on port 5433 instead, right? Why do the port numbers need to be different across the two containers?

smoothcontract: NVM - because they're all running on the same host as per line 12. I understand now ;)

app/models/user_session.rb
smoothcontract: One last thing - presumably `.uniq` and `.uniq!` both preserve the order of the array contents.

azure/template.json
```diff
@@ -391,6 +391,22 @@
                 "name": "DB_HOST",
                 "value": "[reference('postgresql-server').outputs.fullyQualifiedDomainName.value]"
               },
```
smoothcontract: So as per your explanation earlier, eventually these old DB_* settings would be removed

config/database.yml
```diff
@@ -1,89 +1,43 @@
...
-  # For details on connection pooling, see Rails configuration guide
-  # http://guides.rubyonrails.org/configuring.html#database-pooling
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
 
 development:
-  <<: *default
-  database: get_help_to_retrain_development
+  primary:
+    <<: *default
+    database: get_help_to_retrain_development
```
smoothcontract: We'll have to run some DB setup stuff to get local dev environments working anyway, so maybe it's worth taking the opportunity to rename this to `get_help_to_retrain_primary_development` for consistency?

Just a suggestion, not sure if it means additional effort/setup...

config/database.yml
```diff
@@ -1,89 +1,43 @@
...
-  <<: *default
-  database: get_help_to_retrain_test
+  primary:
+    <<: *default
+    database: get_help_to_retrain_test
+  restricted:
+    <<: *default
+    database: get_help_to_retrain_restricted_test
+    migrations_paths: db/restricted_migrate
+    url: <%= ENV['RESTRICTED_DATABASE_URL'] %>
```
smoothcontract: Is this here purely for Azure CI?

config/database.yml
smoothcontract: Perhaps if it's easy to rename existing local databases then we could change this name. It would be a hassle to rerun all of the data setup rake tasks though.

config/database.yml
```diff
@@ -1,89 +1,43 @@
...
+    database: <%= ENV['DB_DATABASE'] %>
+    username: <%= ENV['DB_USERNAME'] %>
+    password: <%= ENV['DB_PASSWORD'] %>
+  restricted:
+    <<: *default
+    host: <%= ENV['RESTRICTED_DB_HOST'] %>
+    database: <%= ENV['RESTRICTED_DB_DATABASE'] %>
+    username: <%= ENV['RESTRICTED_DB_USERNAME'] %>
+    password: <%= ENV['RESTRICTED_DB_PASSWORD'] %>
+    url: <%= ENV['RESTRICTED_DATABASE_URL'] || ENV['DATABASE_URL'] %>
```
smoothcontract: So this means that on a heroku review app both databases end up in the same physical postgres instance. I think that's fine and a nice compromise as it keeps the review apps at minimal cost. Do you not need to add this for the primary database though:

```
url: <%= ENV['DATABASE_URL' %>
```

...so that the app works on heroku for the primary postgres instance?

config/environments/production.rb
```diff
@@ -93,5 +92,8 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  config.active_record.database_selector = { delay: 2.seconds }
```
smoothcontract: What's the 2 second delay for?

config/database.yml
smoothcontract: Reading this again, this means that currently the heroku parent app has two databases and the review apps have one. That's a little confusing but I think it's okay so long as `RESTRICTED_DATABASE_URL` environment var is not inherited from the parent app.

smoothcontract: Maybe add a comment to clarify this arrangement in the YML file?

app/models/user_session.rb
cwrw: `uniq` preserves order as I initially checked. My initial thoughts on implementing this is with `uniq` was that it didn't feel very destructive for someone to keep adding job profiles to their array. I guess it's fair to assume because they will go back multiple times this might happen a few times. I'll add the check to only insert if not included

config/database.yml
cwrw: I would suggest we keep those the same, as we already have setup locally and elsewhere, and its safe to keep the default name as primary from the docs is referring to default generally

config/database.yml
cwrw: yes

smoothcontract: Maybe add a comment 'for Azure CI only' ?

config/database.yml
cwrw: The app is already deployed to https://get-help-to-retrain-pr-167.herokuapp.com/ and works fine without explicitly adding `DATABASE_URL` This is because `DATABASE_URL` is the default value for the database url setup.

The `RESTRICTED_DATABASE_URL` will not be inherited from the app as it is not defined in `app.json`. The reason the review app is different from the deployment app in setting up is that we copy the deployment app main database into the review app in the post deployment script. If we were to have one database in the deployment app we would end up copying both restricted/primary setup into the review app, this something that we don't want to do as we have sessions in the deployment database.

Since the review apps are temporary I'm ok with them talking to one database for now. If the need arises we can add more databases to the review apps then and amend this

I'll add a comment here to explain

config/environments/production.rb
cwrw: see the full explanation here: https://guides.rubyonrails.org/active_record_multiple_databases.html#activating-automatic-connection-switching

cwrw: I'd like to stick to the default 2 seconds unless we notice it is too slow for us

config/database.yml
smoothcontract: Very well explained and perfectly clear now, thankyou. A comment in the YML file would be helpful documentation 👍

config/environments/production.rb
smoothcontract: Read that and am a bit confused now.. that relates to setting up a secondary read only follower database - but here we have two writable databases so we always want to keep traffic directed to the relevant database depending on which model we'll be working with. I don't expect that to involve any on the fly switching as described in the linked article, so I'm not sure it's relevant to our use case and could perhaps be removed?

cwrw: Yes this is for different primary/replica dbs for both the main and restricted database. I added it to be safe in the future when we do have a primary/secondary. Might be ok to comment out for now but keep for later I guess

PR #169, opened 2019-09-05
Adding additional HTML data attributes for cypress

stevewalton28:
### Context
Adding additional HTML data attributes to support good practice for cypress tests to index and location_eligibility views

Adding three 'data-cy' attributes


stevewalton28:
Manually verified that attributes are present on the deployment and successfully ran cypress tests against the PR deployment.

PR #170, opened 2019-09-05
Get 443 notify service

tavy88:
### Context

Integrate with your Notify service. Currently implements the ability to send a notification (with/without personalised content) and get the status of a notification.

### Note

We are using a few templates for testing. You also need to set NOTIFY_API_KEY.

### JIRA

https://dfedigital.atlassian.net/browse/GET-443


spec/services/notify_service_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      let(:response) {
+        service.send_email(
+          email_address: email,
+          template_id: template_id,
+          options: options
+        )
+      }
+
+      it 'sends the message succesfully with customised content' do
+        expect(service.get_body_for(response)).to include('Hi this is a test with custom link: google.com')
```
tavy88: Here I'm tempted to actually rewrite this step and just ensure we pass the right payload and we trust Notify it replaces the placeholder. This way we spare one API call.

app/services/notify_service.rb
```diff
@@ -0,0 +1,42 @@
+require 'notifications/client'
```
smoothcontract: Is this still needed? I thought the whizzy new Rails 6 loader handles requires automatically..

app/services/notify_service.rb
```diff
@@ -0,0 +1,42 @@
...
+    ENV['NOTIFY_API_KEY']
+  end
+
+  def get_notification(response_id)
+    client.get_notification(response_id)
+  end
+
+  def log_error(error_message)
+    error_message
+    # TO DO We need to report this error as well to an Azure monitoring service similar to Rollbar
```
smoothcontract: This should be of the format `# TODO: foo` so that it works with `rails notes`.

spec/services/notify_service_spec.rb
smoothcontract: So is this test reliant on having a specific NOTIFY_API_KEY set with the exact guid for the email template and associated content? If so I think that's maybe more appropriate for a smoke test - I don't think we want to maintain specific settings in our local .env files for this; it's far too easy to inadvertently commit these.

Can this integration be stubbed instead perhaps?

smoothcontract: Probably best to discuss this in person :)

tavy88: Yeah, I agree it was more like part of TDD to make sure it does what it's supposed to. I think just making sure the right payload is passed is enough really.

app/services/notify_service.rb
tavy88: Apparently this gem needs it. If I remove it won't find the constant.

app/services/notify_service.rb
tavy88: Ah ok, cause I followed rubocop's suggestion as it was complaining. Will update.

.env.development
```diff
@@ -1 +1,2 @@
 USER_FEEDBACK_SMART_SURVEY_LINK=https://www.smartsurvey.co.uk/s/8IBQ1/
+NOTIFY_API_KEY=
```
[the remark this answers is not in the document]
tavy88: This file is used for documentation to let you know that you will need to set a KEY on local/test. This will be a key set in the VAULT. Let's talk in person as I feel there's a bit of miscommunication.

.env.test
```diff
@@ -1 +1,2 @@
 USER_FEEDBACK_SMART_SURVEY_LINK=https://www.smartsurvey.co.uk/s/8IBQ1/
+NOTIFY_API_KEY=
```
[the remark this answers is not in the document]
tavy88: This file is used for documentation to let you know that you will need to set a KEY on local/test. This will be a key set in the VAULT. Let's talk in person as I feel there's a bit of miscommunication.

app/services/notify_service.rb
```diff
@@ -0,0 +1,34 @@
+require 'notifications/client'
+
+class NotifyService
+  def send_email(email_address:, template_id:, options: {})
```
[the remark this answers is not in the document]
tavy88: Sorry, I don't really see much difference from my implementation?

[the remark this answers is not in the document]
tavy88: If you call this service in multiple controllers, I don't want to be passing the API key every time when I initialise.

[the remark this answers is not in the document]
tavy88: Also, I don't think `options` which are template placeholders should be mixed with the API key.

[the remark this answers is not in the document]
smoothcontract: I changed my mind slightly from the first comment - don't pass the api key as part of the `send_email` method call, sorry if that wasn't clear. But I do think the suggestion to add an initialize method with an optional api key is cleaner and easier to test. The use of this service in real code won't change, we'll still have something like `NotifyService.new.send_email(email_address: 'foo@bar.com', template_id: '123456')`

Key difference is that when testing you won't need to be stubbing references to `ENV` which I think is a bit hacky. Instead you can have tests e.g.

```ruby
context 'missing api key' do
  let(:service) { described_class.new(api_key: nil) }

  it 'raises an exception`
    ...
  end
end

context 'valid api key' do
  let(:service) { described_class.new(api_key: 'FAKE_API_KEY') }

  it 'does not raise an exception`
    ...
  end
end
```

app/services/notify_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    client.send_email(
+      email_address: email_address,
+      template_id: template_id,
```
smoothcontract: I suspect we'll need to add ENV var's for each of the template id's we want to use, but that will be in later stories that consume this service.

tavy88: Exactly.

app/services/notify_service.rb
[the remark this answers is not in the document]
tavy88: Ok, I agree with what you're suggesting for ease of testing. But what if you have 5 places where you use this service object, and at some point, you need to change the key from `NOTIFY_API_KEY` to `NEW_SERVICE_KEY`? You need to go and change in all 5 places because you pass the key every time?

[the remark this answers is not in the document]
smoothcontract: The api key argument is then optional so consumers of this class won't need to pass it and are unaware of the environment variable completely. So the environment variable is reference only on one line in the code. Only the rspec tests would pass the argument to override the default behaviour of using the environment variable for the API key.

PR #171, opened 2019-09-09
GET-450 Add user and passwordless migrations and models

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-450

Add passwordless gem (https://github.com/mikker/passwordless) to handle authentication and magic link generation. Passwordless requires a passwordless sessions table to store tokens. We also allow the tokens to only be used once by setting the default value to true and adding the claimed_at column to the table.

Passwordless allows emails to be sent by actionmailer. We will be using
Notify to add placeholder till we tackle that next.

In terms of session expiration, I am going with adhering with the defaults for now (1 year). We can override if we aren't happy with that.

As for sign in tokens expiration I am also going with the default fo 1 hour unless we aren't happy with that either.

To use passwordless, generate a user model with an email that should be unique. We are using citext to not worry about emails being case sensitive. A user model will be associated with one session at a time for now.


db/restricted_migrate/20190909104527_create_users.rb
```diff
@@ -0,0 +1,14 @@
+class CreateUsers < ActiveRecord::Migration[6.0]
+  def change
+    enable_extension 'citext'
```
tavy88: What is this for?

db/restricted_migrate/20190909104527_create_users.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    create_table :users do |t|
+      t.citext :email, null: false
```
smoothcontract: 👍

db/restricted_migrate/20190909104527_create_users.rb
smoothcontract: This adds support for case insensitive column types in postgres. I mentioned this before - if email is of type CITEXT then we can just do normal queries and postgres will perform case insensitive comparisons for equality.

https://www.postgresql.org/docs/9.1/citext.html

smoothcontract: i.e. `User.find_by(email: 'cAsEdOeSnOtMaTteR@example.com')` will work regardless of the case user used to provide their email

tavy88: Ah ok I thought it was something related to this gem.

PR #172, opened 2019-09-09
PID Collection

tavy88:
### Context

This feature is enabling us to capture personal
information about our users.

A couple of technical details:

1. The radio_button rails helper method wraps the
component with a div with class: field-with-errors breaking
the UI. We want to prevent this behaviour.

Here's a few resources around this:

https://stackoverflow.com/questions/22209855/radio-buttons-on-form-not-rendering-correctly-after-form-validation-failure-boo

The CSS approach won't solve the issue. Overriding approach:

From default behaviour:

```ruby
ActionView::Base.field_error_proc = Proc.new{ |html_tag, instance|
   "<div class=\"field_with_errors\">#{html_tag}</div>".html_safe
 }
```

```ruby
ActionView::Base.field_error_proc = Proc.new do |html_tag, instance|
  html_tag.html_safe
end
```

Works instead.

2. Handling the date of birth validation.

We collect the 3 components day/month/year separately from the form
and we put them together validating that the composed data is a
valid one.

3. The feature is flagged.

Ticket: https://dfedigital.atlassian.net/browse/GET-413

Some GDS error guides: https://design-system.service.gov.uk/components/radios/

# Screenshots

Screen Shot 2019-09-10 at 16 35 09
Screen Shot 2019-09-10 at 16 35 20

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -30,6 +30,10 @@ class PagesController < ApplicationController
   end
 
   def location_eligibility_through_courses
-    @search.find_courses.any? ? redirect_to(task_list_path) : redirect_to(location_ineligible_path)
+    if @search.find_courses.any?
+      Flipflop.user_personal_data? ? redirect_to(your_information_path) : redirect_to(task_list_path)
+    else
+      redirect_to(location_ineligible_path)
+    end
   end
 end
--- a/app/models/user_personal_data.rb
+++ b/app/models/user_personal_data.rb
@@ -1,0 +1,29 @@
+class UserPersonalData < RestrictedActiveRecordBase
+  attr_accessor :birth_day, :birth_month, :birth_year
+
+  enum gender: { female: 'female', male: 'male', not_mentioned: 'prefer not to say' }
+
+  validates :first_name, presence: true
+  validates :last_name, presence: true
+  validates :postcode, presence: true
+  validates :postcode, postcode: true, allow_blank: true
+  validate  :dob_format
+  validate  :dob_in_the_past, if: -> { dob? }
+  validates :gender, presence: true
+
+  private
+
+  def dob_format
+    self.dob = Date.new(birth_year.to_i, birth_month.to_i, birth_day.to_i)
+  rescue ArgumentError
+    add_invalid_dob_error
+  end
+
+  def dob_in_the_past
+    add_invalid_dob_error if dob > DateTime.now
+  end
+
+  def add_invalid_dob_error
+    errors.add(:dob, I18n.t('dob.invalid', scope: 'activerecord.errors.models.user_personal_data.attributes'))
+  end
+end
--- a/app/views/user_personal_data/_form.html.erb
+++ b/app/views/user_personal_data/_form.html.erb
@@ -1,0 +1,71 @@
+<%= form_with model: @user_personal_data, url: your_information_path, local: true do |f| %>
+  <%= form_group_tag @user_personal_data, :first_name do %>
+    <%= f.label :first_name, class: 'govuk-label' %>
+    <%= errors_tag @user_personal_data, :first_name %>
+    <%= f.text_field :first_name, class: css_classes_for_input(@user_personal_data, :first_name, 'govuk-input govuk-!-width-two-thirds') %>
+  <% end %>
+  <%= form_group_tag @user_personal_data, :last_name do %>
+    <%= f.label :last_name, class: 'govuk-label' %>
+    <%= errors_tag @user_personal_data, :last_name %>
+    <%= f.text_field :last_name, class: css_classes_for_input(@user_personal_data, :last_name, 'govuk-input govuk-!-width-two-thirds') %>
+  <% end %>
+  <%= form_group_tag @user_personal_data, :postcode do %>
+    <%= f.label 'Postcode', class: 'govuk-label' %>
+    <%= errors_tag @user_personal_data, :postcode %>
+    <%= f.text_field :postcode, class: css_classes_for_input(@user_personal_data, :postcode, 'govuk-input govuk-input--width-10') %>
+  <% end %>
+  <div class="govuk-form-group">
+    <%= form_group_tag @user_personal_data, :dob do %>
+      <h3 class="govuk-heading-m govuk-!-margin-bottom-0">
+        Date of birth
+      </h3>
+      <span id="passport-issued-hint" class="govuk-hint">
+        For example, 12 11 1980
+      </span>
+      <%= errors_tag @user_personal_data, :dob %>
+      <div class="govuk-date-input">
+        <div class="govuk-date-input__item">
+          <div class="govuk-form-group">
+            <%= f.label :day, class: 'govuk-label' %>
+            <%= f.number_field :birth_day, class: css_classes_for_input(@user_personal_data, :dob, 'govuk-input--width-2') %>
+          </div>
+        </div>
+        <div class="govuk-date-input__item">
+          <div class="govuk-form-group">
+            <%= f.label :month, class: 'govuk-label' %>
+            <%= f.number_field :birth_month, class: css_classes_for_input(@user_personal_data, :dob, 'govuk-input--width-2') %>
+          </div>
+        </div>
+        <div class="govuk-date-input__item">
+          <div class="govuk-form-group">
+            <%= f.label :year, class: 'govuk-label' %>
+            <%= f.number_field :birth_year, class: css_classes_for_input(@user_personal_data, :dob, 'govuk-input--width-4') %>
+          </div>
+        </div>
+      </div>
+    <% end %>
+  </div>
+  <div class="govuk-radios">
+    <%= form_group_tag @user_personal_data, :gender do %>
+      <h3 class="govuk-heading-m govuk-!-margin-bottom-2">
+        Gender
+      </h3>
+      <%= errors_tag @user_personal_data, :gender %>
+      <div class="govuk-radios__item">
+        <%= f.radio_button :gender, :male, class: 'govuk-radios__input' %>
+        <%= f.label :male, class: 'govuk-label govuk-radios__label' %>
+      </div>
+      <div class="govuk-radios__item">
+        <%= f.radio_button :gender, :female, class: 'govuk-radios__input' %>
+        <%= f.label :female, class: 'govuk-label govuk-radios__label' %>
+      </div>
+      <div class="govuk-radios__item">
+        <%= f.radio_button :gender, :not_mentioned, class: 'govuk-radios__input' %>
+        <%= f.label :prefer_not_to_say, class: 'govuk-label govuk-radios__label' %>
+      </div>
+    <% end %>
+  </div>
+  <p class="govuk-body govuk-!-margin-top-8">By continuing you agree to allow us to use your information as explained in our <%= link_to 'privacy policy', '#', class: 'govuk-link'  %>.</p>
+  <p class="govuk-body govuk-!-margin-top-8"><%= link_to 'Skip this step', skip_step_path, class:'govuk-link' %></p>
+  <%= f.button('Continue', class: 'govuk-button govuk-!-margin-top-0 govuk-button govuk-!-margin-bottom-2') %>
+<% end %>
--- a/app/views/user_personal_data/_form.html.erb
+++ b/app/views/user_personal_data/_form.html.erb
@@ -1,4 +1,4 @@
-<%= form_with model: @user_personal_data, url: your_information_path, local: true do |f| %>
+<%= form_with model: @user_personal_data, url: your_information_path, local: true, html: { novalidate: true } do |f| %>
   <%= form_group_tag @user_personal_data, :first_name do %>
     <%= f.label :first_name, class: 'govuk-label' %>
     <%= errors_tag @user_personal_data, :first_name %>
--- a/app/views/user_personal_data/index.html.erb
+++ b/app/views/user_personal_data/index.html.erb
@@ -1,0 +1,21 @@
+<% page_title :user_personal_data %>
+
+<% content_for :breadcrumb do
+    generate_breadcrumbs(
+      t('breadcrumb.user_personal_data'),
+      [
+        [t('breadcrumb.home'), root_path]
+      ]
+    )
+  end
+%>
+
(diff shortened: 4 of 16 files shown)


app/views/user_personal_data/_form.html.erb
```diff
@@ -0,0 +1,71 @@
...
+        <%= f.label :female, class: 'govuk-label govuk-radios__label' %>
+      </div>
+      <div class="govuk-radios__item">
+        <%= f.radio_button :gender, :not_mentioned, class: 'govuk-radios__input' %>
+        <%= f.label :prefer_not_to_say, class: 'govuk-label govuk-radios__label' %>
+      </div>
+    <% end %>
+  </div>
+  <p class="govuk-body govuk-!-margin-top-8">By continuing you agree to allow us to use your information as explained in our <%= link_to 'privacy policy', '#', class: 'govuk-link'  %>.</p>
+  <p class="govuk-body govuk-!-margin-top-8"><%= link_to 'Skip this step', task_list_path, data: { tracked_event: "Your Personal Information - Skip" }, target: '_blank' %></p>
```
tavy88: Let's discuss here. Not sure how we could do server side tracking here when clicking this link. We would need to track in the task-list controller what the previous page and track it then(that's what comes to my mind atm).

app/controllers/pages_controller.rb
```diff
@@ -30,6 +30,8 @@ def eligibility_params
   end
 
   def location_eligibility_through_courses
+    return redirect_to(your_information_path) if Flipflop.user_personal_data?
```
smoothcontract: We might have to discuss this behaviour with the wider team - looks like if the postcode is not eligible the user is still asked for their personal data but won't be able to make much use of the service. That feels like a poor exchange for the user - but is something that I don't think we've thought about before.

app/models/user_personal_data.rb
```diff
@@ -0,0 +1,33 @@
+class UserPersonalData < RestrictedActiveRecordBase
+  attr_accessor :birth_day, :birth_month, :birth_year
+
+  enum gender: { female: 'female', male: 'male', not_mentioned: 'prefer not to say' }
```
smoothcontract: Yay 👍

app/models/user_personal_data.rb
```diff
@@ -0,0 +1,33 @@
...
+  validate :dob_format
+  validate :dob_in_the_past, if: :dob_present?
+  validates :gender, presence: { message: I18n.t('user_personal_data.no_gender_error') }
+
+  private
+
+  def dob_format
+    self.dob = Date.new(birth_year.to_i, birth_month.to_i, birth_day.to_i)
+  rescue ArgumentError
+    errors.add(:dob, I18n.t('user_personal_data.invalid_birth_date_error'))
```
smoothcontract: Just a small thing, but this is repeated on line 23 so maybe worth DRYing into a method e.g. `add_invalid_dob_error`

app/models/user_personal_data.rb
```diff
@@ -0,0 +1,33 @@
...
+  attr_accessor :birth_day, :birth_month, :birth_year
+
+  enum gender: { female: 'female', male: 'male', not_mentioned: 'prefer not to say' }
+
+  validates :first_name, presence: { message: I18n.t('user_personal_data.no_first_name_error') }
+  validates :last_name, presence: { message: I18n.t('user_personal_data.no_last_name_error') }
+  validates :postcode, presence: { message: I18n.t('courses.no_postcode_error') }
+  validates :postcode, postcode: true, if: :postcode_present?
+  validate :dob_format
+  validate :dob_in_the_past, if: :dob_present?
```
smoothcontract: Not sure if this needs the `dob_present?` method - could you use the default `dob?` method instead or perhaps just use the `allow_blank` option instead. Same thing for the postcode on line 9.

app/models/user_personal_data.rb
```diff
@@ -0,0 +1,33 @@
+class UserPersonalData < RestrictedActiveRecordBase
+  attr_accessor :birth_day, :birth_month, :birth_year
+
+  enum gender: { female: 'female', male: 'male', not_mentioned: 'prefer not to say' }
+
+  validates :first_name, presence: { message: I18n.t('user_personal_data.no_first_name_error') }
```
smoothcontract: Rails has a convention for active validation locale keys, so it should be possible to structure those so that the message keys don't need to be explicitly defined here. Would be nice to align with that if possible. Have a look at https://guides.rubyonrails.org/i18n.html#translations-for-active-record-models

app/views/user_personal_data/_form.html.erb
```diff
@@ -0,0 +1,71 @@
...
+  <% end %>
+  <%= form_group_tag @user_personal_data, :last_name do %>
+    <%= f.label :last_name, class: 'govuk-label' %>
+    <%= errors_tag @user_personal_data, :last_name %>
+    <%= f.text_field :last_name, class: css_classes_for_input(@user_personal_data, :last_name, 'govuk-input govuk-!-width-two-thirds') %>
+  <% end %>
+  <%= form_group_tag @user_personal_data, :postcode do %>
+    <%= f.label 'Postcode', class: 'govuk-label' %>
+    <%= errors_tag @user_personal_data, :postcode %>
+    <%= f.text_field :postcode, class: css_classes_for_input(@user_personal_data, :postcode, 'govuk-input govuk-input--width-10'), value: f.object.postcode || user_session.postcode %>
```
smoothcontract: Not sure it's a nice approach to set the postcode from the session here. I think it would be cleaner to pass the session value in the controller on line 3 i.e.

```ruby
@user_personal_data = UserPersonalData.new(postcode: user_session.postcode)
```

app/views/user_personal_data/_form.html.erb
smoothcontract: I think this will need another action in the controller just for skipping. That can make the `track_event` call and then redirect to `task_list_path`. Leaving the original page open at this point isn't great UX.

app/views/user_personal_data/index.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      ]
+    )
+  end
+%>
+
+<div class="govuk-grid-row govuk-!-margin-top-7">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t('user_personal_data.title') %></h1>
+    <p class="govuk-body">We use this information to improve our service and make sure it's effective.</p>
+    <p class="govuk-body">We will share this information with other government departments.</p>
```
smoothcontract: 🕵️‍♀️

config/initializers/custom_errors.rb
```diff
@@ -0,0 +1,3 @@
+ActionView::Base.field_error_proc = proc do |html_tag, _instance|
+  html_tag.html_safe
```
smoothcontract: I think this is fine so long as it doesn't interfere with other error messaging.

config/locales/en.yml
```diff
@@ -107,6 +109,12 @@ en-GB:
     invalid_postcode_error: Enter a valid postcode
     nonexisting_postcode_error: Enter a real postcode
     no_postcode_error: Enter a postcode
+  user_personal_data:
```
smoothcontract: Maybe have a look at https://guides.rubyonrails.org/i18n.html#translations-for-active-record-models - if you move these to the correct keys for your model then you won't need to specify all of the key names in the model.

app/views/user_personal_data/_form.html.erb
tavy88: Why would I want to create another instance when the object is already carried adn it's accessible via f.object?

app/models/user_personal_data.rb
tavy88: 👍

app/models/user_personal_data.rb
tavy88: Yeah sure.

app/controllers/pages_controller.rb
tavy88: I thought the postcode capture at first page has nothing to do with the personal data one though. Is that not what the ticket says? You can override the postcode at personal information and it should never overwrite the postcode input for courses data.

app/models/user_personal_data.rb
smoothcontract: hmm... good question. I'm not sure. If that doesn't work then there's another approach you could try that the schools experience guys use with a lambda - this should work exactly the same as you have now but without the additional method:

```ruby
validate :dob_in_the_past, if: -> { dob.present? }
```

From this example:  https://github.com/DFE-Digital/schools-experience/blob/aafc248f6e5579ef5957f013cdb3dd4880178e6b/app/services/candidates/registrations/personal_information.rb

spec/requests/user_personal_data_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+          birth_year: '2014'
+        }
+      }
+    }
+
+    it 'creates the resource' do
+      expect {
+        post your_information_path, params: params
+      }.to change(UserPersonalData, :count).from(0).to(1)
+    end
```
tavy88: I think I'll add test steps for attribute values here as well.

app/views/user_personal_data/_form.html.erb
smoothcontract: I think you misunderstand. The controller code for `index` method looks like this now:

```ruby
def index
  @user_personal_data = UserPersonalData.new
end
```

If you change that to:

```ruby
def index
  @user_personal_data = UserPersonalData.new(postcode: user_session.postcode)
end
```

then you won't need to do anything special here in the view. The input can just be like this:

```ruby
<%= f.text_field :postcode, class: css_classes_for_input(@user_personal_data, :postcode, 'govuk-input govuk-input--width-10') %>
```
I think it's a better separation of concerns and more idiomatic approach. This will default the initial postcode value when the page loads from the session and then use the value the user enters for postcode from then on which I think is reasonable UX. If the user deliberately clears the postcode and submits the form, it should show the empty value they just submitted with an error message. I'm not entirely sure what the behaviour will be with the view as you have it - presumably it would show an error message but also show their original postcode.

tavy88: Ok, I see what you mean now. Yeah, it should work.

app/controllers/pages_controller.rb
tavy88: Will think about this.

app/models/user_personal_data.rb
tavy88: They are probably doing it the same reason I am.

smoothcontract: Yep - I think as the lambda is very small it's marginally better than defining a method with the same expression.

smoothcontract: I just did some experimentation in the console and I think using the default generated `postcode?` method will also work. For a string value in a model that returns false for nil, or a blank string so is analogous to `postcode.present?`. Either way I don't think you need the extra methods.

app/views/user_personal_data/_form.html.erb
tavy88: I think we can just pass a query string on the E.g 

```ruby
/task-list?ref=pid
```

And then we track server-side in the task-list controller as we know they came from PID page. 

Why would we need another controller?

app/controllers/pages_controller.rb
smoothcontract: Yes, that the correct behaviour as you describe. The postcode on personal info page is only defaulted from the session but when user submits personal info the session is unaffected. I think we're talking about two different things.

From the code above (which I haven't tested, so am guessing a little bit), it seems that the user gets to this method after submitting the postcode eligibility form. And the personal data form is then shown immediately afterwards if the new feature is enabled. But.. this is skipping the logic on line 35 entirely so the postcode eligibility check is now not being performed.

I don't think that's the correct behaviour - postcode eligibility should still be tested. What we need to discuss with the wider team is what happens if the postcode eligibility fails - do we ask them for their personal info anyway, should we defer the eligibility check until after personal info is captured etc.

config/locales/en.yml
```diff
@@ -121,3 +126,17 @@ en-GB:
...
+        user_personal_data:
+          attributes:
+            first_name:
+              blank: Enter your first name
+            last_name:
+              blank: Enter your last name
+            gender:
+              blank: Enter a gender or chose 'Prefer not to say'
+            postcode:
+              blank: Enter a postcode
```
smoothcontract: 👏

spec/requests/user_personal_data_spec.rb
smoothcontract: Cool, the `have_attributes` matcher is a good fit for this as you can test all of the attributes in a single expectation. There are a few examples of that being used elsewhere in the specs (e.g. `course_import_service_spec.rb`).

app/views/user_personal_data/_form.html.erb
smoothcontract: Either approach works, I don't mind. Generally having a separate action means you won't have conditional logic in the shared action so will end up being cleaner. I'm thinking of something like this (in the personal info controller):

```ruby
def skip
  track_event(:skipped_personal_info)
  redirect_to task_list_path
end
```

Rather than this which adds complexity and bleeds responsibility into the task list controller:

```ruby
def task_list
  track_event(:skipped_personal_info) if params[:pid].present?
end
```

With the redirect suggestion this also avoids submitting multiple tracking events if the user refreshes the task list page, plus keeps the url looking cleaner.

tavy88: Having a large number of redirects can cause your loading speed to slow down to a certain extent.

https://varvy.com/pagespeed/minimize-redirects.html

app/controllers/pages_controller.rb
tavy88: Confirmed with Lois
Screen Shot 2019-09-11 at 11.33.59.png 
Screen Shot 2019-09-11 at 11.33.59.png


Will still lead to PID

app/views/user_personal_data/_form.html.erb
smoothcontract: Hmm... not sure I fully subscribe to "minimize redirects" as an absolute rule. Save and redirect is an established pattern used for destructive or one-off actions. In this case, like I said I don't mind too much which approach you go with but it should definitely be a server side tracking action.

PR #174, opened 2019-09-12
GET-27 Allow Users to save their progress and sign in

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-27


spec/features/user_registration_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+  scenario 'User sent sign in email if already registered' do
+    register_user
+    register_user
+
+    expect(client).to have_received(:send_email)
+      .with(confirmation_email)
+      .with(sign_in_email)
+  end
+
+  scenario 'User sent sign in email if they click send email again' do
```
cwrw: add tests for different case in email entered

PR #176, opened 2019-09-12
Updated ERD rake task

smoothcontract:
### Context
The rake task to generate an Entity Relationship Diagram kinda broke when we moved to Rails 6. It doesn't include all of the relations for the primary database, and none of the models in the restricted database.

I added a dedicated rake task to our repo. It seems the fundamental issue is that the erd library is using `Rails.application.eager_load!` to load all of the model classes into memory and then generate the diagram from there. But this fails with Rails 6 (or potentially because we introduced two databases).

So I've taken an alternate approach to just reference each of the relevant model classes; this is sufficient to lazy load the classes and they're then included in the diagram. And since we have two separate databases now it made sense to split the ERD into two separate files.

This is more explicit now so that we only get the named models on each diagram, but it does mean that the list of models needs to be maintained in the new rake task.

### Guidance to review
See updated README - but essentially it's still a case of running `rails erd`.

Two files should be produced as per the attachments.

[primary_erd.pdf](https://github.com/DFE-Digital/get-help-to-retrain/files/3606443/primary_erd.pdf)
[restricted_erd.pdf](https://github.com/DFE-Digital/get-help-to-retrain/files/3609258/restricted_erd.pdf)


lib/tasks/erd.rake
```diff
@@ -0,0 +1,26 @@
+desc 'Generate Entity-Relationship diagrams for primary and restricted models'
+task erd: :environment do
+  # Explicitly list models to include on each diagram
+  primary_models = [JobProfile, Category, Skill, JobProfileCategory, JobProfileSkill, Course]
```
cwrw: instead of manually adding those, is there a way we could list the primary models, maybe `PrimaryActiveRecordBase.descendants`? Same below

smoothcontract: No, sadly not easily. I spent a LONG time trying to make this work nicely including iterating over the model directory files. That worked to a point but meant all of the models showed up on the same diagram. It also didn't pick up the Passwordless stuff as that's in a separate gem.

I'm resigned to explicitly listing the classes that get included on each diagram - have burned too much time on this already.

cwrw: So something like this should work:
```
Zeitwerk::Loader.eager_load_all
PrimaryActiveRecordBase.subclasses.map(:name)
RestrictedActiveRecordBase.subclasses.map(:name) + ['PasswordlessSession']
```
what do you think?

smoothcontract: The key problem is getting the models loaded into memory - just referencing the class names in the array is enough to trigger that reliably. But all the usual approaches of trying to kick off Rails eager loading are not working as expected - which is the cause of the original problems with the ERD missing out lots of stuff.

PR #178, opened 2019-09-13
Add Google Analytics tracking snippet

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-352

Refactored the config for smart survey so that `ENV` no longer needs to be stubbed - this was conflicting when I added the GA snippet as there are then two ENV vars in the application layout. So now `Rails.configuration` is updated with the two relevant environment variables which makes it easier to stub in tests.

Moved logic for including the relevant markup to the partial itself which makes the main application layout a bit cleaner (along with other minor cleanup).

### Guidance to review
The new env var GOOGLE_ANALYTICS_TRACKING_ID will only be defined on the QA environment, but I'll temporarily set that up on the Heroku review app once deployed so it can be tested. I'm not sure if this is enough to start sending tracking info to GA - suspect we'll need to do some more steps to trust the tracking source, particularly if it comes from heroku review app.


config/application.rb
```diff
@@ -31,5 +31,8 @@ class Application < Rails::Application
     config.exceptions_app = routes
 
     config.i18n.default_locale = :'en-GB'
+
+    config.google_analytics_tracking_id = ENV['GOOGLE_ANALYTICS_TRACKING_ID']
+    config.smart_survey_user_feedback_link = ENV['USER_FEEDBACK_SMART_SURVEY_LINK']
```
tavy88: Should we do the same with our `NotifyService` ?

tavy88:
Also rubocop doesn't look happy.

config/application.rb
smoothcontract: Good idea - just added that too

.env.test
```diff
@@ -1,2 +1,3 @@
 USER_FEEDBACK_SMART_SURVEY_LINK=https://www.smartsurvey.co.uk/s/8IBQ1/
 NOTIFY_API_KEY=test
```
tavy88: I don't think we need to give the `test` value here. I know it's not your change, but I would like if we could remove it.

smoothcontract: Outside the scope of this PR I think, but we should discuss again.

PR #179, opened 2019-09-13
GET-412 Add return to saved results

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-27

Add return to saved results partial which send a registered user a magic link to sign in through. Please go commit through commit as those are designed to make reviewing eaiser

### DIFF
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -2,6 +2,18 @@ load Passwordless::Engine.root.join('app/controllers/passwordless/sessions_contr
 Passwordless::SessionsController.class_eval do
   after_action :restore_user_session, only: [:show], if: -> { current_user }
 
+  def show
+    BCrypt::Password.create(params[:token])
+
+    sign_in passwordless_session
+
+    redirect_to main_app.task_list_path
+  rescue Passwordless::Errors::TokenAlreadyClaimedError
+    raise ActionController::RoutingError, 'Not Found'
+  rescue Passwordless::Errors::SessionTimedOutError
+    raise ActionController::RoutingError, 'Not Found'
+  end
+
   def restore_user_session
     current_user.restore_session(session)
   end
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -3,6 +3,7 @@ Passwordless::SessionsController.class_eval do
   after_action :restore_user_session, only: [:show], if: -> { current_user }
 
   def show
+    # Make it "slow" on purpose to make brute-force attacks more of a hassle
     BCrypt::Password.create(params[:token])
 
     sign_in passwordless_session
--- a/app/controllers/passwordless/sessions_controller.rb
+++ b/app/controllers/passwordless/sessions_controller.rb
@@ -3,6 +3,6 @@ Passwordless::SessionsController.class_eval do
   after_action :restore_user_session, only: [:show], if: -> { current_user }
 
   def restore_user_session
-    current_user.restore_session(user_session)
+    current_user.restore_session(session)
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -7,10 +7,20 @@ class UsersController < ApplicationController
   def create
     @user = User.find_or_initialize_by(email: user_params[:email])
     if @user.valid?
-      setup_user
+      register_user
       render(:show)
     else
       render(:new)
+    end
+  rescue NotifyService::NotifyAPIError
+    # TODO: show user an error page, for now render link sent page
+    render(:show)
+  end
+
+  def sign_in
+    @user = User.find_or_initialize_by(email: user_params[:email])
+    unless @user.valid?
+      redirect_back(fallback_location: root_path, flash: { error: @user.errors[:email] })
     end
   rescue NotifyService::NotifyAPIError
     # TODO: show user an error page, for now render link sent page
@@ -23,12 +33,16 @@ class UsersController < ApplicationController
     params.permit(:email, :authenticity_token)
   end
 
-  def setup_user
+  def register_user
     if @user.new_record?
       @user.register_new_user(user_session, request.base_url)
     else
-      passwordless_session = build_passwordless_session(@user)
-      @user.register_existing_user(passwordless_session, request.base_url, user_session)
+      sign_in_user
     end
+  end
+
+  def sign_in_user
+    passwordless_session = build_passwordless_session(@user)
+    @user.register_existing_user(passwordless_session, request.base_url, user_session)
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -8,18 +8,21 @@ class UsersController < ApplicationController
     @user = User.find_or_initialize_by(email: user_params[:email])
     if @user.valid?
       register_user
-      render(:show)
+      render(:registration_link_sent)
     else
       render(:new)
     end
   rescue NotifyService::NotifyAPIError
     # TODO: show user an error page, for now render link sent page
-    render(:show)
+    render(:registration_link_sent)
   end
 
   def sign_in
     @user = User.find_or_initialize_by(email: user_params[:email])
-    unless @user.valid?
+    if @user.valid?
+      sign_in_user unless @user.new_record?
+      redirect_to(link_sent_path(email: @user.email))
+    else
       redirect_back(fallback_location: root_path, flash: { error: @user.errors[:email] })
     end
   rescue NotifyService::NotifyAPIError
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -27,7 +27,7 @@ class UsersController < ApplicationController
     end
   rescue NotifyService::NotifyAPIError
     # TODO: show user an error page, for now render link sent page
-    render(:show)
+    redirect_to(link_sent_path(email: @user.email))
   end
 
   private
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -42,10 +42,11 @@ class UsersController < ApplicationController
     else
       sign_in_user
     end
+    user_session.registered = true
   end
 
   def sign_in_user
     passwordless_session = build_passwordless_session(@user)
-    @user.register_existing_user(passwordless_session, request.base_url, user_session)
+    @user.register_existing_user(passwordless_session, request.base_url)
   end
 end
--- a/app/helpers/authentication_helper.rb
+++ b/app/helpers/authentication_helper.rb
@@ -1,0 +1,14 @@
+module AuthenticationHelper
+  def user_with_validation
+    @user_with_validation ||= begin
+      authentication_user.errors.add(:email, flash[:error].first) if flash[:error].present?
+      authentication_user
+    end
+  end
+
+  private
+
+  def authentication_user
+    @authentication_user ||= User.new
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -32,7 +32,10 @@ class User < RestrictedActiveRecordBase
   def restore_session(new_session)
     return unless new_session
 
-    UserSession.merge_session(new_session, session.data)
+    UserSession.merge_sessions(
+      new_session: new_session,
+      previous_session_data: session.data
+    )
     add_session_to_user(new_session)
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -31,11 +31,11 @@ class User < RestrictedActiveRecordBase
     )
   end
 
-  def restore_session(user_session)
-    return unless user_session
+  def restore_session(new_session)
+    return unless new_session
 
-    user_session.merge_session(session.data)
-    add_session_to_user(user_session.session)
+    UserSession.merge_session(new_session, session.data)
+    add_session_to_user(new_session)
   end
 
   private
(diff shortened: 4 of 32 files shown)


app/controllers/users_controller.rb
```diff
@@ -7,14 +7,27 @@ def new
...
+  end
+
+  def sign_in
+    @user = User.find_or_initialize_by(email: user_params[:email])
+    if @user.valid?
+      sign_in_user unless @user.new_record?
+      redirect_to(link_sent_path(email: @user.email))
+    else
+      redirect_back(fallback_location: root_path, flash: { error: @user.errors[:email] })
+    end
```
tavy88: Could we make this more readable:

```ruby
@user = User.find_or_initialize_by(email: user_params[:email])

redirect_back(fallback_location: root_path, flash: { error: @user.errors[:email] }) unless @user.valid?

sign_in_user unless @user.new_record?

redirect_to(link_sent_path(email: @user.email))
```

cwrw: I prefer the current implementation and readability, as it shows that there are two different paths, rather than hiding them all at the same level

app/helpers/authentication_helper.rb
```diff
@@ -0,0 +1,14 @@
+module AuthenticationHelper
+  def user_with_validation
+    @user_with_validation ||= begin
+      authentication_user.errors.add(:email, flash[:error].first) if flash[:error].present?
```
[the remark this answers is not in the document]
cwrw: I don't think this warrants a new method as `user_with_validation` is basically the method where we add validations (this line), and return the authenticated user, already defined in a new method.

app/controllers/users_controller.rb
```diff
@@ -23,12 +36,17 @@ def user_params
...
-  def setup_user
+  def register_user
     if @user.new_record?
       @user.register_new_user(user_session, request.base_url)
     else
-      passwordless_session = build_passwordless_session(@user)
-      @user.register_existing_user(passwordless_session, request.base_url, user_session)
+      sign_in_user
     end
+    user_session.registered = true
```
tavy88: Same here:

```ruby
sign_in_user unless @user.new_record?

@user.register_new_user(user_session, request.base_url)
user_session.registered = true

```

cwrw: This breaks the actual flow of events we don't want to sign in and register users, we want to either sign in OR register users

app/models/user.rb
```diff
@@ -31,11 +29,11 @@ def register_existing_user(passwordless_session, url, user_session)
...
   end
 
-  def restore_session(user_session)
-    return unless user_session
+  def restore_session(new_session)
+    return unless new_session
 
-    user_session.merge_session(session.data)
-    add_session_to_user(user_session.session)
+    UserSession.merge_session(new_session, session.data)
```
tavy88: Could you maybe use a more explicit method that indicates what `session.data` is? E.g

```ruby
def last_session
session.data
end

UserSession.merge_session(new_session, last_session)
```

tavy88: Wonder if this could be named to `merge_sessions` and then it would take as args 2 sessions:

```ruby
UserSession.merge_sessions(new_session, session)
```

app/models/user_session.rb
```diff
@@ -10,6 +10,10 @@ class UserSession
     version
   ].freeze
 
+  def self.merge_session(new_session, previous_session_data)
+    new_session.merge!(previous_session_data.slice(*KEYS_TO_RESTORE))
```
tavy88: And then here you could do

```ruby
def self.merge_session(new_session, previous_session)
    new_session.merge!(previous_session.data.slice(*KEYS_TO_RESTORE))
```

app/models/user.rb
tavy88: And then check https://github.com/DFE-Digital/get-help-to-retrain/pull/179/files#diff-dac2a7a42571b251f8887d7149fb7ee9R13

app/views/shared/_return_to_saved_results.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <div class="govuk-related-items govuk-!-padding-bottom-4 top-border__container">
+    <h2 class="govuk-heading-m"><%= t('return_to_saved_results.title') %></h2>
+    <p class="govuk-body"><%= t('return_to_saved_results.sub_title') %></p>
+    <%= form_with local: true, url: sign_in_path do |form| %>
+      <%= form_group_tag user_with_validation, :email do %>
+        <label class="govuk-label" for="email">
+          Email
+        </label>
+        <%= errors_tag user_with_validation, :email %>
+        <%= text_field_tag(:email, params.fetch(:email, nil), class: css_classes_for_input(user_with_validation, :email, " govuk-!-width-full")) %>
```
tavy88: I think you can use `form.text_field` as you are not passing any model object in `form_with`. https://medium.com/@tinchorb/form-with-building-html-forms-in-rails-5-1-f30bd60ef52d

app/views/shared/_return_to_saved_results.html.erb
```diff
@@ -0,0 +1,17 @@
+<% return unless user_not_authenticated? %>
+<div class="govuk-grid-column-one-third">
+  <div class="govuk-related-items govuk-!-padding-bottom-4 top-border__container">
+    <h2 class="govuk-heading-m"><%= t('return_to_saved_results.title') %></h2>
+    <p class="govuk-body"><%= t('return_to_saved_results.sub_title') %></p>
+    <%= form_with local: true, url: sign_in_path do |form| %>
+      <%= form_group_tag user_with_validation, :email do %>
+        <label class="govuk-label" for="email">
+          Email
+        </label>
```
tavy88: Can we use Rails's helper methods here:

```ruby
<%= form.label :email, class: 'govuk-label' %>
```

app/controllers/passwordless/sessions_controller.rb
```diff
@@ -2,7 +2,19 @@
 Passwordless::SessionsController.class_eval do
   after_action :restore_user_session, only: [:show], if: -> { current_user }
 
+  def show
+    BCrypt::Password.create(params[:token])
```
smoothcontract: where does the output of this method go? it’s not assigned to anything...

smoothcontract: ah, does it throw an exception?

app/models/user.rb
```diff
@@ -11,18 +11,16 @@ def register_new_user(user_session, url)
     return unless user_session && url
 
     add_session_to_user(user_session.session)
-    user_session.registered = true
     notify_service.send_email(
       email_address: email,
       template_id: NotifyService::CONFIRMATION_TEMPLATE_ID,
```
tavy88: I can't remember but I think the consensus was that `NotifyService::CONFIRMATION_TEMPLATE_ID` are coming from ENV in this PR? Ignore me if you plan on doing this on another one.

app/models/user.rb
cwrw: We don't need to pass the whole previous session in, there is no need to pass whole objects if they are not needed, we should only be passing the objects/data we need

app/models/user_session.rb
cwrw: see above

app/models/user.rb
tavy88: I can't remember but I think the consensus was that NotifyService::CONFIRMATION_TEMPLATE_ID are coming from ENV in this PR? Ignore me if you plan on doing this on another one. You can use Andrew's approach for the rest of the ENV keys maybe (https://github.com/DFE-Digital/get-help-to-retrain/pull/178/files#diff-b1fe55db50c712fef0673345e5b9c0d9)

spec/features/user_sign_in_spec.rb
```diff
@@ -39,10 +64,118 @@ def sign_in_user
...
+    send_sign_in_email
+
+    expect(page).to have_text(/test@test.test/)
+  end
+
+  scenario 'User does not get error message if they enter their same email but different case' do
+    register_user
+    visit(root_path)
+    fill_in('email', with: 'Test@Test.Test')
+    page.driver.header('User-Agent', 'some-agent')
```
smoothcontract: Why do we care about the user agent?

app/controllers/users_controller.rb
tavy88: Interesting. Why are you confused about my suggestion @smoothcontract ?

app/models/user.rb
cwrw: I've decided not to add this due to the following reasons:
1) Those are not sensitive data, and are unlikely to change
2) Overhead in adding those keys to 4 different environments and heroku, and azure templates in code/key vaults
3) If I was to change this, instead of changing in one place I have to change it in at least 4 places for both keys.

So I think its best to keep them in one place here instead

app/controllers/users_controller.rb
tavy88: I think it does the same as your code does it not?

app/helpers/authentication_helper.rb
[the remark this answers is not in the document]
tavy88: It would just read better. It looks like this method is doing a lot. It's just a personal preference and I find it reads better.

app/models/user.rb
tavy88: Yeah I know that but it's not very obvious what `UserSession.merge_session(new_session, session.data)` does when you first see it.

app/views/shared/_return_to_saved_results.html.erb
cwrw: It's the opposite, `text_field` usually accepts a model, from rails docs and the link you sent: https://apidock.com/rails/ActionView/Helpers/FormTagHelper/text_field_tag
https://apidock.com/rails/ActionView/Helpers/FormHelper/text_field

app/controllers/users_controller.rb
tavy88: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause

spec/features/user_sign_in_spec.rb
cwrw: `Passwordless::Session` model validations depend on it, and mark an instance as invalid otherwise. Since in those tests user agent isn't defined we are doing that explicitly

app/controllers/passwordless/sessions_controller.rb
cwrw: I didn't want to amend the controller action from the gem, it has the following comment over it: 
# Make it "slow" on purpose to make brute-force attacks more of a hassle

app/controllers/users_controller.rb
cwrw: This is not a guard though, this is quiet different, there are two different flows instead of one

app/models/user.rb
tavy88: You will also need to do a deployment if the templates need changing though. The whole end to end process can take up to 1 hour instead of 2 mins. What do you think?

app/models/user.rb
tavy88: Maybe `UserSession.merge_session(new_session, with: session.data)` at least, although I get your point about passing unnecessary stuff.

app/views/shared/_return_to_saved_results.html.erb
tavy88: I am talking about `form_with` have you seen my sample code above?

app/controllers/passwordless/sessions_controller.rb
smoothcontract: Okay, so it doesn't throw an exception or return a response but does slow the request. May be worth including the original comment from the gem too? Unless there's a cleaner way to modify the default behaviour of the gem than reimplementing this method 🙏

app/controllers/users_controller.rb
tavy88: Ok, Rubocop style guides also suggest my approach. 🤷‍♂

app/models/user.rb
smoothcontract: I've revised my stance on this a little. Other than the sensitive values which are kept in the Azure vault, other config settings are externalised in the Azure templates within this repo, which eventually get passed to the application as environment variables. So to change those we still need to make a code change and go through the whole Azure deployment cycle, plus there's more stuff to maintain in the azure templates etc.

I'm not very happy about the situation, but given how much of a pain to work with Azure seems to be, moving these to external config doesn't actually achieve the benefits we would normally expect. And as mentioned, the template id's are not sensitive. If we do get to do A/B testing or anything interesting like that we can revise this approach. This is adequate for what we need now though.

spec/features/user_sign_in_spec.rb
smoothcontract: cool 👍

app/views/shared/_return_to_saved_results.html.erb
smoothcontract: Rather than `params.fetch(:email, nil)` which is a little redundant, could you use simply `params[:email]` which I think behaves the same but is a bit more succinct?

cwrw: I'm not sure I appreciate the tone with the previous comment. I have seen your code, and read the link and responded appropriately. I missed the deprecated part though in the link, which should be the main driver for the amendment here

tavy88: Abeer not sure why you feel there was any tone. I said it in the kindest way possible that with `form_with` you don't need the _tag fields and wanted to make sure you checked the link I sent which has both examples proving what I'm trying to say. We can jump on a call to clarify. You did misinterpret my tone completely I assure you.

PR #180, opened 2019-09-16
GET-412 Retain email on validation on sign in, and add anchor to url

cwrw:
When a user adds an invalid email and is redirected back from the sign in
page, we should retain the email. We pass this as a query param in the redirect url.
Since `redirect_back` does not allow us to do this, build this functionality ourselves.
Also add an anchor to the email to make sure we scroll back down to the sign in form.

Move previous redirect logic to same URL Parser class as it shares the url is part of host
functionally.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,7 +1,7 @@
 class UsersController < ApplicationController
   def new
     @user = User.new
-    user_session.registration_triggered_from(request.referer, [save_your_results_path]) if request.referer
+    set_redirect_path_for_registration
   end
 
   def create
@@ -23,7 +23,7 @@ class UsersController < ApplicationController
       sign_in_user unless @user.new_record?
       redirect_to(link_sent_path(email: @user.email))
     else
-      redirect_back(fallback_location: root_path, flash: { error: @user.errors[:email] })
+      redirect_to(build_redirect_url || root_path, flash: { error: @user.errors[:email] })
     end
   rescue NotifyService::NotifyAPIError
     # TODO: show user an error page, for now render link sent page
@@ -48,5 +48,26 @@ class UsersController < ApplicationController
   def sign_in_user
     passwordless_session = build_passwordless_session(@user)
     @user.register_existing_user(passwordless_session, request.base_url)
+  end
+
+  def set_redirect_path_for_registration
+    return unless request.referer
+
+    redirect_url = url_parser.get_redirect_path(paths_to_ignore: [save_your_results_path])
+    user_session.registration_triggered_path = redirect_url
+  end
+
+  def build_redirect_url
+    return unless request.referer
+
+    url_parser.build_redirect_url_with(
+      param_name: 'email',
+      param_value: @user.email,
+      anchor: 'sign-in'
+    )
+  end
+
+  def url_parser
+    @url_parser ||= UrlParser.new(request.referer, request.host)
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -51,15 +51,11 @@ class UsersController < ApplicationController
   end
 
   def set_redirect_path_for_registration
-    return unless request.referer
-
     redirect_url = url_parser.get_redirect_path(paths_to_ignore: [save_your_results_path])
     user_session.registration_triggered_path = redirect_url
   end
 
   def build_redirect_url
-    return unless request.referer
-
     url_parser.build_redirect_url_with(
       param_name: 'email',
       param_value: @user.email,
--- a/app/lib/url_parser.rb
+++ b/app/lib/url_parser.rb
@@ -1,0 +1,33 @@
+class UrlParser
+  attr_reader :uri, :host
+
+  def initialize(referer, host)
+    @uri = URI(referer)
+    @host = host
+  end
+
+  def build_redirect_url_with(param_name:, param_value:, anchor:)
+    return unless recognized_host?
+
+    query = Rack::Utils.parse_query(uri.query)
+    query[param_name] = param_value
+    uri.query = Rack::Utils.build_query(query)
+    uri.fragment = anchor
+    uri.to_s
+  end
+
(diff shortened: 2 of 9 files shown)


app/controllers/users_controller.rb
```diff
@@ -49,4 +49,25 @@ def sign_in_user
     passwordless_session = build_passwordless_session(@user)
     @user.register_existing_user(passwordless_session, request.base_url)
   end
+
+  def set_redirect_path_for_registration
+    return unless request.referer
+
+    redirect_url = url_parser.get_redirect_path(paths_to_ignore: [save_your_results_path])
+    user_session.registration_triggered_from(redirect_url)
```
tavy88: This looks like it persists something on the session, but from the name, it's not very obvious what it does unless you check it out. Would you consider maybe renaming to `persist_<something>`?

app/controllers/users_controller.rb
```diff
@@ -49,4 +49,25 @@ def sign_in_user
     passwordless_session = build_passwordless_session(@user)
     @user.register_existing_user(passwordless_session, request.base_url)
   end
+
+  def set_redirect_path_for_registration
```
tavy88: Same here maybe use `persist_redirect_path_for_registration`, because at the end of the day that what this whole method does.

app/controllers/users_controller.rb
cwrw: I think using ` set` in the method name should make it obvious? as we aren't really "persisting" anything in the db, which persist would  insinuate. I could add a bang `!` to help with that

cwrw: actually scratch that, `!` does not fit this case

cwrw: oh you mean this method itself, so this is in `user_session`, I can change this to follow other examples we've done as in `user_session.registration_triggered_from = value`

app/controllers/users_controller.rb
cwrw: please see my comment about `persist` vs `set`

app/controllers/users_controller.rb
```diff
@@ -49,4 +49,25 @@ def sign_in_user
...
+    return unless request.referer
```
smoothcontract: I'm not sure you need this line.. if `request.referer` is nil then an exception will be thrown within the `UrlParser`:

```ruby
> URI(nil)
Traceback (most recent call last):
        1: from (irb):1
ArgumentError (bad argument (expected URI object or URI string))
```

I'm not sure if this is a scenario you need to check for, and perhaps handle more gracefully within the UrlParser class, or is just unnecessary.

app/controllers/users_controller.rb
```diff
@@ -49,4 +49,25 @@ def sign_in_user
...
+
+    redirect_url = url_parser.get_redirect_path(paths_to_ignore: [save_your_results_path])
+    user_session.registration_triggered_path = redirect_url
+  end
+
+  def build_redirect_url
+    return unless request.referer
```
smoothcontract: Same here as above

app/lib/url_parser.rb
```diff
@@ -0,0 +1,32 @@
+class UrlParser
+  attr_reader :uri, :host
```
smoothcontract: Leave a blank line here pls :)

app/controllers/users_controller.rb
tavy88: Yep! I like that.

app/controllers/users_controller.rb
tavy88: Yep.

app/controllers/users_controller.rb
smoothcontract: If it's a valid scenario to have a `nil` referer, maybe add a test for this too?

app/controllers/users_controller.rb
```diff
@@ -49,4 +49,25 @@ def sign_in_user
     passwordless_session = build_passwordless_session(@user)
     @user.register_existing_user(passwordless_session, request.base_url)
   end
+
+  def set_redirect_path_for_registration
+    return unless request.referer
+
+    redirect_url = url_parser.get_redirect_path(paths_to_ignore: [save_your_results_path])
+    user_session.registration_triggered_path = redirect_url
```
tavy88: 👍

app/lib/url_parser.rb
```diff
@@ -0,0 +1,32 @@
...
+  def initialize(referer, host)
+    @uri = URI(referer)
```
tavy88: Could the referer ever be nil here?

spec/lib/url_parser_spec.rb
```diff
@@ -0,0 +1,89 @@
+require 'rails_helper'
+
+RSpec.describe UrlParser do
```
tavy88: Again I think it's worth adding a step for when the `referal` is potentially `nil`?

app/controllers/users_controller.rb
smoothcontract: Discussed offline - would be nice to move this check inside `UrlParser` class if possible.

app/lib/url_parser.rb
```diff
@@ -0,0 +1,35 @@
+class UrlParser
+  attr_reader :uri, :host
+
+  def initialize(referer, host)
+    @uri = URI(referer) if referer
```
tavy88: 👍

spec/lib/url_parser_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      referer = 'http://myhost.com/save-my-results?some-query'
+      parser = described_class.new(referer, 'myhost.com')
+
+      expect(parser.get_redirect_path(paths_to_ignore: ['/save-my-results'])). to be_nil
+    end
+
+    it 'returns nothing if no url passed' do
+      parser = described_class.new(nil, 'myhost.com')
+
+      expect(parser.get_redirect_path(paths_to_ignore: ['/save-my-results'])). to be_nil
```
tavy88: Is this working? I can see a space between `. to be_nil`.

PR #182, opened 2019-09-16
Filter out sessions when specific endpoint is hit

tavy88:
### Update

Visit `/status_check` and you should not get a session.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,8 +1,6 @@
 class ApplicationController < ActionController::Base
   include Passwordless::ControllerHelpers
 
-  before_action :set_request_path
-
   def user_session
     @user_session ||= UserSession.new(session)
   end
@@ -25,9 +23,5 @@ class ApplicationController < ActionController::Base
 
   def protect_feature(feature)
     redirect_to task_list_path unless Flipflop.enabled?(feature)
-  end
-
-  def set_request_path
-    user_session.request_path = request.path
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,8 @@
 class ApplicationController < ActionController::Base
   include Passwordless::ControllerHelpers
 
+  before_action :set_request_path
+
   def user_session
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -24,4 +26,8 @@ def track_event(event_key, properties = {})
   def protect_feature(feature)
     redirect_to task_list_path unless Flipflop.enabled?(feature)
   end
+
+  def set_request_path
+    user_session.request_path = request.path
+  end
```
smoothcontract: I'm not very keen on this approach. This means that every request (apart from the health endpoint) to the site is going to require a database write to update this key on the session. That feels like a bad idea when trying to build a scalable site. We should only be updating the session content where necessary and when the user undertakes a specific action we care about.

app/models/session.rb
```diff
@@ -1,3 +1,13 @@
 class Session < RestrictedActiveRecordBase
   has_one :user
+
+  validate :health_checkpoint
+
+  private
+
+  def health_checkpoint
+    return if data.present? && data['request_path'].exclude?('/status_check')
```
smoothcontract: Avoiding saving the session when it's empty is an interesting idea; there may be some mileage in that but would be good if we can find an alternate technique to raising validation errors.

smoothcontract: Perhaps look at overriding `save` method or adding a `before_save` callback instead.. 🤔

app/controllers/status_check_controller.rb
```diff
@@ -0,0 +1,7 @@
+class StatusCheckController < ApplicationController
+  def index
+    request.session_options[:skip] = true
```
cwrw: are you sure this works? I checked the issues on active record session and it seems that this option was removed?

cwrw: https://github.com/rails/activerecord-session_store/issues/122

tavy88: I've tested it yes. Feel free to test as well.

tavy88:
This is no longer needed.

PR #189, opened 2019-09-18
GET-420 Add new page for registration resend email

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-420

Add new page when user clicks resend email on registration. This will continue on to where they clicked register, and show their email.

Also we needed to change the breadcrumbs for the previous confirmation page on registration as well as the title of the page

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -17,6 +17,15 @@ class UsersController < ApplicationController
     render(:registration_results_saved)
   end
 
+  def registration_send_email_again
+    @user = User.find_or_initialize_by(email: user_params[:email])
+    register_user
+    render(:registration_email_sent_again)
+  rescue NotifyService::NotifyAPIError
+    # TODO: show user an error page, for now render email sent again page
+    render(:registration_email_sent_again)
+  end
+
   def sign_in
     @user = User.find_or_initialize_by(email: user_params[:email])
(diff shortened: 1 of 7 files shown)


app/controllers/users_controller.rb
```diff
@@ -8,13 +8,22 @@ def create
...
+    render(:registration_results_saved)
+  end
+
+  def registration_send_email_again
+    @user = User.find_or_initialize_by(email: user_params[:email])
+    register_user
+    render(:registration_email_sent_again)
+  rescue NotifyService::NotifyAPIError
+    # TODO: show user an error page, for now render email sent again page
+    render(:registration_email_sent_again)
```
smoothcontract: Couple of things here perhaps. Firstly, why is the email validation skipped here? If we don't skip that then the body of this method looks almost identical to the one above so potential to DRY this up and just render the appropriate view I think.

smoothcontract: Discussed in person, the two methods are slightly different, and email validation is not required so not as easy to DRY as it first seems.

cwrw: talked in person and will clean this up in next ticket

cwrw:
closing in favour of https://github.com/DFE-Digital/get-help-to-retrain/pull/191 which has both changes

PR #191, opened 2019-09-19
GET-494 Return to results resend email

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-494

Add new path and page for resend link on user sign in

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -39,6 +39,15 @@ class UsersController < ApplicationController
     redirect_to(link_sent_path(email: @user.email))
   end
 
+  def sign_in_send_email_again
+    @user = User.find_or_initialize_by(email: user_params[:email])
+    sign_in_user unless @user.new_record?
+    render(:sign_in_email_sent_again)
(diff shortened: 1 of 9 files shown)


app/controllers/users_controller.rb
```diff
@@ -57,12 +47,40 @@ def set_redirect_path_for_registration
...
       anchor: 'sign-in'
     )
   end
 
   def url_parser
     @url_parser ||= UrlParser.new(request.referer, request.host)
   end
+
+  def sign_in_with(path: nil, partial: nil) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+    if user.valid?
```
tavy88: I think we should tackle this rather than disable it and push for refactoring...

cwrw: This was only one point over because of adding `path ? redirect_to(path) : render(partial)` in the rescue, which will be tackled in a later ticket in this sprint. Otherwise I think this method is pretty straightforward otherwise

PR #192, opened 2019-09-19
Filter sensitive data in logs

tavy88:
This PR is doing the following:

1. Ensures form field values are filtered in the logs so there
is no risk potential for leaking sensitive data.
2. Changes the log level to :INFO in PRD so the SQL statements
are not visible anymore as they carry sensitive information.

### Screenshots

Screen Shot 2019-09-19 at 10 11 32
Screen Shot 2019-09-19 at 10 42 03
Screen Shot 2019-09-19 at 10 52 57
Screen Shot 2019-09-19 at 11 10 28
Screen Shot 2019-09-19 at 11 12 03
Screen Shot 2019-09-19 at 14 33 54
Screen Shot 2019-09-19 at 15 51 04

### IMPORTANT

Needs checking Loggly as well.

### Ticket

https://dfedigital.atlassian.net/browse/GET-475


config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,14 @@
...
+Rails.application.config.filter_parameters += %i[
+  password
+  first_name
+  last_name
+  gender
+  birth_day
+  birth_month
+  birth_year
+  postcode
+  email
```
cwrw: do we not also want passwordless session token as well?

tavy88: Is that a form field?

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,15 @@
...
+  authenticity_token
```
cwrw: isn't it just called token?

tavy88: Please check the attached screenshots.

config/initializers/filter_parameter_logging.rb
cwrw: no its when we create passwordless sessions for users when they sign in

config/initializers/filter_parameter_logging.rb
cwrw: for clarity we're talking about two different things, the csrf token (the one in the screenshots called `authenticity_token`) and the passwordless token (called `token`)

tavy88: Yeah so filtering the token is not enough as per Slack conversation
Screen Shot 2019-09-19 at 11 34 47
You can still see the token on the GET request.

tavy88: Still looking for a way to hide potentially certain GET requests.

smoothcontract: This might not be that big of a deal - the tokens are single use anyway so even if someone has access to historic tokens they wouldn't be able to be used to steal user data. If you don't find an easy solution to this then I wouldn't worry about it.

app/models/custom_formatter.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    super
+  end
+
+  private
+
+  def hide_token(timestamp)
+    "Started GET \"/sign-in/[FILTERED]\" at #{timestamp}"
+  end
+end
```
tavy88: I thought about speccing but it's usually not recommended to spec loggers. Not seeing a lot of value in it here.

smoothcontract: It's an innovative idea but I don't think we need to do this. The passwordless tokens can't be used more than once so don't present a security issue. I'd prefer to keep this clean and just filter the posted params as per the original PR implementation.

smoothcontract: Also the key focus for this feature is to filter submitted personal information from the logs - which is achieved by extending the list of filtered params. I'd suggest reverting this logging change.

tavy88: Are you sure it's not going to fail the security check?

PR #193, opened 2019-09-19
Add health check endpoint

smoothcontract:
### Context
Health endpoint reports status of dependencies (Data and API endpoints) and meta data.
https://dfedigital.atlassian.net/browse/GET-340

This adds new Healthcheck service and check classes for key models and required integration points. The endpoint also includes select metadata - on Azure we'll need a pipeline change to populate `GIT_SHA` ENV var that can be read in `application.rb`.

This also adds a new health_check method to Notify service; that simulates sending a health check template to a smoke test email address. This exercises the Notify service without sending an actual email. If this fails a warn state is returned as we can still use the service otherwise.

For postcodes.io a geocoding call to a postcode is made - that should return lat/long, if not then a warn state is returned as we can still skip past the postcode eligibility page.

For split.io a new health_check feature flag has been added - this is turned on in staging and prod environments so should always return true. If not then a fail status is returned as the site may not work as expected with all features disabled.

Reference data checks for courses and job profiles are made simply by counting records. If these are not present a fail status is returned. Session count is also included to ensure that the restricted database is available.

If all checks pass then the overall result is a pass with a 200 response returned. If any checks fail the overall result is a fail with a 5XX response. Otherwise a warn result (partially degraded service) is returned with a 2XX response. I'm open to other suggestions for response codes from the list here (but also see notes in the health check standard above). https://apidock.com/rails/ActionController/Base/render#254-List-of-status-codes-and-their-symbols

### Guidance to review
The health endpoint is available at `/status` - that can be checked on the review app.

This returns a JSON structure so I'd recommend installing a JSON formatting browser extension. I'm using this one with Chrome: https://chrome.google.com/webstore/search/json%20formatter

The `GIT_SHA` environment variable isn't working yet - a problem with setting this in the docker image. I'm hoping @mafonso can give a hand with that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM ruby:2.6.4-alpine as assets
 
+ARG GIT_SHA="unknown"
+ENV GIT_SHA="${GIT_SHA}"
+
 ENV RAILS_ENV production
 ENV NODE_ENV production
 ENV BUILD_PACKAGES build-base nodejs yarn tzdata postgresql-dev git
@@ -30,6 +33,8 @@ ENV RAILS_SERVE_STATIC_FILES true
 ENV RAILS_LOG_TO_STDOUT true
 ENV NODE_ENV production
 ENV BUILD_PACKAGES nodejs yarn tzdata libpq
+
+RUN export GIT_SHA=$(git rev-parse --short HEAD)
 
 RUN apk add --update $BUILD_PACKAGES && \
     rm -rf /var/cache/apk/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,5 @@
 FROM ruby:2.6.4-alpine as assets
 
-ARG GIT_SHA="unknown"
-ENV GIT_SHA="${GIT_SHA}"
-
 ENV RAILS_ENV production
 ENV NODE_ENV production
 ENV BUILD_PACKAGES build-base nodejs yarn tzdata postgresql-dev git
@@ -34,7 +31,8 @@ ENV RAILS_LOG_TO_STDOUT true
 ENV NODE_ENV production
 ENV BUILD_PACKAGES nodejs yarn tzdata libpq
 
-RUN export GIT_SHA=$(git rev-parse --short HEAD)
+ARG GIT_SHA="unknown"
+ENV GIT_SHA=${GIT_SHA}
 
 RUN apk add --update $BUILD_PACKAGES && \
     rm -rf /var/cache/apk/*
--- a/app/services/health_check/check_base.rb
+++ b/app/services/health_check/check_base.rb
@@ -1,0 +1,28 @@
+module HealthCheck
+  class CheckBase
+    attr_reader :output
+
+    def name
+      raise NotImplementedError
+    end
+
+    def status
+      raise NotImplementedError
+    end
+
+    def time
+      Time.now.iso8601
+    end
+
+    def detail
+      {
+        status: status,
+        time: time
+      }.tap do |details|
+        details[:metricValue] = value if respond_to?(:value)
+        details[:metricUnit] = unit if respond_to?(:unit)
+        details[:output] = output if output.present?
+      end
+    end
+  end
+end
--- a/app/services/health_check/features_check.rb
+++ b/app/services/health_check/features_check.rb
@@ -5,7 +5,6 @@ module HealthCheck
     end
 
     def value
-      @output = Flipflop::FeatureSet.current.features.map(&:as_json)
       @value ||= Flipflop.health_check?
     rescue StandardError => e
       @output = "SplitIO error: #{e.inspect}"
--- a/app/services/health_check/features_check.rb
+++ b/app/services/health_check/features_check.rb
@@ -8,11 +8,11 @@ module HealthCheck
       @value ||= Flipflop.health_check?
     rescue StandardError => e
       @output = "SplitIO error: #{e.inspect}"
-      nil
+      false
     end
 
     def unit
-      'JSON'
+      'Boolean'
     end
 
     def status
--- a/app/services/health_check/features_check.rb
+++ b/app/services/health_check/features_check.rb
@@ -1,0 +1,23 @@
+module HealthCheck
+  class FeaturesCheck < CheckBase
+    def name
+      'api:split.io'
+    end
+
+    def value
+      @output = Flipflop::FeatureSet.current.features.map(&:as_json)
+      @value ||= Flipflop.health_check?
+    rescue StandardError => e
+      @output = "SplitIO error: #{e.inspect}"
+      nil
+    end
+
+    def unit
+      'JSON'
+    end
+
+    def status
+      value ? :pass : :fail
+    end
+  end
+end
--- a/app/services/health_check/notify_check.rb
+++ b/app/services/health_check/notify_check.rb
@@ -5,14 +5,14 @@ module HealthCheck
     end
 
     def value
-      @value ||= NotifyService.new.health_check.id
-    rescue Notifications::Client::RequestError, RuntimeError => e
+      @value ||= NotifyService.new.health_check
+    rescue Notifications::Client::RequestError, RuntimeError, ArgumentError => e
       @output = "Notify API error: #{e.message}"
       nil
     end
 
     def unit
-      'GUID'
+      'UUID'
     end
 
     def status
--- a/app/services/health_check/notify_check.rb
+++ b/app/services/health_check/notify_check.rb
@@ -1,0 +1,22 @@
+module HealthCheck
+  class NotifyCheck < CheckBase
+    def name
+      'api:notifications.service.gov.uk'
+    end
+
+    def value
+      @value ||= NotifyService.new.health_check.id
+    rescue Notifications::Client::RequestError, RuntimeError => e
+      @output = "Notify API error: #{e.message}"
+      nil
+    end
+
+    def unit
+      'GUID'
+    end
+
+    def status
+      value.present? ? :pass : :warn
+    end
+  end
+end
--- a/app/services/health_check/postcodes_check.rb
+++ b/app/services/health_check/postcodes_check.rb
@@ -8,7 +8,7 @@ module HealthCheck
       @value ||= Geocoder.coordinates('B1 2JP')
     rescue SocketError, Timeout::Error, Geocoder::Error => e
       @output = "Geocoder API error: #{e.message}"
-      nil
+      []
     end
 
     def unit
--- a/app/services/health_check/postcodes_check.rb
+++ b/app/services/health_check/postcodes_check.rb
@@ -1,0 +1,22 @@
+module HealthCheck
+  class PostcodesCheck < CheckBase
+    def name
+      'api:postcodes.io'
+    end
+
+    def value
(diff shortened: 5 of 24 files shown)


config/features.rb
```diff
@@ -4,6 +4,7 @@
   strategy :split_client, api_key: ENV['SPLIT_API_KEY'] if ENV['SPLIT_API_KEY'].present?
 
   feature :foo, description: 'Example feature flag', default: true
+  feature :health_check, description: 'Dummy feature for split.io health check'
```
cwrw: can we use foo instead?

app/services/health_check/report_service.rb
```diff
@@ -0,0 +1,43 @@
...
+        preload_check_classes
+        CheckBase.descendants.map(&:new)
+      end
+    end
+
+    def preload_check_classes
+      # Zeitwerk handles eager loading all classes in production mode
+      return if Rails.env.production?
+
+      Dir.glob(Rails.root.join('app', 'services', 'health_check', '*.rb')) { |f| require f }
```
cwrw: Im not sure about this, can't we load this in a different place than the class itself for envs other than production?

cwrw:
really good clean implementation, will wait till the final copy is done for a review, but I really like the approach

app/services/health_check/report_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def status
+      return :pass if statuses.all?(:pass)
+      return :fail if statuses.any?(:fail)
+
+      :warn
+    end
+
+    def details
+      checks.map { |check| [check.name, [check.detail]] }.to_h
```
tavy88: Not sure entirely here what is the structure you are trying to have in the end. Can you give an example?

app/services/health_check/check_base.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def detail
+      {
+        status: status,
+        time: time
+      }.tap do |details|
+        details[:metricValue] = value if respond_to?(:value)
+        details[:metricUnit] = unit if respond_to?(:unit)
+        details[:output] = output if output.present?
+      end
```
tavy88: Is this the Azure expected format?

config/features.rb
smoothcontract: `foo` is already defaulted to true in the features file, so returns true even if we can't read from split.io. I added this explicit feature with a default value of false, and is only enabled via split.io so we can be sure the API call is working.

app/services/health_check/report_service.rb
smoothcontract: have a look at the /status endpoint :)

app/services/health_check/check_base.rb
smoothcontract: There's not really an expected Azure format. So this adopts the standard @mafonso suggested in the Jira ticket: https://tools.ietf.org/html/draft-inadarei-api-health-check-01

smoothcontract:
> Just a few questions. Also, should we put SmartSurvey on the radar? And the PID/Sessios stuff DB connection checked?

I'm not sure how we would test SmartSurvey easily; it wasn't originally in scope for the service status as it's not something we integrate with but just an external site link. If it was down user would see a 404 on the SmartSurvey site which would be okay I think - and doesn't mean our service is particularly affected.

I've added the session count which covers off access to the restricted database - it's not essential to check that every model is populated so one table from each database should suffice. However courses and job profiles are populated via different sets of rake tasks, and are needed for the service so if something went wrong with one of those tasks then this would flag it (e.g. the courses import task first deletes all existing courses). For the restricted database the session count is a good choice as it's also something we want to keep an eye on easily.

app/services/health_check/report_service.rb
tavy88: Yeah so my confusion was: 
```ruby

  "status": "warn",
  "version": null,
  "details": {
    "api:notifications.service.gov.uk": [
      {
        "status": "warn",
        "time": "2019-09-19T15:57:37+00:00",
        "metricValue": null,
        "metricUnit": "GUID",
        "output": "Notify API error: BadRequestError: Can’t send to this recipient using a team-only API key"
      }
    ],
    "api:postcodes.io": [
      {
        "status": "pass",
        "time": "2019-09-19T15:57:37+00:00",
        "metricValue": [
          52.476692,
          -1.910337
        ],
        "metricUnit": "Array"
      }
    ],
```
Why does every service endpoint have to be an array and not just a Hash e.g:

Why can't:
```ruby
"api:postcodes.io": [
      {
        "status": "pass",
        "time": "2019-09-19T15:57:37+00:00",
        "metricValue": [
          52.476692,
          -1.910337
        ],
        "metricUnit": "Array"
      }
    ]
```

Just be:

```ruby
"api:postcodes.io": {
        "status": "pass",
        "time": "2019-09-19T15:57:37+00:00",
        "metricValue": [
          52.476692,
          -1.910337
        ],
        "metricUnit": "Array"
      }
```

smoothcontract: Because the standard I followed says:

> Since each sub-component may be backed by several nodes with varying health statuses, the key points to an array of objects.  In case of a single-node sub-component (or if  presence of nodes is not relevant), a single-element array should be used as the value, for consistency.

app/services/health_check/notify_check.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    def value
+      @value ||= NotifyService.new.health_check
+    rescue Notifications::Client::RequestError, RuntimeError, ArgumentError => e
+      @output = "Notify API error: #{e.message}"
+      nil
+    end
+
+    def unit
+      'GUID'
```
cwrw: Im not sure about this, should it be a string instead?

app/services/health_check/features_check.rb
```diff
@@ -0,0 +1,22 @@
...
+  class FeaturesCheck < CheckBase
+    def name
+      'api:split.io'
+    end
+
+    def value
+      @value ||= Flipflop.health_check?
+    rescue StandardError => e
+      @output = "SplitIO error: #{e.inspect}"
+      nil
```
cwrw: should this be false to match the unit?

app/services/health_check/postcodes_check.rb
```diff
@@ -0,0 +1,22 @@
...
+  class PostcodesCheck < CheckBase
+    def name
+      'api:postcodes.io'
+    end
+
+    def value
+      @value ||= Geocoder.coordinates('B1 2JP')
+    rescue SocketError, Timeout::Error, Geocoder::Error => e
+      @output = "Geocoder API error: #{e.message}"
+      nil
```
cwrw: should this be an empty array to match the unit?

smoothcontract: No, I think `nil` is the correct return value here

cwrw: could you explain why that is?

app/services/health_check/notify_check.rb
smoothcontract: The return value is a GUID, although wrapped as a String. I've tried to use the most informative return type here so GUID seems reasonable - the suggestion in the standard is that these values should be:

> A common and standard term from a well-known source such as schema.org, IANA or microformats.

I think GUID is more informative, although maybe `UUID` would be better. https://schema.org/identifier

app/services/health_check/report_service.rb
tavy88: Ok cool.

Dockerfile
```diff
@@ -31,6 +34,8 @@ ENV RAILS_LOG_TO_STDOUT true
 ENV NODE_ENV production
 ENV BUILD_PACKAGES nodejs yarn tzdata libpq
 
+RUN export GIT_SHA=$(git rev-parse --short HEAD)
```
mafonso: Does this work? I don't think you can use a runtime command to persist something into the image.  This won't be triggered at runtime.

mafonso: You could use `ENV var=$ARG_VALUE` as this will create a persistent layer with the value as it were at build time.

app/services/notify_service.rb
```diff
@@ -20,6 +22,13 @@ def send_email(email_address:, template_id:, options: {})
     raise NotifyAPIError
   end
 
+  def health_check
+    client.send_email(
+      email_address: HEALTH_CHECK_EMAIL,
+      template_id: HEALTH_CHECK_TEMPLATE_ID
+    ).id
+  end
```
mafonso: Is it possible to check if the API is working without actually sending an email? We are not going to test reception of email so it's never a complete test.  I think it's safe to assume that if the API is responsive the email sending will work. 
Maybe the API has an health check of it's own that we can piggy back on?

config/features.rb
mafonso: My preference would be cascading on their own health check, if there is one. Relying on a a dummy feature status could render the whole service inoperative if we toggle that feature by mistake or if it defaults to the wrong thing in an split.io outage.

app/services/health_check/check_base.rb
mafonso: Is it too much work make so that the main check is a "ms" instead of a "boolean" for the pass status? 
That way we know in a single test if it passed and how long are the requests taking (we could plot this?) 

Because a DB call could be passing but taking seconds instead of ms and that would be a problem. That also raises the issue if that should become a "warning" status rather than a pass, but we could leave that for later.

config/features.rb
smoothcontract: Split.io don't have anything like that, so this is as good as we can do really. It does exercise the split.io API - and if we did change the feature setting either on split.io (to off) then healthcheck would fail and we'd know. If someone changed the default in the features file the unit tests would fail and we'd know. So I think we're good here.

app/services/health_check/check_base.rb
smoothcontract: That's a good idea - essentially extending this to check performance rather than just presence of data. But I'm not sure it's within the intended scope of the health check endpoint. Multiple metrics for each check are supported by the standard but would mean some refactoring to include these - perhaps we can defer that for now and cover in a future story. We already have other ways to check the response time of the site don't we?

app/services/health_check/notify_check.rb
smoothcontract: Changed to UUID.

app/services/health_check/postcodes_check.rb
smoothcontract: Discussed offline - changed to empty array for consistency with the published data type (this is also how geocoding failures get persisted).

Dockerfile
smoothcontract: Nope, it doesn't work. We've been back and forth on this, think we should revert to the technique of passing in a build argument rather than running git commands in the container. But.. I'll need some assistance with that to determine the best value to pass in from the Azure pipeline @mafonso

app/services/notify_service.rb
smoothcontract: This is essentially a health check. The email address is a magic value that causes the service to pretend to send an email and return a sent message id without actually sending an email. There's no other endpoint that would be suitable for a health check.

https://docs.notifications.service.gov.uk/ruby.html#ruby-client-documentation
https://docs.notifications.service.gov.uk/ruby.html#smoke-testing

> The smoke test phone numbers and email addresses will validate the request and simulate a successful response, but won’t send a real message, produce a delivery receipt or persist the notification to the database.

PR #194, opened 2019-09-20
Get 425

mafonso:
### Context
The pushed production parameters file was not respecting the naming convention

This PR fixes that.


mafonso:
> can you remove the other `prod.params` file (that's missing `.json`)?

That explains it. I did a `git mv old new` but for some reason the PR didn't look the way I expected.  I think it's because I did a rebase from master before pushing it and the old file came along.
Yeah.. removing it.

PR #199, opened 2019-09-23
GET-423 Skills Builder v3

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-423

* Fix skills builder bugs and add new content
* Add ability to remove job profile from skills page

### DIFF
--- a/app/controllers/job_profiles_controller.rb
+++ b/app/controllers/job_profiles_controller.rb
@@ -6,9 +6,6 @@ class JobProfilesController < ApplicationController
   def destroy
     user_session.remove_job_profile(resource.id)
 
-    query_params = { search: job_profile_params[:search] }
-    query_params[:job_profile_id] = job_profile_params[:job_profile_id] unless user_session.job_profile_skills?
-
     redirect_to(
       skills_path(query_params)
     )
@@ -22,5 +19,13 @@ class JobProfilesController < ApplicationController
 
   def job_profile_params
     params.permit(:job_profile_id, :search, :id)
+  end
+
+  def query_params
+    # TODO: this is to support skills removal. Revisit after removing skills builder v2 flag and fixing navigation
+    query = { search: job_profile_params[:search] }
+    query[:job_profile_id] = job_profile_params[:job_profile_id] unless user_session.job_profile_skills?
+
+    query
   end
 end
--- a/app/controllers/job_profiles_controller.rb
+++ b/app/controllers/job_profiles_controller.rb
@@ -3,9 +3,24 @@ class JobProfilesController < ApplicationController
     @job_profile = JobProfileDecorator.new(resource)
   end
 
+  def destroy
+    user_session.remove_job_profile(resource.id)
+
+    query_params = { search: job_profile_params[:search] }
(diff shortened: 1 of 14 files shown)


app/controllers/job_profiles_controller.rb
```diff
@@ -3,9 +3,24 @@ def show
     @job_profile = JobProfileDecorator.new(resource)
   end
 
+  def destroy
+    user_session.remove_job_profile(resource.id)
+
+    query_params = { search: job_profile_params[:search] }
+    query_params[:job_profile_id] = job_profile_params[:job_profile_id] unless user_session.job_profile_skills?
```
smoothcontract: I don't understand why this has an unless clause - is that for backwards compatibility? Might be worth adding a comment to explain this, perhaps with a TODO if it's something that needs to be cleaned up when we roll out the new skills builder feature.

app/controllers/skills_controller.rb
```diff
@@ -39,7 +39,9 @@ def job_profiles
   end
 
   def show_skills_builder_v2
-    return redirect_to task_list_path unless job_profiles.any? && user_session.skill_ids.any?
+    return redirect_to task_list_path unless
+      skills_params[:job_profile_id] ||
```
tavy88: Do you want

```ruby
skills_params[:job_profile_id].present?
```

here?

app/models/skills_builder.rb
```diff
@@ -11,15 +11,18 @@ def initialize(skills_params:, job_profile:, user_session:)
...
-    return unless skills_params.present?
+    return unless skills_params.present? && valid?
 
     user_session.set_skills_ids_for_profile(job_profile.id, formatted_skill_params)
   end
 
   def skill_ids
-    @skill_ids ||=
+    @skill_ids ||= begin
+      return [] if skills_params.present? && !valid?
```
tavy88: I think this can live outside the `begin` block. What do you think?

cwrw: I think it's better to be in the begin block to be memoized as we call this method quite alot

app/controllers/job_profiles_controller.rb
```diff
@@ -3,9 +3,29 @@ def show
     @job_profile = JobProfileDecorator.new(resource)
   end
 
+  def destroy
+    user_session.remove_job_profile(resource.id)
```
tavy88: Nice!

app/models/skills_builder.rb
tavy88: Ok fair.

spec/models/user_sessions_spec.rb
```diff
@@ -246,32 +280,20 @@
 
   describe '#job_profile_skills?' do
     it 'returns true if the session already contains job profile skills' do
-      session[:job_profile_skills] = { '1' => [1] }
+      session[:job_profile_ids] = [1]
```
tavy88: Assuming session[:job_profile_ids] is an array you can do 

```ruby
session[:job_profile_ids] << 1
```

here?

cwrw: I chose to explicitly do this as this means the job profile ids array only equals 1 as opposed to appending it to any existing array with values

PR #201, opened 2019-09-24
Get 384 tracking refactor

smoothcontract:
### Context
Refactor Azure Application Insights implementation to support better testing

This adds a new TrackingService class for custom event tracking so that we can set expectations on tracking data within associated feature specs. Moved AAI configuration to centralised Rails config object and cleaned up the existing initialiser code so that only the client side JS tracking and exception tracking are created on startup.

Custom tracking events lazily load Azure tracking telemetry client and cached at class level.

Also added feature specs for custom tracking to ensure this gets called with the correct payload.

### Guidance to review
This shouldn't affect any existing tracking, but just means we're able to add test coverage. I noticed whilst working on this that we don't have any custom event tracking for skills builder v2; will need to discuss with wider team to see if this is still a requirement (given we can now determine selected job profiles and skills from session data instead).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,9 +16,8 @@ class ApplicationController < ActionController::Base
   private
 
   def track_event(event_key, properties = {})
-    return unless defined?(TELEMETRY_CLIENT)
-
-    TELEMETRY_CLIENT.track_event(I18n.t(event_key, scope: :events), properties: properties)
+    event = I18n.t(event_key, scope: :events)
+    properties.present? ? TrackingService.track_event(event, properties) : TrackingService.track_event(event)
   end
 
   def protect_feature(feature)
--- a/app/services/tracking_service.rb
+++ b/app/services/tracking_service.rb
@@ -1,0 +1,45 @@
+class TrackingService
+  def self.track_event(event_text, properties = {})
+    return false unless key.present?
+
+    client.track_event(event_text, properties: properties)
+  end
+
+  def self.key
+    @key ||= Rails.configuration.app_insights_instrumentation_key
+  end
+
+  def self.key=(key)
+    @key = key
+    @sender = nil
+    @queue = nil
+    @channel = nil
+    @client = nil
+  end
+
+  def self.sender
+    @sender ||= ApplicationInsights::Channel::AsynchronousSender.new
+  end
+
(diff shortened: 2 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -16,9 +16,8 @@ def current_user
   private
 
   def track_event(event_key, properties = {})
-    return unless defined?(TELEMETRY_CLIENT)
-
-    TELEMETRY_CLIENT.track_event(I18n.t(event_key, scope: :events), properties: properties)
+    event = I18n.t(event_key, scope: :events)
+    properties.present? ? TrackingService.track_event(event, properties) : TrackingService.track_event(event)
```
smoothcontract: This may seem like an unnecessary test, but it means that expectations for checking this was called correctly won't need to specify an empty properties hash if no properties were passed (e.g. when skilling personal info page).

spec/features/check_your_skills_spec.rb
```diff
@@ -28,6 +28,16 @@
     expect(page).to have_text('No results found')
   end
 
+  scenario 'tracks search string' do
```
smoothcontract: Open to better suggestions for naming these scenarios

app/services/tracking_service.rb
```diff
@@ -0,0 +1,46 @@
...
+    @queue ||= ApplicationInsights::Channel::AsynchronousQueue.new(sender)
+  end
+
+  def self.channel
+    @channel ||= ApplicationInsights::Channel::TelemetryChannel.new(nil, queue)
+  end
+
+  def self.client
+    @client ||= ApplicationInsights::TelemetryClient.new(key, channel).tap do |tc|
+      # binding.pry
```
cwrw: 🦂

config/initializers/application_insights.rb
```diff
@@ -1,30 +1,11 @@
-if Rails.env.production?
```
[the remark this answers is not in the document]
mafonso: Makes sense. That was the way I had it initially, but I was suggested to scope it to production only. But I think it's best to let the definition (or lack of) of the ENV variable to determine it.

app/services/tracking_service.rb
smoothcontract: Damnabbit 😠

PR #204, opened 2019-09-24
Add growth score explanations

tavy88:
### Context

Add accordion component for growth scores

Add explanation for each growth type using the accordion
GDS component.

### Screenshots
Screen Shot 2019-09-24 at 16 55 54
Screen Shot 2019-09-24 at 16 56 01

https://dfedigital.atlassian.net/browse/GET-489

### DIFF
--- a/app/decorators/job_profile_decorator.rb
+++ b/app/decorators/job_profile_decorator.rb
@@ -78,6 +78,17 @@ class JobProfileDecorator < SimpleDelegator # rubocop:disable Metrics/ClassLengt
 
     'Growing strongly'
   end
+
+  def growth_explanation
+    return unless growth
+
+    case growth
+    when (-Float::INFINITY...-5) then I18n.t(:falling, scope: :job_growth_explanation)
+    when -5...5 then I18n.t(:stable, scope: :job_growth_explanation)
+    when 5...50 then I18n.t(:growing, scope: :job_growth_explanation)
+    else I18n.t(:growing_strongly, scope: :job_growth_explanation)
+    end
+  end
   # rubocop:enable Metrics/CyclomaticComplexity
 
   def skills_match(score) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
--- a/app/views/skills_matcher/_job_matches.html.erb
+++ b/app/views/skills_matcher/_job_matches.html.erb
@@ -1,7 +1,7 @@
 <ul class="govuk-list">
   <% job_profiles.each do |job_profile| %>
     <li class="govuk-!-padding-bottom-1">
-      <h3 class="govuk-!-margin-bottom-0">
+      <h3 class="govuk-!-margin-bottom-0 govuk-!-margin-top-2">
          <%= link_to job_profile.name, job_profile_path(job_profile.slug, search: params[:search]), class: 'govuk-link no-text-decoration' %>
       </h3>
       <% if job_profile.alternative_titles.present? %>
@@ -12,6 +12,16 @@
       <p class="govuk-body-s govuk-!-margin-bottom-1">Skills match: <span class="strong"><%= job_profile.skills_match(@scores[job_profile.id]) %></span></p>
       <% if job_profile.growth.present? %>
         <p class="govuk-body-s govuk-!-margin-bottom-1">Recent job growth: <span class="strong"><%= job_profile.growth_type %></span></p>
+        <details class="govuk-details govuk-details-custom" data-module="govuk-details">
+          <summary class="govuk-details__summary govuk-details__summary-custom">
+            <span class="underlined-text">
+              What does this mean?
+            </span>
+          </summary>
+          <div class="govuk-details__text govuk-body-m">
+            <%= job_profile.growth_explanation %>
+          </div>
+        </details>
       <% end %>
     </li>
   <% end %>
--- a/spec/features/skills_matcher_spec.rb
+++ b/spec/features/skills_matcher_spec.rb
@@ -49,6 +49,28 @@ RSpec.feature 'Skills matcher', type: :feature do
     expect(page).to have_current_path(task_list_path)
   end
 
+  scenario 'Each section from job matches has the What does this mean? link if there is a job growth information' do
+    visit_skills_for_current_job_profile
+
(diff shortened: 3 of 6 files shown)


app/decorators/job_profile_decorator.rb
```diff
@@ -20,6 +20,22 @@ class JobProfileDecorator < SimpleDelegator # rubocop:disable Metrics/ClassLengt
   OTHER_ROUTES_SECTION_XPATH = "//section[contains(@class, 'job-profile-subsection') and contains (@id, 'otherroutes')]".freeze
   # rubocop:enable Metrics/LineLength
 
+  GROWING_STRONGLY_EXPLANATION = 'In the last few years, employers have been
```
cwrw: could we move those to the translations file?

app/decorators/job_profile_decorator.rb
```diff
@@ -78,6 +94,16 @@ def growth_type
...
+
+  def growth_explanation
+    return unless growth
+
+    return FALLING_GROWTH_EXPLANATION if growth <= -5
+    return STABLE_GROWTH_EXPLANATION if growth > -5 && growth <= 5
+    return GROWING_GROWTH_EXPLANATION if growth > 5 && growth <= 50
+
+    GROWING_STRONGLY_EXPLANATION
+  end
```
smoothcontract: I think this would be a nice use case for a switch statement instead - and agree with removing the constants, those should be in the locale file instead. Something like this:

```ruby
def growth_explanation
  case growth
    when -10000...-5 then I18n.t(:falling, scope: :job_growth_explanation)
    when -5...5 then I18n.t(:stable, scope: :job_growth_explanation)
    when 5...50 then I18n.t(:growing, scope: :job_growth_explanation)
    when 50...10000 then I18n.t(:growing_strongly, scope: :job_growth_explanation)
  end
end
```

app/decorators/job_profile_decorator.rb
tavy88: I don't think it's a good idea as it is scoped just to this decorator. We are not going to reuse them anywhere else.

spec/features/skills_matcher_spec.rb
```diff
@@ -49,6 +49,28 @@ def visit_skills_for_current_job_profile
...
+  end
+
+  scenario 'When clicking on What does this mean? link user gets the explanation for that score' do
+    visit_skills_for_current_job_profile
+
+    create(:job_profile, :with_html_content, :growing, name: 'Cabin Crew', skills: [skill1, skill2])
+
+    visit '/job-matches'
+
+    find('.underlined-text').click
```
cwrw: this test passes without this step, could you add js: true so that is fails without this?

app/decorators/job_profile_decorator.rb
tavy88: May I ask why you feel strongly about moving these to locales? We are not going to reuse this anywhere else and its scope to this decorator. Are there strong arguments to do that?

spec/features/skills_matcher_spec.rb
tavy88: Sure.

app/decorators/job_profile_decorator.rb
smoothcontract: It's not a question of re-use. Text strings don't add understanding to the decorator and so shouldn't be defined within the decorator. The locale file is the place for that, even if the strings are only used once. The same reasoning applies to controllers for things like flash messages etc. - it looks cleaner.

As an aside, I came across this funky new syntax for ranges which you might also like to use: https://developer.squareup.com/blog/rubys-new-infinite-range-syntax-0/

app/decorators/job_profile_decorator.rb
cwrw: In the current project we have consistently maintained having our text in the translations file, I would like to keep the consistency with this as well, also it is a cleaner approach as these are blocks of text that are being treated as frozen constants, where in reality they aren't.  Instead we should have the text where we expect it to be.

app/views/skills_matcher/_job_matches.html.erb
```diff
@@ -12,6 +12,16 @@
       <p class="govuk-body-s govuk-!-margin-bottom-1">Skills match: <span class="strong"><%= job_profile.skills_match(@scores[job_profile.id]) %></span></p>
       <% if job_profile.growth.present? %>
         <p class="govuk-body-s govuk-!-margin-bottom-1">Recent job growth: <span class="strong"><%= job_profile.growth_type %></span></p>
+        <details class="govuk-details govuk-details-custom" data-module="govuk-details">
```
cwrw: I like the current approach, but out of curiosity, why didn't https://design-system.service.gov.uk/components/accordion work?

smoothcontract:
> can you remove the growth csv file as its wrongly commited?

@tavy88 looks like you added another commit which removes the file - could you instead amend the original commit which added the file? This is more secure as otherwise the commit with the file remains forever accessible to anyone with access to the repo (i.e. everyone). Thanks.

app/decorators/job_profile_decorator.rb
tavy88: Also re:switch statement instead how about these methods:

```ruby
def growth_icon
    return '' unless growth

    return 'arrow-falling-icon' if growth <= -5
    return 'arrow-stable-icon' if growth > -5 && growth <= 5
    return 'arrow-growing-icon' if growth > 5 && growth <= 50

    'arrow-growing-strongly-icon'
  end

  def growth_type
    return unless growth

    return 'Falling' if growth <= -5
    return 'Stable' if growth > -5 && growth <= 5
    return 'Growing' if growth > 5 && growth <= 50

    'Growing strongly'
  end
```

app/views/skills_matcher/_job_matches.html.erb
tavy88: https://design-system.service.gov.uk/components/details/

app/decorators/job_profile_decorator.rb
```diff
@@ -78,6 +78,17 @@ def growth_type
 
     'Growing strongly'
   end
+
+  def growth_explanation
+    return unless growth
+
+    case growth
+    when (-Float::INFINITY...-5) then I18n.t(:falling, scope: :job_growth_explanation)
```
smoothcontract: People definitely don't want to do a job with negative infinite growth 😀

PR #205, opened 2019-09-24
Sentry.io integration

smoothcontract:
### Context
Setup integration with Sentry.io to provide production exception tracking.

This requires new environment variable SENTRY_DSN to be set. If not present then Sentry is not activated. Also includes some additions to health endpoint - this now includes the hostname and environment name (dev / qa / uat etc.), which are also used for context in Sentry.

### Guidance to review
Ask me for sentry credentials. Relevant SENTRY_DSN has been setup on Heroku already, but also needs to be populated in Azure vault. Also need to think about how each environment will be defined so we can identify them in Sentry UI.

### DIFF
--- a/azure/config/ut.parameters.json
+++ b/azure/config/ut.parameters.json
@@ -40,6 +40,14 @@
         "secretName": "ut-splitApiKey"
       }
     },
+    "sentryDsn": {
+      "reference": {
+        "keyVault": {
+          "id": "/subscriptions/f35bd249-3028-48bd-9aab-d00e9d1a56b7/resourceGroups/s108d01-shared/providers/Microsoft.KeyVault/vaults/s108d01-shared-kv-01"
+        },
+        "secretName": "ut-sentryDsn"
+      }
+    },
     "notifyApiKey": {
       "reference": {
         "keyVault": {
--- a/azure/template.json
+++ b/azure/template.json
@@ -174,6 +174,7 @@
   "variables": {
     "deploymentUrlBase": "https://raw.githubusercontent.com/DFE-Digital/get-help-to-retrain-devops/master/templates/",
     "resourceNamePrefix": "[toLower(resourceGroup().name)]",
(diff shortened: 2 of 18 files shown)


spec/services/health_check/environment_check_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  describe '#status' do
+    context 'with no environment name configured' do
+      it 'returns fail' do
+        Rails.configuration.environment_name = nil
+        expect(check.status).to eq :fail
+      end
+    end
+
+    context 'with host name configured' do
```
cwrw: ✂️ host/environment

smoothcontract: Good spot!

azure/config/ut.parameters.json
```diff
@@ -40,6 +40,14 @@
         "secretName": "ut-splitApiKey"
       }
     },
+    "sentryDsn": {
+      "reference": {
+        "keyVault": {
+          "id": "/subscriptions/f35bd249-3028-48bd-9aab-d00e9d1a56b7/resourceGroups/s108d01-shared/providers/Microsoft.KeyVault/vaults/s108d01-shared-kv-01"
```
smoothcontract: I think this is the wrong vault actually. Oops.

azure/template.json
```diff
@@ -118,6 +118,13 @@
         "description:": "User feedback in page Smart Survey URL"
       }
     },
+    "sentryDsn": {
+      "type": "string",
```
mafonso: Is it worth it making this a `secureString'?
If this works like a token, I'd say yes.

smoothcontract: Discussed offline - the value isn't something that needs to be kept exceptionally secret as there is very low potential for misuse if it was made visible on the application configuration. It's just to allow exception reporting.

PR #208, opened 2019-09-25
Alter locked sections user flow on task-list

tavy88:
The new flow is as follows:

1. When the user has no session, lands on task-list page
clicks on Check your existing skills, gets taken to search
for current job title form (check_your_skills path).

2. When the user selects a set of skills, which are persisted
already in the session by that time, all the sections on
task-list get unlocked. The user DOES NOT have to click on
Find what you can do next button anymore as per discussion with
Shelina & Lois.

3. If a user already has skills persisted on the session, when
landing to the task-list and clicking Check your skills link,
this will take the user to the skills summary page (skills_path).

This is compatible with bothe Skills Builder v1 & v2.

https://dfedigital.atlassian.net/browse/GET-417

### NOTE
We are aware that we are breaking the breadcrumbs search link currently as we are not passing a search term anymore at the task list level. Lois is aware of it and it will be tackled in the next sprints.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -7,6 +7,16 @@ class PagesController < ApplicationController
     user_session.track_page('next_steps')
   end
 
+  def task_list
+    if user_session.job_profile_skills?
+      current_job = JobProfile.find(user_session.job_profile_ids.first)
+
+      @skills_summary_path = skills_path(job_profile_id: current_job.slug)
+    else
+      @skills_summary_path = check_your_skills_path
+    end
+  end
+
   def location_eligibility
     track_event(:pages_location_eligibility_search, search: postcode) if postcode.present?
     @search = CourseGeospatialSearch.new(postcode: postcode)
--- a/app/views/pages/task_list_items/_item_1.html.erb
+++ b/app/views/pages/task_list_items/_item_1.html.erb
@@ -5,8 +5,8 @@
   <ul class="app-task-list__items">
     <li class="app-task-list__item">
       <div class="app-task-list__task-name">
-        <%= link_to check_your_skills_path, class: 'govuk-link' do %>
-          Check your existing skills <span class="app-step-nav__context">5 minutes</span>
+        <%= link_to @skills_summary_path, class: 'govuk-link' do %>
+          Check your existing skills
         <% end %>
         <p class="govuk-body govuk-!-margin-bottom-0">You'll be surprised at what you already know.</p>
       </div>
--- a/app/views/pages/task_list_items/_item_2.html.erb
+++ b/app/views/pages/task_list_items/_item_2.html.erb
@@ -5,8 +5,8 @@
   <ul class="app-task-list__items">
     <li class="app-task-list__item">
       <div class="app-task-list__task-name">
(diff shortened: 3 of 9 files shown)


app/controllers/pages_controller.rb
```diff
@@ -7,6 +7,10 @@ def next_steps
     user_session.track_page('next_steps')
   end
 
+  def task_list
+    @job_profile = JobProfile.find(user_session.job_profile_ids.first) if user_session.job_profile_skills?
```
cwrw: instead of first id I would go with last id, which would be the last job profile they selected

app/views/pages/task_list_items/_item_1.html.erb
```diff
@@ -5,8 +5,8 @@
   <ul class="app-task-list__items">
     <li class="app-task-list__item">
       <div class="app-task-list__task-name">
-        <%= link_to check_your_skills_path, class: 'govuk-link' do %>
-          Check your existing skills <span class="app-step-nav__context">5 minutes</span>
+        <%= link_to user_session.job_profile_skills? ? skills_path(job_profile_id: @job_profile.slug) : check_your_skills_path, class: 'govuk-link' do %>
```
cwrw: instead of doing the user session job profile skill check twice, maybe we do it once in the controller and build the path there instead, and just pass it in here?

something like this maybe
```
def task_list
  @something_path = user_session.job_profile_skills? skills_path(job_profile_id: last_job_profile) : check_your_skills_path
end

private 
def last_job_profile
  job_profile = JobProfile.find(user_session.job_profile_ids.last).slug
end
```
or maybe check for job profile instead here

app/views/pages/task_list_items/_item_2.html.erb
```diff
@@ -5,8 +5,8 @@
   <ul class="app-task-list__items">
     <li class="app-task-list__item">
       <div class="app-task-list__task-name">
-        <%= link_or_blocked(path: skills_matcher_index_path, enabled: user_session.page_visited?('skills_matcher_index'), span_id: 2) do %>
```
cwrw: since we don't care about this anymore, can we remove the `track_page` logic from all the pages we won't be looking for anymore? This would be in the `pages_controller` as well as `skills_matcher_controller`

app/controllers/pages_controller.rb
tavy88: I think we always want to have the current job. Last will be previous jobs so out of date. I intentionally picked first.

app/views/pages/task_list_items/_item_2.html.erb
tavy88: This line is out of date.

tavy88: Sorry not following why do you want to remove the `track_page` - no page has been removed, just the unlocking mechanism changed.

tavy88:
@cwrw Spoke with Lois FYI and it's been decided:

1. The summary will keep the current profile id 
2. We still want to keep tracking sections 3 & 4 on sessions as we might need to revert back to this.

app/controllers/pages_controller.rb
tavy88: See my comment on the bottom of the PR.

PR #213, opened 2019-09-27
GET-497 Order by growth then alphabetically on skills matching

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-497

After grouping on skills match, order by growth, and then order alphabetically. I'd like to tackle the flakiness next in another ticket

Please find below a few examples of the ordering. I’ve uncovered the values for testing purposes:

First order through score:
Screenshot 2019-09-27 at 13 00 02
Then through growth if the score is equal:
Screenshot 2019-09-27 at 12 47 14
Then Alphabetically if the score and growth are equal:
Screenshot 2019-09-27 at 12 46 08
Edgecase for growth being nil means it will appear at the top of the equal score group:
Screenshot 2019-09-27 at 12 46 20

### DIFF
--- a/spec/models/skills_matcher_spec.rb
+++ b/spec/models/skills_matcher_spec.rb
@@ -126,13 +126,81 @@ RSpec.describe SkillsMatcher do
       )
 
       matcher = described_class.new(UserSession.new(session))
-      puts matcher.send(:build_query).to_sql
-      puts ActiveRecord::Base.connection.execute('select datname, datcollate from pg_database;').first
-
       expect(matcher.match).to eq(
         [
           job_profile4,
           job_profile3,
+          job_profile5,
+          job_profile2
+        ]
+      )
+    end
+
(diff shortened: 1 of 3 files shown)


spec/models/skills_matcher_spec.rb
```diff
@@ -126,13 +126,81 @@
...
+          job_profile2
+        ]
+      )
+    end
+
+    it 'arranges job profiles in job growth order unless job growth is nil' do
+      skill = create(:skill)
+      job_profile1 = create(:job_profile, skills: [skill])
+      job_profile2 = create(:job_profile, skills: [skill], growth: -5)
+      job_profile3 = create(:job_profile, skills: [skill], growth: nil)
```
cwrw: confirmed with Lois about the nil coming first, but going to double check with shelina as well in review

PR #214, opened 2019-09-27
GET-511 Change breadcrumbs on sign in pages, and fix old wrong ones

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-511

### DIFF
--- a/app/views/pages/location_eligibility.html.erb
+++ b/app/views/pages/location_eligibility.html.erb
@@ -3,7 +3,7 @@
     generate_breadcrumbs(
       t('breadcrumb.location_eligibility'),
       [
-        [t('breadcrumb._home'), root_path]
+        [t('breadcrumb.home'), root_path]
       ]
     )
   end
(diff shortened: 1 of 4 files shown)


tavy88:
No broken specs?

cwrw:
No, there aren't general specs for breadcrumbs, only the edgecases in breadcrumbs

app/views/pages/location_eligibility.html.erb
```diff
@@ -3,7 +3,7 @@
     generate_breadcrumbs(
       t('breadcrumb.location_eligibility'),
       [
-        [t('breadcrumb._home'), root_path]
+        [t('breadcrumb.home'), root_path]
```
smoothcontract: Very odd, was that _home just a typo previously? Not mentioned in the locale file anywhere.

cwrw: This is the error that came up in the accessibility test I think when I walked through it, where it couldn't find it so defaulted

PR #216, opened 2019-09-30
Add custom healthcheck and swap parameters

mafonso:
### Context

The Slot Swap operation is a bit flaky. 
After going through the documentation and by experimentation we surfaced a few internal settings that seem to improve how this internal operation works. 

This PR adds a custom health-check endpoint which is more adequate to probe the healthiness of the application. And another setting aimed and improve the behaviour for Linux custom containers that do not interface with Azure the same way native .NET apps do.

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -194,7 +194,9 @@
     "domainDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-domainPsql')]",
     "restrictedDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-restricedPsql')]",
     "actionGroupName": "[concat(variables('resourceNamePrefix'), '-ag')]",
-    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]"
+    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]",
+    "warmupPingPath": "/status",
+    "containerStartTimeLimit": "300"
   },
   "resources": [
     {
@@ -483,6 +485,18 @@
           },
           "appServiceAppSettings": {
             "value": [
+              {
+                "name": "WEBSITE_SLOT_POLL_WORKER_FOR_CHANGE_NOTIFICATION",
+                "value": "0"
+              },
+              {
+                "name": "WEBSITE_SWAP_WARMUP_PING_PATH",
+                "value": "[variables('warmupPingPath')]"
+              },
+              {
+                "name": "WEBSITES_CONTAINER_START_TIME_LIMIT",
+                "value": "[variables('containerStartTimeLimit')]"
+              },
               {
                 "name": "ENVIRONMENT_NAME",
                 "value": "[variables('environmentName')]"


azure/template.json
```diff
@@ -484,8 +485,12 @@
           "appServiceAppSettings": {
             "value": [
               {
-                "name": "ENVIRONMENT_NAME",
-                "value": "[variables('environmentName')]"
+                "name": "WEBSITE_SLOT_POLL_WORKER_FOR_CHANGE_NOTIFICATION",
```
mafonso: Not much information on this setting, apart from [this tweet](https://twitter.com/hedgeb/status/951589547632218112). It does seem to make the swap process smoother.

azure/template.json
```diff
@@ -194,7 +194,8 @@
     "domainDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-domainPsql')]",
     "restrictedDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-restricedPsql')]",
     "actionGroupName": "[concat(variables('resourceNamePrefix'), '-ag')]",
-    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]"
+    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]",
+    "WEBSITE_SWAP_WARMUP_PING_PATH": "/status"
```
cwrw: should this be in app settings?

azure/template.json
```diff
@@ -484,8 +485,12 @@
           "appServiceAppSettings": {
             "value": [
               {
-                "name": "ENVIRONMENT_NAME",
```
cwrw: do we not need those anymore?

azure/template.json
cwrw: can we follow the convention of camel case instead of snake here for naming?

mafonso: Yes, its just that apparently that is the naming convention for app variables for "internal" use.
But in most examples they are even hardcoded directly in the webapp resource.

azure/template.json
mafonso: We do. I just rebased before to make sure this was in. I'll check.

azure/template.json
```diff
@@ -194,7 +194,10 @@
     "domainDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-domainPsql')]",
     "restrictedDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-restricedPsql')]",
     "actionGroupName": "[concat(variables('resourceNamePrefix'), '-ag')]",
-    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]"
+    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]",
+    "warmupPingPath": "/status",
+    "warmupPingStatus": "200",
+    "containerStartTimeLimit": "300"
```
mafonso: Added more checks.

azure/template.json
```diff
@@ -194,7 +194,10 @@
     "domainDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-domainPsql')]",
     "restrictedDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-restricedPsql')]",
     "actionGroupName": "[concat(variables('resourceNamePrefix'), '-ag')]",
-    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]"
+    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]",
+    "warmupPingPath": "/status",
+    "warmupPingStatus": "200",
+    "containerStartTimeLimit": "600"
```
smoothcontract: Is there any benefit in making this more aggressive? 10 minutes seems like quite a long time to wait for the service to start up. If the average is 1-2 minutes then perhaps make this more like 5 minutes?

mafonso: As discussed it was not clear how this value was impacting the process.   At the time of this comment it was 600s but the latest value was 300.

PR #218, opened 2019-09-30
GET-407 Refactor query and tidy up because there is no need for complexity

cwrw:
When unnesting an aggregate array with ordinality with minimal syntax:
`WITH ORDINALITY ordered_id`, this was minimising the following
syntax:
`WITH ORDINALITY some_table(ids, ordinality_number)`
But this does not guarantee the order of the unnested arrays is maintained.
Explicitly ordering on the ordinality number as a secondary order
after the skills match makes sure the order is retained.
`

EDIT:
After some consideration realised we can tidy up this query and remove the unnecessary step of array aggregate ordering and un-nesting. This can be done with a normal `.order`

The following query should maintain the previous behaviour as well as eliminate the flakiness

jira-ticket: https://dfedigital.atlassian.net/browse/GET-407

### DIFF
--- a/app/models/skills_matcher.rb
+++ b/app/models/skills_matcher.rb
@@ -20,7 +20,7 @@ class SkillsMatcher
 
   private
 
-  def job_profile_skills_subquery
+  def skills_matched_query
     JobProfileSkill
       .select(:job_profile_id, 'COUNT(job_profile_id) AS skills_matched')
       .where(skill_id: user_session.skill_ids)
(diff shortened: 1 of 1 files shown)


app/models/skills_matcher.rb
```diff
@@ -20,38 +20,29 @@ def job_profile_scores
 
   private
 
-  def job_profile_skills_subquery
+  def ranked_job_profiles_query
     JobProfileSkill
       .select(:job_profile_id, 'COUNT(job_profile_id) AS skills_matched')
```
smoothcontract: maybe consider renaming this to `ranked_job_profiles` - `skills_matched` doesn't immediately tie this together in the `build_query` method below. Alternatively, rename the method to `skills_matched_query` ?

PR #230, opened 2019-10-03
GET-440: Address accessibility issues

tavy88:
Address accessibility issues.

Header:

missing image description for Gov UK - addressed

breadcrumbs need navigation section indicator - addressed

breadcrumb groups confusing (always repeats on last element) - can't replicate

Footer: 

OGL unlabelled image - addressed

feedback form after selecting no not easy to navigate to - this will be changing when we update the in page survey anyway - not worth the effort as this is going away

General:

Skip to main content: includes beta banner and breadcrumbs, might skip to either breadcrumbs or actual content instead - investigate GDS or DAC guidance - addressed

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -30,9 +30,7 @@
         <div class="govuk-header__logo">
           <%= link_to "https://www.gov.uk/", class: "govuk-header__link govuk-header__link--homepage" do %>
             <span class="govuk-header__logotype">
-              <svg role="img" focusable="false" class="govuk-header__logotype-crown" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 132 97" height="32" width="36" aria-labelledby="govukBannerTitle govukBannerDesc">
-                <title id="govukBannerTitle">Crown Logo</title>
-                <desc id="govukBannerDesc">White crown on a black background</desc>
+              <svg role="img" focusable="false" class="govuk-header__logotype-crown" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 132 97" height="32" width="36" aria-label="White crown on a black background">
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -25,15 +25,16 @@
...
-                <image src="/assets/images/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image"></image>
+                <image src="/assets/images/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image" alt=""></image>
```
smoothcontract: Is there any benefit in adding an empty alt tag at all?

app/views/layouts/application.html.erb
```diff
@@ -56,7 +56,7 @@
             </span>
           </p>
         </div>
-        <nav class="govuk-breadcrumbs" aria-label="Breadcrumb">
+        <nav class="govuk-breadcrumbs" aria-label="Breadcrumb" id="main-content">
```
smoothcontract: I thought we had planned to move the `main-content` to the first main heading on the page, rather than the breadcrumbs? That would suggest the `yield` part within the main application layout. Perhaps add a wrapper div around that instead? e.g.

```html
<div id="main-content">
        <%= yield %>
</div>
```

smoothcontract:
Looks good but left a couple of comments, mostly for my education I think :)

app/views/layouts/application.html.erb
tavy88: Please check the commit message.

app/views/layouts/application.html.erb
tavy88: The ticket has both options open-ended. My suggestion was to wait and see what DAC says maybe? Then we can move it to the yield. Or we can add it now I don't mind. There's a commit message around it.

app/views/layouts/application.html.erb
smoothcontract: I did read the commit message, and still have the same question: Why is adding an empty alt tag better than excluding the alt tag.. 🤔

app/views/layouts/application.html.erb
smoothcontract: Not sure it's a question for DAC really, it's more of a UI/business decision for us to make. Maybe see what the wider team feels when this is on a review environment then.

app/views/layouts/application.html.erb
```diff
@@ -71,7 +72,7 @@
         <div class="govuk-footer__meta">
           <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
 
-            <svg role="img" focusable="false" class="govuk-footer__licence-logo" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 483.2 195.7" height="17" width="41">
+            <svg role="img" focusable="false" class="govuk-footer__licence-logo" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 483.2 195.7" height="17" width="41" aria-label="Open Government Licence initials">
```
cwrw: ✂️ Licence/License

app/views/layouts/application.html.erb
tavy88:
```ruby
Some images do not need descriptions. Use alt="" if the image is:

- decorative and does not include important content
- an icon that has a text label that the alt text would repeat
- used in a link and the image is not needed to understand the link
```
From https://design-system.service.gov.uk/styles/images/

app/views/layouts/application.html.erb
tavy88: I did add it as you suggested. Seems like gov.uk does the same, they don't have breadcrumbs though.

app/views/layouts/application.html.erb
tavy88: https://www.grammar-monster.com/easily_confused/licence_license.htm

PR #232, opened 2019-10-03
GET-441 Fix Job Profile page accessibility as well as courses and navigation

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-441

For background images accessiblity there were quite a few implementations.
I went with the simpler one seen here: https://stackoverflow.com/questions/48913759/how-to-add-alt-text-to-background-images-making-background-images-accessible/#48913760
There is another approach with much more markup, having tested this, this seems adequate for now.

Also add hidden text to courses for providers 

Whilst we can't remove the extra list due to the clearfix class
being added by govuk frontend, we can point to the current page:
https://design-system.service.gov.uk/components/breadcrumbs/

We are adhering exactly to the rules of breadcrumbs seen there,
so Im going to keep it that way for now

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -11,7 +11,7 @@ module ApplicationHelper
             link_to(title, link, class: 'govuk-breadcrumbs__link')
           }
         },
-        content_tag(:li, current_page, class: 'govuk-breadcrumbs__list-item')
+        content_tag(:li, current_page, class: 'govuk-breadcrumbs__list-item', aria: { current: 'page' })
       ]
     )
   end
--- a/app/views/job_profiles/_highlights.html.erb
+++ b/app/views/job_profiles/_highlights.html.erb
@@ -2,16 +2,12 @@
   <div class="govuk-grid-column-one-half">
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-one-quarter govuk-!-margin-top-5 govuk-!-margin-bottom-6">
-        <div class="profile-highlights salary-icon"></div>
+        <div class="profile-highlights salary-icon" role="img" aria-label="Pound sterling icon"></div>
       </div>
       <div class="govuk-grid-column-three-quarters">
         <h4 class="govuk-heading-s govuk-!-margin-top-4 govuk-!-margin-bottom-1">Average Salary</h4>
         <hr class="govuk-section-break govuk-section-break--visible govuk-!-margin-bottom-2 responsive-section-break">
-        <ul class="govuk-list">
-          <li>
-            <%= @job_profile.salary_range %>
-          </li>
-        </ul>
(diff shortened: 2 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -11,7 +11,7 @@ def generate_breadcrumbs(current_page, previous_pages)
             link_to(title, link, class: 'govuk-breadcrumbs__link')
           }
         },
-        content_tag(:li, current_page, class: 'govuk-breadcrumbs__list-item')
+        content_tag(:li, current_page, class: 'govuk-breadcrumbs__list-item', aria: { current: 'page' })
```
tavy88: Does this work, because according to https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html it should be added to `a`:

Screen Shot 2019-10-04 at 09 17 56

cwrw: I was going with the govuk markup: https://design-system.service.gov.uk/components/breadcrumbs/
```
<div class="govuk-breadcrumbs">
  <ol class="govuk-breadcrumbs__list">
    <li class="govuk-breadcrumbs__list-item">
      <a class="govuk-breadcrumbs__link" href="#">Home</a>
    </li>
    <li class="govuk-breadcrumbs__list-item">
      <a class="govuk-breadcrumbs__link" href="#">Passports, travel and living abroad</a>
    </li>
    <li class="govuk-breadcrumbs__list-item" aria-current="page">Travel abroad</li>
  </ol>
</div>
```

tavy88: Ok, still unsure if it actually works... even on their design system. It kind of contradicts the documentation from w3.org.

cwrw: also looks like it can be used elsewhere: http://design-patterns.tink.uk/aria-current/

app/views/job_profiles/_highlights.html.erb
```diff
@@ -2,29 +2,25 @@
   <div class="govuk-grid-column-one-half">
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-one-quarter govuk-!-margin-top-5 govuk-!-margin-bottom-6">
-        <div class="profile-highlights salary-icon"></div>
+        <div class="profile-highlights salary-icon" role="img" aria-label="Pound sterling icon"></div>
```
smoothcontract: Yes so with these icons as I mentioned in the channel yesterday, I think they should be hidden from screen readers entirely. What do you think?

cwrw: discussed in person but for posterity keeping for now until we get the DAC assessment to get more guidance

app/helpers/application_helper.rb
tavy88: Yeah let's see what DAC says.

PR #233, opened 2019-10-04
Bing Spell Check API

tavy88:
### Description

Integrate Bing Spell Check API

Create a service that consumes the Bing Spell Check API.

The spell check strategy used is 'spell' as opposed to 'proof'.

'Proof' uses most of the language codes including en-gb:

https://docs.microsoft.com/en-us/azure/cognitive-services/bing-spell-check/language-support

setLang does not need to be set as the default is EN.

https://docs.microsoft.com/en-us/rest/api/cognitiveservices-bingsearch/bing-spell-check-api-v7-reference#setlang

### Screenshot

Screen Shot 2019-10-08 at 07 23 29

Screen Shot 2019-10-09 at 14 17 45

Screen Shot 2019-10-09 at 14 17 56

### Playground

https://azure.microsoft.com/en-gb/services/cognitive-services/spell-check/

Terms to test: thrpy, shopr


app/controllers/check_your_skills_controller.rb
```diff
@@ -3,6 +3,7 @@ def index
     redirect_to(skills_path) if Flipflop.skills_builder_v2? && user_session.job_profiles_cap_reached?
 
     @job_profile_search = JobProfileSearch.new(term: search, profile_ids_to_exclude: profile_ids_to_exclude)
+
```
cwrw: ✂️

app/controllers/check_your_skills_controller.rb
```diff
@@ -13,6 +14,8 @@ def results
 
     @job_profile_search = JobProfileSearch.new(term: search, profile_ids_to_exclude: profile_ids_to_exclude)
     @job_profiles = Kaminari.paginate_array(@job_profile_search.search).page(params[:page])
+
+    spell_check_searched_term!
```
cwrw: definition of bang methods: 
```
Bang (!) Methods
Ruby methods that modify an object in-place and end in an exclamation mark are known as bang methods. By convention, the bang labels a method as dangerous - specifically, as the dangerous equivalent of a method with the same name but without the bang.
```
Im not sure this is a destructive action as nothing is being modified in place, so can we remove the `!`?

app/services/spell_check_service.rb
```diff
@@ -0,0 +1,90 @@
...
+    @api_key = api_key
+  end
+
+  def scan(search_term: nil)
+    return unless search_term
+
+    service_enabled?
+
+    bing_spell_check(search_term: search_term)
+  rescue RuntimeError => e
```
cwrw: why not standard error?

app/services/spell_check_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  SpellCheckServiceError = Class.new(StandardError)
+  SpellCheckBingAPIError = Class.new(SpellCheckServiceError)
+  BadRequestError = Class.new(SpellCheckBingAPIError)
+  UnauthorizedError = Class.new(SpellCheckBingAPIError)
+  ForbiddenError = Class.new(SpellCheckBingAPIError)
+  RateLimitExceededError = Class.new(SpellCheckBingAPIError)
+
+  API_ENDPOINT = 'https://api.cognitive.microsoft.com/bing/v7.0/spellcheck'.freeze
+
+  HTTP_OK_CODE = 200
```
cwrw: why do we need to segregate all those responses? can't we just log the error as is with the status?

app/controllers/check_your_skills_controller.rb
```diff
@@ -30,4 +33,16 @@ def search
   def job_profile_params
     params.permit(:search)
   end
+
+  def spell_check_service
+    @spell_check_service ||= SpellCheckService.new
+  end
+
+  def spell_check_searched_term!
+    return if params[:page].present? || search.nil?
```
cwrw: does this mean if I go to a page directly by a link I won't get the spell check?

app/services/spell_check_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def api_key
+    raise 'BingSpellCheck API Key is not set' unless @api_key.present?
+
+    @api_key
+  end
+
+  def service_enabled?
+    raise 'This service is not currently enabled' unless Flipflop.spell_check?
```
cwrw: Since we have a feature flag for this, I think we should add this flag at the partial level than here, so that can be stripped out easily and not raise an exception

app/services/spell_check_service.rb
```diff
@@ -0,0 +1,90 @@
...
+    @client ||= Faraday.new { |client|
+      client.request :url_encoded
+      client.adapter Faraday.default_adapter
+      client.headers['Content-Type'] = 'application/x-www-form-urlencoded'
+      client.headers['Ocp-Apim-Subscription-Key'] = api_key
+    }
+  end
+
+  def api_key
+    raise 'BingSpellCheck API Key is not set' unless @api_key.present?
```
cwrw: I don't htink we need to raise an exception when there is  no key available, but rather fail silently. As this makes development and tests fail otherwise when they touch this page

app/views/shared/search/_form.html.erb
```diff
@@ -5,5 +5,6 @@
     <%= errors_tag @job_profile_search, :term %>
     <%= text_field(nil, :search, class: 'govuk-input search-input govuk-!-width-one-half', placeholder: t('placeholder', scope: scope), aria: { label: 'Search' }) %>
     <%= button_tag('Search', class: 'govuk-button search-button', aria: { label: 'Search button' }) %>
+    <%= render 'shared/search/spell_check' %>
```
cwrw: don't think we need this here

app/views/shared/search/_results_form.html.erb
```diff
@@ -4,6 +4,8 @@
       <%= errors_tag @job_profile_search, :term %>
       <%= text_field(nil, :search, value: params[:search], class: 'search-input govuk-input govuk-!-width-one-half', placeholder: t('placeholder', scope: scope), aria: { label: 'Search' }) %>
       <%= button_tag('Search', class: 'govuk-button search-button', aria: { label: 'Search button' }) %>
+      <%= render 'shared/search/spell_check' %>
     <% end %>
   <% end %>
 </div>
+
```
cwrw: ✂️

app/views/shared/search/_spell_check.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if @spell_checked_search.present? %>
+  <p class="govuk-body govuk-!-margin-top-3">
+    Did you mean <%= link_to @spell_checked_search, check_your_skills_path(search: @spell_checked_search), class: 'govuk-link' %>
```
cwrw: this should be `results_check_your_skills_path`

config/environments/test.rb
```diff
@@ -43,6 +43,8 @@
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
 
+  config.bing_spell_check_api_key = 'valid API key'
```
cwrw: can we add variables to .env files instead, as that is what we've been doing throughout the codebase, as now we have 3 strategies and its getting inconsistent and confusing

app/services/spell_check_service.rb
```diff
@@ -0,0 +1,90 @@
...
+      req.params['text'] = search_term
+    }
+
+    return suggested_correction if response_successful?
+
+    handle_api_errors
+  end
+
+  def suggested_correction
+    JSON.parse(response.body)['flaggedTokens'].map { |token| token['suggestions'].first['suggestion'] }.join(' ')
```
cwrw: are we not planning on using the score that comes back as well?

cwrw: Another question, what happens if people repeat words in their search?

app/services/spell_check_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def client
+    @client ||= Faraday.new { |client|
```
cwrw: I prefer to use something native to ruby like net http instead of the gem faraday, which has been added as a dependency to other gems rather than explicitly, meaning that if we remove those gems we will lose this gem and this won't work

app/controllers/check_your_skills_controller.rb
tavy88: I see it as mutating a value instead of returning a value. It sets that instance variable, hence the reason why I wanted to use bang.

app/services/spell_check_service.rb
tavy88: What would a possible example of a StandardError be here?

app/controllers/check_your_skills_controller.rb
tavy88: Yes. And I think it should be the correct behaviour. Going to a page doesn't mean you are doing a search, therefore it needs no correction.

app/services/spell_check_service.rb
tavy88: I think it's a cleaner implementation of the service this way.

app/views/shared/search/_form.html.erb
tavy88: Can you explain what you mean please?

app/views/shared/search/_spell_check.html.erb
tavy88: It will redirect to `results_check_your_skills_path` and do the validation of the input before right? Do we want to skip the validation?

config/environments/test.rb
tavy88: We could yeah.

app/services/spell_check_service.rb
tavy88: We are not planning to use the scores I don't think so. Where would that fit in the scope of the current story? The results are ordered by score, and as BING itself does it only picks the first token variation.

app/services/spell_check_service.rb
tavy88: This is just a preference though. Faraday is common client library used in service integration & development. Ditching that now would take a lot of work and I don't really see the value in doing so.

app/controllers/check_your_skills_controller.rb
cwrw: setting the value is not the same as mutating a value in place/having a destructive method. A good example can be the following:
```
irb(main):001:0> foo = "BAR"
=> "BAR"
irb(main):002:0> foo.downcase
=> "bar"
irb(main):003:0> foo
=> "BAR"
irb(main):004:0> foo.downcase!
=> "bar"
irb(main):005:0> foo
=> "bar"
```
Instead we are setting an instance variable, something we do across the code without setting the '!' example job profile across all controllers

app/services/spell_check_service.rb
tavy88: Losing the gem is also equivalent of saying dropping the Notify gem would stop our app from sending emails I think. If we ever come to that decision, we will react then I guess.

app/services/spell_check_service.rb
cwrw: This extra code is serving us no value, unless we do need to check for all those status codes, we should only go for the successful code "200" or fail otherwise, as this bloats the class and adds unnecessary complexity. We lose nothing if we log the status code in the error message and add the full inspection

app/controllers/check_your_skills_controller.rb
tavy88: I would like to have a new line between this and `redirect_to`.

app/views/shared/search/_spell_check.html.erb
cwrw: The validation happens on both pages, on this one it happens here, see https://github.com/DFE-Digital/get-help-to-retrain/blob/master/app/models/job_profile_search.rb#L13

app/views/shared/search/_form.html.erb
cwrw: This form only appears on the check your skills index page: https://github.com/DFE-Digital/get-help-to-retrain/search?q=%22shared%2Fsearch%2Fform%22&unscoped_q=%22shared%2Fsearch%2Fform%22, where we do not need to show the spell check

app/services/spell_check_service.rb
tavy88: And to the second question:

```ruby
{
  "_type": "SpellCheck",
  "flaggedTokens": [
    {
      "offset": 0,
      "token": "thrpy",
      "type": "UnknownToken",
      "suggestions": [
        {
          "suggestion": "therapy",
          "score": 1.0
        }
      ]
    },
    {
      "offset": 6,
      "token": "thrpy",
      "type": "UnknownToken",
      "suggestions": [
        {
          "suggestion": "therapy",
          "score": 1.0
        }
      ]
    }
  ]
}
```

app/views/shared/search/_spell_check.html.erb
tavy88: I was thinking it would spare the API call but it won't so yeah I guess I can use the results_path.

app/controllers/check_your_skills_controller.rb
cwrw: but we are doing a search on subsequent pages every time, we just limit the window to the page. If someone bookmarks page 2, I expect them to see the spell check as well, also an example of this is in nrs as well: https://nationalcareers.service.gov.uk/search-results?searchTerm=paper+mker&page=2

app/controllers/check_your_skills_controller.rb
tavy88: I know what a mutation is and what bang does. I will drop the bang but I was explaining the reasoning behind it. :)

app/services/spell_check_service.rb
cwrw: I think it falls into scope as this is to implement the spell check and make sure we get back *good* results. If you still feel this is out of scope, we should flag this to business owners to investigate how good our spell checker returns values for more vague spellings, and to keep an eye out if we need to amend further

app/services/spell_check_service.rb
tavy88: These codes are returned from the API and they are documented as they return corresponding error messages. We need to capture them. Here's an example:

Screen Shot 2019-10-08 at 14 17 59

We need to know exactly the reason for the failure and the returned code.

app/controllers/check_your_skills_controller.rb
smoothcontract: This might need some more thought - it makes sense to preserve the spelling suggestion, but is also expensive to make repeated API calls for the same search term when a user paginates. Can you investigate an approach for caching the search perhaps? We can use the in memory cache store for now as we don't have any better infrastructure support: https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-memorystore

app/services/spell_check_service.rb
cwrw: `StandardError` is the parent of `RuntimeError` thus covers a wider range of errors including runtime

app/services/spell_check_service.rb
cwrw: So this gem is not directly added to our gemfile but is rather a dependency of other gems, and is used in Sentry and Split io, if we were to remove those gems later on we will lose support for faraday. This is why I prefer to use something native to Ruby.

app/services/spell_check_service.rb
cwrw: so does that mean we will show therapy twice, I think we should show it once instead, might be good to ask the POs about this to make sure and add a test etc.

app/controllers/check_your_skills_controller.rb
tavy88: We discussed that navigating through the results would kill the API. I will raise it with Lois and explain the dangers.

app/views/shared/search/_form.html.erb
tavy88: Fair.

app/services/spell_check_service.rb
cwrw: I think we shouldn't add it as a dependency though, if you prefer to keep this, that is fine, but we should keep this in mind as an issue that might happen

app/services/spell_check_service.rb
tavy88: Question for POs. Personally I don't think we should interpret results. If that's what they enter that is what they get. Garbage in garbage out. It's an edge case.

app/services/spell_check_service.rb
tavy88: As I said the change required is too big and the effort is not justified. I'd like to keep Faraday. If one of the gems is removed we will capture the regression on our tests. It won't go unnoticed. Does it make sense?

app/services/spell_check_service.rb
smoothcontract: The service should wrap underlying complexity of invoking the API - if we don't intend to implement behaviour that depends on these response codes then there's little value in distinguishing between them.

app/services/spell_check_service.rb
smoothcontract: Hmm.. I think we have this pattern in a couple of other services already, but the expected behaviour for this service is subtly different. It's essentially an optional service, so if no API key is defined then the service shouldn't be invoked. I'm not sure where the best place for that logic would be, however with the current implementation this would mean the controller needs to rescue from this exception and presumably ignore it. It's generally not recommended to raise an exception by design for an expected case - they should be used (as the name suggests) for exceptional scenarios).

app/services/spell_check_service.rb
smoothcontract: I agree, it doesn't make sense to include feature flags within the service itself. We haven't done that for other features - the logic is tested at view and routing level generally.

app/services/spell_check_service.rb
cwrw: I agree, we need to show the full message, which you do in `handle_api_errors`. But instead of having 5 error classes we can have one, with the full logging, until we do need to have specific behaviour for the different failures

config/environments/test.rb
```diff
@@ -43,6 +43,8 @@
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
 
+  config.bing_spell_check_api_key = ENV['BING_SPELL_CHECK_API_KEY']
```
smoothcontract: This line is redundant - it's already taken care of in `application.rb` above

app/services/spell_check_service.rb
tavy88: Using `StandardError` here will raise `SpellCheckService::SpellCheckServiceError` instead of `SpellCheckService::UnauthorizedError`. Code related erros will be RuntimeError. We do the same for NotifyService.

app/services/spell_check_service.rb
tavy88: What if you call the service outside the view?

app/services/spell_check_service.rb
tavy88: I think it's more useful to see a clear reason rather than seeing nil. Besides we have the same pattern for Notify.

app/services/spell_check_service.rb
cwrw: I agree mostly, but I would rather suggest we avoid the full dependency, and add a comment like you suggested about not using it, and add a tech debt ticket to rewrite in `Net::Http`, unless there is time to rewrite it now which I would prefer

app/services/spell_check_service.rb
tavy88: Best practices say that you should handle exceptions as granular as possible hence why I chose that route. I have simplified the implementation.

app/services/spell_check_service.rb
cwrw: it would look something like this, haven't tested this though:
```
  def response
    Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https', read_timeout: 120) do |http|
      response = http.request(request)
      fail SomError, "#{uri.host} - #{response.status} : #{response.message}" unless response.is_a?(Net::HTTPSuccess)

      response.body
    end
  rescue StandardError => e
    something
  end

  def request
    request = Net::HTTP::Get.new(uri)
    request.add_field('Ocp-Apim-Subscription-Key', api_key)
    request.content_type ='application/x-www-form-urlencoded'
    request
  end

  def uri
    uri = Addressable::URI.parse(base_url)
    uri.query_values = {
      mkt: 'en-gb',
      mode: 'spell',
      text: search_term
    }
    uri
  end
```

tavy88: I am leaving this for another PR. Ruben has told me there are urgent things coming on the pipeline so there's no time to tackle this now.

app/services/spell_check_service.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def api_key
+    raise 'BingSpellCheck API Key is not set' unless @api_key.present?
+
+    @api_key
+  end
+
+  def service_enabled?
+    raise 'This service is not currently enabled' unless Flipflop.spell_check?
```
tavy88: This check is at this level to prevent the consumption of this service in any context. It's easier to tackle this at this level rather than keep adding if statements everywhere we use this service. Yes, I had to choose between blocking this at the init level or at the service call level. However, I didn't want to make the initializer too noisy, hence the reason why I added at this level.

app/services/spell_check_service.rb
smoothcontract: Just to continue this thread.. we have our own code now that depends on a specific version of Faraday, hence we should add that to our Gemfile. Otherwise there is a risk that the other gems update to a newer version of faraday with a breaking change. I strongly believe it’s best to track that dependency so it doesn’t surprise us. And please let’s not use Faraday for anything else in future. There are lots of choices available for this, sadly Faraday is the most feature bloated one and really is overkill for our simple use case.

app/services/spell_check_service.rb
smoothcontract: Sorry I don’t agree. Notify is a dependency whereas this is optional functionality - it won’t be configured with an API key locally or on dev environments so doesn’t need to raise exceptions.

tavy88: In which case the feature will be off. We can't have the service ON with no API key. What's the point?

spec/features/check_your_skills_spec.rb
```diff
@@ -20,6 +22,80 @@
...
+            {
+              'suggestion': 'Gates',
+              'score': 1
+            }
+          ]
+        }
+      ]
+    }.to_json
+
+    allow(Faraday).to receive(:new).and_return(client)
```
cwrw: From what I can see here, this isn't the way we should be testing faraday, there are examples here on how to use with rspec: https://github.com/lostisland/faraday/blob/master/examples/client_spec.rb
Which aren't great. We can use `stub_request` instead easily here instead without having to step through all the method stubs for faraday, same for all tests below

spec/features/check_your_skills_spec.rb
```diff
@@ -20,6 +22,80 @@
...
+    allow(faraday_response).to receive(:body).and_return(response_body)
+    allow(faraday_response).to receive(:status).and_return(200)
+    allow(client).to receive(:get).with(SpellCheckService::API_ENDPOINT).and_return(faraday_response)
+
+    visit(check_your_skills_path)
+    fill_in('search', with: 'Gates')
+    find('.search-button').click
+
+    expect(page).not_to have_text('Did you mean Gates')
```
cwrw: There are a few scenarios we should also be testing for, those include clicking through the suggestion to see results, as well as seeing this suggestion on zero results pages as well

tavy88: The page has no results as we only build one profile with the name: bodyguard, but search for Gates.

app/controllers/check_your_skills_controller.rb
smoothcontract: @tavy88 so am I right in thinking that with the current implementation spelling suggestions are only shown on the first page?

app/services/spell_check_service.rb
smoothcontract: Well, we wouldn't. The feature flag starts with the user interface and that's where all the features are controlled. The service really does not need to check the feature flag is enabled here - it's unnecessary complexity and more tech debt to unpick later when we promote the feature.

app/controllers/check_your_skills_controller.rb
tavy88: Yes. But to be honest, I can't get a correction with results - and even more paginated results. So this scenario is not easy to achieve.

app/services/spell_check_service.rb
```diff
@@ -0,0 +1,72 @@
...
+    Rails.logger.error("Spell Check service error: #{e.message}")
+    raise SpellCheckServiceError, e.message
+  end
+
+  private
+
+  def bing_spell_check(search_term: nil)
+    uri.query = URI.encode_www_form(mkt: 'en-gb', mode: 'spell', text: search_term)
+
+    @response = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https') { |http|
```
smoothcontract: This should use the `do |http|` form of block syntax as it spans multiple lines. If rubocop isn't enforcing this it's because we had to nobble it slightly as it kept doing strange auto corrections with blocks.

app/services/spell_check_service.rb
tavy88: Alternatively, I can just let the user make API calls when navigating through pages, which I don't think will be a kill, because as I said I can't find a case like that yet.

app/controllers/check_your_skills_controller.rb
tavy88: Alternatively, I can just let the user make API calls when navigating through pages, which I don't think will be a kill, because as I said I can't find a case like that yet.

app/services/spell_check_service.rb
tavy88: Rubocop was the reason actually. It did require {}

app/services/spell_check_service.rb
tavy88: Why is it more complex to remove the feature check from here rather than controller level. What if you use this service in multiple controllers? You would need to add multiple Flipflop checks. I don't really get the point you are making.

app/controllers/check_your_skills_controller.rb
```diff
@@ -30,4 +33,17 @@ def search
...
     params.permit(:search)
   end
+
+  def spell_check_service
+    @spell_check_service ||= SpellCheckService.new
+  end
+
+  def spell_check_searched_term
+    return unless Flipflop.spell_check?
+    return if params[:page].present? || search.nil?
```
smoothcontract: Please change as discussed:

```ruby
return unless Flipflop.spell_check? && search.present?
```

cwrw:
spoke in person, changes proposed:
1) Making sure full sentence returned in search
2) spacing in form

otherwise all else is well, will wait for changes

cwrw:
I'm on my phone so can't do much, so I had a strange usecase with the following text: Paper makr paper mker which then returns paper maker paper, which I found odd, maybe we should take out the uniq now as it no longer applies with this change, what do you think?

tavy88:
> Paper makr paper mker

It’s not odd. Paper and paper are not equal:
irb(main):003:0> ['Paper', 'paper'].uniq
=> ["Paper", "paper"]

but mkr and mker both correct to maker which is duplicated.

cwrw:
looks good, tested all the things 👍

PR #234, opened 2019-10-04
GET-532 If user selected skills they should see save your results in sidebar

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-532

* On all pages if user selected skills they should see save your
results partial, otherwise they should see return to save results.

* Maintain logic in hiding partials on registration/sign in
* Fix tests to check all pages instead of certain ones for partials

### DIFF
--- a/app/views/shared/_save_or_return_to_your_results.html.erb
+++ b/app/views/shared/_save_or_return_to_your_results.html.erb
@@ -1,0 +1,5 @@
+<% if user_session.job_profile_skills? %>
+  <%= render 'shared/save_your_results'  %>
+<% else %>
+  <%= render 'shared/return_to_saved_results' %>
+<% end %>
--- a/app/webpacker/styles/_course.scss
+++ b/app/webpacker/styles/_course.scss
@@ -1,3 +1,11 @@
 .title {
   color: $govuk-link-colour;
 }
+
+.courses-margin {
+  margin-bottom: 20px;
+
+  @include govuk-media-query($from: tablet) {
+    margin-bottom: 315px;
+  }
+}
--- a/app/views/check_your_skills/index.html.erb
+++ b/app/views/check_your_skills/index.html.erb
@@ -18,9 +18,5 @@
     <% end %>
   </div>
   <%= render "shared/contact_us" %>
-  <% if Flipflop.skills_builder_v2? && user_session.job_profile_skills? %>
-    <%= render "shared/save_your_results"  %>
-  <% else %>
-    <%= render 'shared/return_to_saved_results' %>
-  <% end %>
+  <%= render 'shared/save_or_return_to_your_results' %>
 </div>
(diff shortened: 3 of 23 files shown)


app/views/shared/_save_or_return_to_your_results.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if user_session.job_profile_skills? %>
```
smoothcontract: Should this not retain the previous `if Flipflop.skills_builder_v2?` check too?

app/webpacker/styles/_course.scss
```diff
@@ -1,3 +1,11 @@
 .title {
   color: $govuk-link-colour;
 }
+
+.courses-margin {
```
smoothcontract: Seems unrelated?

app/views/shared/_save_or_return_to_your_results.html.erb
cwrw: We no longer need this check as initially we were showing this partial on the "add other jobs page" which would come up if we had a skill, but now we don't anymore. That page would only show on skills builder v2 (the locale changed on the page)
Now we want this rule everywhere, regardless if on that page or not so that flag no longer applies

app/webpacker/styles/_course.scss
cwrw: Adding this margin is required as that now we show the "larger" partial for the return to saved results, the sidebar is quite longer than the content column, so the sidebar will wrap under it. This makes sure this doesn't happen, see: https://get-help-to-retrain-pr-234.herokuapp.com/courses/maths?postcode=NW11+9TE and try removing the class from the div to replicate

app/views/shared/_save_or_return_to_your_results.html.erb
smoothcontract: Hmm... I'm nearly convinced. Does this definitely work as expected when skills builder v2 flag is turned off? Does the skills builder v1 not set job profile skills in the session too?

app/webpacker/styles/_course.scss
smoothcontract: Okay, I've tried that and can see the issue you describe. But.. that feels like there should be a more general case fix than just for the courses  list to prevent the sidebar from wrapping under the main content. Isn't there something that can be done at the layout level to fix this? Otherwise this issue could crop up on any page and it's really dependent on the respective heights of main and sidebar content. The fix here is addressing the symptom really, rather than the root cause. If it's beyond the scope of this ticket to address that then happy for that to be in a new ticket/PR. Thoughts?

cwrw: yea I agree, we are doing this currently in another 2 places in the codebase, which I'm not happy about. I wanted to read more about how to stop wrapping grid elements across on height. I think for now we can keep this but add a ticket to revisit this in future and fix this as Im sure there is an elegant solution

smoothcontract: cool 👍

PR #235, opened 2019-10-04
Get 534 search hidden titles

smoothcontract:
### Context
Add support for searching against hidden alternative titles for job profiles.

Adds new `hidden_titles` attribute to job profile, and extends job profile import service to support importing a new spreadsheet containing job profile slugs and hidden alternative titles. Added a rake task to support that and extended active admin to cover the new attribute too.

Modified job profile search to search for hidden titles if present - this is ranked between alternative title and description matches in the search results.

### Guidance to review
Will need the relevant spreadsheet to test locally - I'll add that to the channel. Then run db migration followed by the relevant rake task; very few job profiles will get updated but you should see output something like this:

```bash
rails "data_import:import_job_hidden_titles[JobProfileHiddenAlternativeTitles.xlsx]"
Importing job hidden titles data from JobProfileHiddenAlternativeTitles.xlsx...
Failed to find matching job profile for "car-salesperson"
Failed to find matching job profile for "customs-officer"

Results:
{:job_profiles_total=>865, :job_profiles_with_hidden_titles=>89, :job_profiles_missing_hidden_titles=>776, :errors=>2}
```

Once imported the new attributes can be seen at http://localhost:3000/admin/job_profiles

Job profile search will also respect the new job profile attribute. Note that this attribute will not be shown to the user, it's just used to provide search enrichment.


app/services/job_profile_import_service.rb
```diff
@@ -31,16 +36,36 @@ def import_growth_stats
...
   end
 
+  def import_hidden_titles(filename)
+    file = Roo::Spreadsheet.open(filename)
+    sheet = file.sheet('JobProfile')
+
+    sheet.each(HIDDEN_TITLES_COLUMN_HEADINGS) do |data|
+      next if data == HIDDEN_TITLES_COLUMN_HEADINGS
+
+      update_job_profile(data, :slug)
```
cwrw: shouldn't this be hidden titles?

spec/services/job_profile_import_service_spec.rb
```diff
@@ -38,4 +39,32 @@
...
+      expect { importer.import_hidden_titles('foo.xlsx') }.to raise_exception(IOError)
+    end
+
+    it 'does not raise error with valid filename' do
+      expect { importer.import_hidden_titles(titles_path) }.not_to raise_exception
+    end
+
+    it 'updates matching job profiles' do
+      importer.import_hidden_titles(titles_path)
+      expect(ceo).to have_attributes(hidden_titles: 'Boss man,Big cheese')
```
cwrw: 😆

PR #236, opened 2019-10-07
GET-561 Run two instances in QA, UAT, PROD

mafonso:
### Context

Before going live we want to run our application in HA, so that we don't suffer an outage if one instance of application dies due to a crash or other instance specific isshue. 

This PR passes a parameter that overrides the default number of instances of 1.

### Guidance to review

This parameter overrides the default defined in the generic templates.

https://github.com/DFE-Digital/get-help-to-retrain-devops/blob/master/templates/app-service-plan.json#L32-L35

### DIFF
--- a/azure/config/qa.parameters.json
+++ b/azure/config/qa.parameters.json
@@ -92,6 +92,9 @@
     },
     "applyAlerts": {
       "value": true
+    },
+    "appServicePlanInstances": {
+      "value": 2
     }
   }
 }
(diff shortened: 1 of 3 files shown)


azure/config/qa.parameters.json
```diff
@@ -92,6 +92,9 @@
     },
     "applyAlerts": {
       "value": true
+    },
+    "appServicePlanInstances": {
+      "value": 2
     }
```
mafonso: I'm doing QA as well so that we can mimic the same conditions on our test environment. 
It adds a bit to the running costs of the environment, but we'll have a test environment where we can see and reproduce the issue in our tests.

PR #238, opened 2019-10-09
GET-239 Add Job Vacancies section to job profiles

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-238

* Add API client for Find a job
* Add frontend partial to job profiles
* Add keys to azure/heroku
* Add health check


.env.test
```diff
@@ -1,3 +1,5 @@
 USER_FEEDBACK_SMART_SURVEY_LINK=https://www.smartsurvey.co.uk/s/8IBQ1/
 NOTIFY_API_KEY=test
 GOOGLE_ANALYTICS_TRACKING_ID=
+FIND_A_JOB_API_ID=test
```
cwrw: Thinking of removing and stubbing client in feature tests instead 🤔 this will avoid making calls to the api if we ever add more tests, thoughts?

app/services/find_a_job_service.rb
```diff
@@ -0,0 +1,75 @@
+class FindAJobService
+  attr_reader :api_id, :api_key, :query, :postcode, :distance, :path
+
+  BASE_URL = 'https://findajob.dwp.gov.uk/api/'.freeze
+  ResponseError = Class.new(StandardError)
+
+  def initialize(
```
smoothcontract: It seems a little odd to me to have the parameters for job vacancy lookup being passed in the constructor arguments as options. That would mean if I wanted to find vacancies for an alternate postcode or distance then the class needs to be reinstantiated. I think it would be preferable to pass only the required arguments to instantiate the class here i.e. api id and api key.

app/services/find_a_job_service.rb
```diff
@@ -0,0 +1,75 @@
...
+  )
+    @api_id = api_id
+    @api_key = api_key
+    @query = options[:query]
+    @postcode = options[:postcode]
+    @distance = options[:distance]
+    @path = options[:path] || 'search'
+  end
+
+  def job_vacancies
```
smoothcontract: As above, move options argument here to allow repeated calls to the method using the same object instance. It might also be better to split out the individual options as arguments so the intent of the method is more clearly defined. e.g.

```ruby
def job_vacancies(query:, postcode:, distance:)
```

smoothcontract: Also, could the query argument be omitted - it feels like that argument is an implementation detail of the DWP API, and when calling the job vacancies method we should be able to just specify the search parameters.

app/services/find_a_job_service.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def health_check
+    JSON.parse(perform_request)
+  end
+
+  private
+
+  def response_body
+    return '{}' unless api_id && api_key
```
smoothcontract: I'm not sure about this behaviour. This is a required service so if no API key is defined we should see an error raised - I think this should raise an exception as per the pattern on the other services. If the service was optional then this implementation would be correct though.

app/services/find_a_job_service.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    perform_request
+  rescue StandardError => e
+    Rails.logger.error("Find a Job Service API error: #{e.inspect}")
+
+    '{}'
+  end
+
+  def perform_request
+    Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https', read_timeout: 120) do |http|
```
smoothcontract: If this timeout is in seconds then it needs to be way more aggressive - maybe a few seconds maximum. We don't want page requests to block for this long waiting on the DWP API.

app/services/find_a_job_service.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def job_vacancies
+    JSON
+      .parse(response_body)
+      .dig('pager', 'total_entries')
+  end
+
+  def health_check
+    JSON.parse(perform_request)
```
smoothcontract: The health_check method should automatically specify the `path: 'ping` options when invoking the API call. It shouldn't be the responsibility of the user of this service to understand the path option.

app/views/job_profiles/_jobs_in_your_area.html.erb
```diff
@@ -0,0 +1,8 @@
+<% return unless @job_vacancies.present? %>
+<h2 class="govuk-heading-m">Jobs in your area</h2>
+<% if @job_vacancies > 0 %>
```
smoothcontract: Maybe change to `unless @job_vacancies.zero?`

app/controllers/job_profiles_controller.rb
```diff
@@ -1,6 +1,7 @@
 class JobProfilesController < ApplicationController
   def show
     @job_profile = JobProfileDecorator.new(resource)
+    @job_vacancies = find_a_job_service.job_vacancies if user_session.postcode.present?
```
smoothcontract: I think this is returning a count of job vacancies, rather than a list of vacancies? If so the naming could be more explicit e.g. `@job_vacancy_count`

app/views/job_profiles/_jobs_in_your_area.html.erb
smoothcontract: Also does this throw an error when `@job_vacancies` is nil? (i.e. when the session does not have a postcode)

azure/config/dev3.parameters.json
```diff
@@ -72,6 +72,22 @@
         "secretName": "dev-notifyApiKey"
       }
     },
+    "findAJobApiId": {
```
smoothcontract: Please see https://github.com/DFE-Digital/get-help-to-retrain/pull/239 - I'd prefer to remove this file entirely.

app/services/find_a_job_service.rb
cwrw: the query is the job profile name, I'll change that argument name if that is confusing

app/views/job_profiles/_jobs_in_your_area.html.erb
cwrw: we return before this on line 1, so it will never get to this stage at `nil`

app/controllers/job_profiles_controller.rb
```diff
@@ -28,4 +29,14 @@ def query_params
...
   end
+
+  def find_a_job_service
+    FindAJobService.new(
+      options: {
+        query: resource.name,
+        postcode: user_session.postcode,
+        distance: 20
+      }
+    )
```
tavy88: I think you should be able to create one object instance e.g:

```ruby
service = Service.new
service.call(*args)
```

This way you can reuse the same object. With your approach, you need to create a new instance every time you want to do a query and I think it can be avoided. I do something similar in both Notify & SpellChecker.

app/views/job_profiles/_jobs_in_your_area.html.erb
cwrw: hmm regarding the zero check, I'd rather keep this, as I want to be explicit about being *more* than zero vs is it zero or something else

app/services/find_a_job_service.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def query_values
+    {
+      api_id: api_id,
+      api_key: api_key,
+      q: query,
+      w: outcode,
+      d: distance
+    }.reject { |_k, v| v.nil? }
```
tavy88: Wonder if it's better `{ |_k, v| v.present? }`. Not sure what happens if postcode is blank, or query is blank etc.

spec/services/find_a_job_service_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+
+    it 'does not perform query if there is no api id' do
+      service = described_class.new(
+        api_id: nil,
+        api_key: 'test',
+        options: { query: 'developer', postcode: 'NW11 8QE', distance: 20 }
+      )
+
+      service.job_vacancies
+      expect(a_request(:get, /findajob/)).not_to have_been_made
```
tavy88: The reason I keep raising exceptions when a service is missing the API keys is for debugging purposes. Returning nil is not obvious what the problem is with the service.

Since every API we use needs the key to work, we should always reflect that in our services/wrappers I think. It makes life so much easier when debugging.

spec/services/find_a_job_service_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+    it 'raises error if the query is not authorised' do
+      service = described_class.new(
+        api_id: 'wrong-id',
+        api_key: 'wrong-id',
+        options: { path: 'ping' }
+      )
+
+      stub_request(:get, /findajob/)
+        .to_return(status: 401)
+
```
tavy88: You could avoid repeating this, by putting it in a `before` block, and only customise the status at every step you need. E.g: let(:status) { 401 }.

```ruby
before do
  stub_request(:get, /findajob/)
        .to_return(status: status)
end

it 'does soemthing' do
let(:status) { 401 }
end
```

app/services/find_a_job_service.rb
cwrw: oops, copy pasta :D

app/services/find_a_job_service.rb
cwrw: yea that's definitely a valid point, I'll refactor 👍

app/services/find_a_job_service.rb
smoothcontract: It wasn't immediately obvious, so renaming to `name` or similar might be better.

app/services/find_a_job_service.rb
smoothcontract: Discussed offline and agreed to leave this as is.

app/services/find_a_job_service.rb
smoothcontract: Interesting thought

app/views/job_profiles/_jobs_in_your_area.html.erb
smoothcontract: Okay 👍

spec/services/find_a_job_service_spec.rb
cwrw: I don't think that change matches the style of tests here or helps in readability/understanding, as the before block means I need to add contexts to all expectations returning statuses and set the let per status adding too much noise. I prefer self contained tests, and if something is quite large like a response and reused constantly then it's fair to define it using a let

spec/services/find_a_job_service_spec.rb
cwrw: discussed as a group, and agreed that each API integration is different, and that we are going to keep the implementation as so

app/services/find_a_job_service.rb
cwrw: changed to blank which is the same as present for reject

app/services/find_a_job_service.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def initialize(
+    api_id: Rails.configuration.find_a_job_api_id,
+    api_key: Rails.configuration.find_a_job_api_key
+  )
+    @api_id = api_id
+    @api_key = api_key
+  end
+
+  def job_vacancy_count(options:)
```
smoothcontract: Probably not much point making this a named argument now. Just a minor nitpick though, not essential.

spec/features/job_profile_spec.rb
```diff
@@ -60,4 +60,64 @@
 
     expect(page).not_to have_content('Apprenticeship')
   end
+
+  scenario 'User can see number of job vacancies near them' do
+    find_a_job_service = instance_double(FindAJobService, job_vacancy_count: 3)
```
smoothcontract: Nice 👍

spec/features/job_profile_spec.rb
```diff
@@ -60,4 +60,64 @@
...
+    allow(FindAJobService).to receive(:new).and_return(find_a_job_service)
+    job_profile = create(:job_profile, :with_html_content)
+
+    user_enters_location
+    visit(job_profile_path(job_profile.slug))
+
+    expect(page).to have_content('there are no jobs')
+  end
+
+  scenario 'User does not see job vacancies is API is down' do
```
smoothcontract: `if` API is down

PR #240, opened 2019-10-09
Add pagination to courses page

tavy88:
Show 10 results per page on the courses page.

Screen Shot 2019-10-09 at 09 27 58

https://dfedigital.atlassian.net/browse/GET-369

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -6,7 +6,8 @@ class CoursesController < ApplicationController
       postcode: postcode,
       topic: courses_params[:topic_id]
     )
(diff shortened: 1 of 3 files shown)


app/controllers/courses_controller.rb
```diff
@@ -20,4 +21,8 @@ def postcode
   def courses_params
     params.permit(:postcode, :topic_id)
   end
+
+  def course_search
+    @course_search = @search.find_courses.map { |c| CourseDecorator.new(c) }
```
smoothcontract: I think the instance variable is redundant here, no need to assign that.

tavy88: Fair - updated!

PR #242, opened 2019-10-09
Add privacy policy page

tavy88:
### Description

Add privacy policy page. Also, link to it from PID.

https://dfedigital.atlassian.net/browse/GET-462

### How to reach

`/privacy-policy` or from PID form.

### DIFF
--- a/app/views/pages/privacy_policy.html.erb
+++ b/app/views/pages/privacy_policy.html.erb
@@ -53,9 +53,9 @@
       <li>Or by post at Emma Wharram, Departmental Data Protection Officer, 2 Rivergate, Bristol, BS1 6EW</li>
     </ul>
     <h2 class="govuk-heading-m govuk-!-margin-top-6">How to find out what personal information we hold about you</h2>
-    <p class="govuk-body">To find out what personal information we hold about you, or to exercise any of your rights you can contact us using the question option on our <%= link_to('Contact Form', ' https://form.education.gov.uk/en/AchieveForms/?form_uri=sandbox-publish://AF-Process-f1453496-7d8a-463f-9f33-1da2ac47ed76/AF-Stage-1e64d4cc-25fb-499a-a8d7-74e98203ac00/definition.json&redirectlink=%2Fen&cancelRedirectLink=%2Fen', target: '_blank') %> and select ‘something else’.</p>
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -13,6 +13,7 @@
 
   get 'task-list', to: 'pages#task_list'
   get 'next-steps', to: 'pages#next_steps'
+  get 'privacy-policy', to: 'pages#privacy_policy'
```
cwrw: should prob have this under the PID flag

app/views/pages/privacy_policy.html.erb
```diff
@@ -0,0 +1,61 @@
...
+    <h2 class="govuk-heading-m">Last updated</h2>
+    <p class="govuk-body">This version was last updated on 04 October 2019.</p>
+    <h2 class="govuk-heading-m">Contact us</h2>
+    <p class="govuk-body">If you have any questions about how your personal information will be processed, contact the Data Protection Officer (DPO) by:</p>
+    <ul>
+      <li>Email at <%= mail_to('advice.dpo@education.gov.uk') %></li>
+      <li>Or by post at Emma Wharram, Departmental Data Protection Officer, 2 Rivergate, Bristol, BS1 6EW</li>
+    </ul>
+    <h2 class="govuk-heading-m govuk-!-margin-top-6">How to find out what personal information we hold about you</h2>
+    <p class="govuk-body">To find out what personal information we hold about you, or to exercise any of your rights you can contact us using the question option on our <%= link_to('Contact Form', ' https://form.education.gov.uk/en/AchieveForms/?form_uri=sandbox-publish://AF-Process-f1453496-7d8a-463f-9f33-1da2ac47ed76/AF-Stage-1e64d4cc-25fb-499a-a8d7-74e98203ac00/definition.json&redirectlink=%2Fen&cancelRedirectLink=%2Fen', target: '_blank') %> and select ‘something else’.</p>
```
smoothcontract: Additional space at the start of the link url - also is that the shortest link that would work?

app/views/pages/privacy_policy.html.erb
```diff
@@ -0,0 +1,61 @@
...
+    <p class="govuk-body">Please be as specific as you can about the personal information that you are asking us about. If possible, state which part of the department is holding the information and why they hold it.</p>
+    <p class="govuk-body">If we cannot satisfactorily confirm or establish your identity, we will request for proof of your identity so we can check your right to access the information and to ensure that we only disclose your personal information to the right person(s).</p>
+    <p>We should respond to your request within one month of receiving it. If your request is complex, we may extend the period by a further two months. In those circumstances, we will let you know of the extended time within one month of receiving your request and explain to you why the extension is necessary.</p>
```
smoothcontract: Missing class on the `<p>` tag

config/routes.rb
tavy88: My understanding is that this needs to be accessible when we release but the PID will be off. Will check with PO guys tomorrow where this is accessed from.

smoothcontract: I don't think there's a need to feature flag this - it's currently only linked from the personal info page which itself is already feature flagged. If someone guesses the url and visits the page directly then that's not a big deal really.

PR #243, opened 2019-10-10
GET-538 Do not destroy session if versions are different

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-538

When users were hitting the page for the first time we were deleting
the session from the database as they didn't have a version in the data.
This was causing unnecessary read/write to the database also causing the
database to seem like it was "skipping" ids, which were records created and
deleted unnecessarily.

This also caused an issue with authenticated users linked to sessions. If a user was signed
in with a session and the version changed, which caused their session to be deleted they could
not log in again, but rather see an error where we were trying to access a non existing session.

Now we "clear" the session instead of deleting it, only if the session has data, meaning there are no
unnecessary writes to the db, and we don't risk deleting a session linked to a user.

However I am unhappy with this versioning solution, as after we persist sessions this
fix seems more of a pain than a solution. We now have control over sessions, and can make
sure to be backward compatible when releasing different versions. I propose after going live
with phase 3 we get rid of versioning, as it is needed for this release only. I added a commented
out test that shows another failure this causes and as a reminder for us to remove after the release.

### DIFF
--- a/app/models/user_session.rb
+++ b/app/models/user_session.rb
@@ -17,7 +17,9 @@ class UserSession
 
   def initialize(session)
     @session = session
-    @session.destroy unless version == expected_version
+
+    # TODO: This should be removed after we go live with skills builder v2
+    @session.clear unless version == expected_version
 
     @session[:visited_pages] ||= []
     @session[:job_profile_skills] ||= {}
--- a/spec/features/user_sign_in_spec.rb
+++ b/spec/features/user_sign_in_spec.rb
@@ -344,4 +344,39 @@ RSpec.feature 'User sign in' do
 
     expect(page).not_to have_text(/Enter a job title/)
   end
+
+  scenario 'user does not lose session if the session version changes' do
+    disable_feature!(:skills_builder_v2)
+    enable_feature!(:user_authentication)
+
+    register_user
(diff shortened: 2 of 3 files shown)


app/models/user_session.rb
```diff
@@ -17,7 +17,9 @@ def self.merge_sessions(new_session:, previous_session_data:)
 
   def initialize(session)
     @session = session
-    @session.destroy unless version == expected_version
+
+    # This should be removed after we go live with skills builder v2
```
smoothcontract: Maybe add a `FIXME:` annotation?

spec/features/user_sign_in_spec.rb
```diff
@@ -344,4 +344,39 @@ def sign_in_user(reset_sessions: true)
...
+    visit(check_your_skills_path)
+    Capybara.reset_sessions!
+
+    sign_in_user
+    visit(skills_path)
+
+    expect(page).not_to have_selector('tbody tr', count: 2)
+  end
+
+  # Added to show what would happen if we keep the session clear option, which we
```
smoothcontract: This comment could be moved to a `pending` line within the test body

PR #244, opened 2019-10-10
Refractor deployment into a template for re-use

mafonso:
### Context

Refractor deployment into a template for re-use

This PR just moves the deployment logic into a template, so that it DRYs the deployment to multiple environments when needed. 

Some code cleaning before adding the test stages.

### DIFF
--- a/azure-pipelines-deploy-template.yml
+++ b/azure-pipelines-deploy-template.yml
@@ -1,0 +1,65 @@
+parameters:
+  debug: false
+  subscriptionServiceConnection:
+  resourceGroupName:
+  environment:
+  containerImageReference:
+  supportEmailAddresses: '[{"name":"GHtR Alerts","email":"ghtr-alerts@digital.education.gov.uk"}]'
+
+jobs:
+  - deployment: deploy_${{parameters.environment}}
+    displayName: 'Deploy App to ${{parameters.environment}}'
+    environment: '${{parameters.environment}}'
+
+    pool:
+      vmImage: 'vs2017-win2016'
+
+    strategy:
+      runOnce:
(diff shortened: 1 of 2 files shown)


azure-pipelines-deploy-template.yml
```diff
@@ -0,0 +1,65 @@
+parameters:
+  debug: false
+  subscriptionServiceConnection:
```
cwrw: do those need to be here empty as they are inherited from the azurre-pipeline.yml?

mafonso: We do, because they act as input arguments with optional defaults. 
They need to be declared.

azure-pipelines-deploy-template.yml
```diff
@@ -0,0 +1,65 @@
...
+        deploy:
+          steps:
+            - task: AzureResourceGroupDeployment@2
+              displayName: 'Azure Deployment:Create Or Update Resource Group action on $(resourceEnvironmentName)-$(serviceName)'
+              inputs:
+                action: 'Create Or Update Resource Group'
+                location: 'West Europe'
+                azureSubscription: '${{parameters.subscriptionServiceConnection}}'
+                resourceGroupName: '${{parameters.resourceGroupName}}'
+                csmFile: '$(Pipeline.Workspace)\ARMTemplates\azure\template.json'
```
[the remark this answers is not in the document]
mafonso: `$(Pipeline.Workspace)`?   yes it is. The reason I need to resort to it, although being a good practice because it can change outside our control, is that in this job we are using a Windows box, so it will be something like `D:\something` rather then ´/home/something`

PR #245, opened 2019-10-10
Add cookies policy page

smoothcontract:
### Context
Adds new cookies policy page and links to both that and privacy policy page in the footer.
Draft for now - agreeing content first on Heroku review app.

Designs for footer are based on those found in the heroku prototype.
Also found an existing screen reader CSS class while working on this so cleaned up the custom `sr-only` class we recently added.

### Guidance to review
Content is signed off now :)

### DIFF
--- a/app/views/check_your_skills/_results_list.html.erb
+++ b/app/views/check_your_skills/_results_list.html.erb
@@ -5,7 +5,7 @@
         <%= link_to job_profile.name, job_profile_skills_path(job_profile.slug, search: params[:search]), class: 'govuk-link no-text-decoration' %>
       </h3>
       <% if job_profile.alternative_titles.present? %>
-        <span class="sr-only">Alternative titles for this job include:</span>
+        <span class="govuk-visually-hidden">Alternative titles for this job include:</span>
         <p class="govuk-body-s govuk-!-margin-bottom-2 muted-text"><%= job_profile.alternative_titles %></p>
       <% end %>
       <p class="govuk-!-margin-0"><%= job_profile.description %></p>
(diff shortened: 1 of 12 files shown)


app/views/check_your_skills/_results_list.html.erb
```diff
@@ -5,7 +5,7 @@
         <%= link_to job_profile.name, job_profile_skills_path(job_profile.slug, search: params[:search]), class: 'govuk-link no-text-decoration' %>
       </h3>
       <% if job_profile.alternative_titles.present? %>
-        <span class="sr-only">Alternative titles for this job include:</span>
+        <span class="govuk-visually-hidden">Alternative titles for this job include:</span>
```
cwrw: 👍

app/views/pages/cookies_policy.html.erb
```diff
@@ -0,0 +1,149 @@
...
+    <table class="govuk-table">
+      <caption class="govuk-table__caption small govuk-visually-hidden">Cookie explanation</caption>
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th class="govuk-table__header" scope="col">Name</th>
+          <th class="govuk-table__header" scope="col">Purpose</th>
+          <th class="govuk-table__header" scope="col">Expires</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
```
cwrw: are the bold lines intentional on tables, as I can see here: https://design-system.service.gov.uk/components/table/ it doesn't have a bold last line?

smoothcontract: This was due to a custom rule we had in place for the skills builder table. I've renamed the class for that to be more specific to skills, so that should resolve this.

PR #246, opened 2019-10-11
GET-571 Remove Location Eligibility page

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-571

* Remove PID feature flag (but keen in split until this goes live)
* Remove location eligibility page
* Add geolocation logic to PID
* Remove skip link and unneeded content

Testing: check tracking on azure


app/controllers/user_personal_data_controller.rb
```diff
@@ -11,18 +11,12 @@ def create
     if @user_personal_data.save
       user_session.pid = true
```
smoothcontract: I think you can remove this line. The pid_submitted? method in user session can now check that a postcode is present instead rather than adding an explicit boolean flag.

app/helpers/user_journey_helper.rb
```diff
@@ -1,11 +1,5 @@
 module UserJourneyHelper
   def pid_step
-    return unless Flipflop.user_personal_data?
-
     your_information_path unless user_session.pid_submitted?
   end
-
-  def postcode_step
-    location_eligibility_path unless user_session.postcode.present?
-  end
```
smoothcontract: Is this module now only included in one place? If so could move the remaining method to relevant controller and delete the helper.

app/controllers/home_controller.rb
```diff
@@ -1,5 +1,5 @@
 class HomeController < ApplicationController
   def index
-    @postcode_pid_or_task_list_path = helpers.postcode_step || helpers.pid_step || task_list_path
+    @pid_or_task_list_path = user_session.postcode.present? ? task_list_path : your_information_path
```
tavy88: Ideally, I would like if we could have `user_session.postcode?` so have a method defined in the UserSession class.

app/controllers/user_personal_data_controller.rb
```diff
@@ -1,28 +1,20 @@
 class UserPersonalDataController < ApplicationController
   def index
-    redirect_to task_list_path if user_session.pid_submitted?
+    redirect_to task_list_path if user_session.postcode.present?
```
tavy88: Ideally, I would like if we could have user_session.postcode? so have a method defined in the UserSession class.

cwrw: It feels a bit overkill, we already have a getter/setter, and using a `present?` normally should be ok as there isn't any strange logic to check if a postcode exists. We added other methods for `pid` and `registered` as we didn't have getters for them and they were explicitly there to check for boolean values

tavy88: Fair it's just the ruby convention is ? at the end of a method that returns boolean. But as I said no biggie.

PR #247, opened 2019-10-11
Add cookie banner

smoothcontract:
### Context
Adds a modal cookie banner as per Jira ticket: https://dfedigital.atlassian.net/browse/GET-175

Adds a new partial and JS event handler. Modal is shown on every page except for the cookies policy page. If user accepts the policy a cookie is set with 30 day expiry to prevent them seeing the modal again.

### Guidance to review
New cookie is named `seen_cookie_message` - delete that in the browser to test.


app/views/layouts/application.html.erb
```diff
@@ -25,21 +25,10 @@
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     </script>
     <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
+    <%= render 'shared/cookies_banner' unless current_page?(cookies_policy_path) %>
     <header class="govuk-header" role="banner" data-module="header">
       <div class="govuk-header__container govuk-width-container">
-        <div class="govuk-header__logo">
```
smoothcontract: I moved this into a partial because I couldn't see a nice approach for moving the modal backdrop down below the header, and keep the header from scrolling. So this header part is included twice on the page - once in the layout and again in the modal. Not ideal but I guess is not a huge problem.

app/webpacker/packs/cookies-banner.js
```diff
@@ -0,0 +1,44 @@
...
+    var cookiesModal = document.querySelector('#cookies-banner');
+
+    if (typeof(cookiesModal) !== 'undefined' && cookiesModal != null) {
+      if (getCookie('seen_cookie_message') !== 'true') {
+        displayCookiesModal(cookiesModal);
+      }
+    }
+  }
+
+  function getCookie(name) {
```
smoothcontract: Stolen from https://plainjs.com/javascript/utilities/set-cookie-get-cookie-and-delete-cookie-5/ - not sure if we already have utility functions for stuff like this within the toolkit.

app/views/layouts/application.html.erb
cwrw: Maybe we can move the whole header into a partial, this way we dry it up more? 
```
    <header class="govuk-header" role="banner" data-module="header">
      <div class="govuk-header__container govuk-width-container govuk-!-padding-bottom-2">
        <%= render 'shared/navigation/header_logo' %>
        <%= render 'shared/navigation/main_header' %>
      </div>
    </header>
```
Although it will be repeated twice atleast the whole thing will be a in a partial this time

spec/features/home_spec.rb
```diff
@@ -103,6 +103,14 @@
     expect(page).to have_current_path(task_list_path)
   end
 
+  scenario 'User can access the cookies policy page from the cookies modal', :js do
+    visit(root_path)
+
+    click_on('Cookie settings')
+
+    expect(page).to have_text('Cookies are used')
+  end
```
cwrw: can we  also add a test for clicking accept cookies and the modal goes away?

spec/features/home_spec.rb
```diff
@@ -103,6 +103,14 @@
     expect(page).to have_current_path(task_list_path)
   end
 
+  scenario 'User can access the cookies policy page from the cookies modal', :js do
+    visit(root_path)
+
+    click_on('Cookie settings')
+
+    expect(page).to have_text('Cookies are used')
```
tavy88: I would check that you have navigated to the correct location here:

```ruby
expect(page).to have_current_path(/cookies-policy-path)
```

spec/features/home_spec.rb
cwrw: And maybe one that checks this doesn't appear on cookie policy page itself, but that might be overkill

app/webpacker/styles/_cookies.scss
```diff
@@ -0,0 +1,20 @@
+.cookies-modal {
+  display: none;
+  position: fixed;
+  z-index: 1;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  overflow: auto;
+  background-color: rgb(0,0,0);
```
cwrw: There should be spaces after comma liek `rgb(0, 0, 0)`, same below

app/views/shared/_cookies_banner.html.erb
```diff
@@ -0,0 +1,18 @@
...
+        <%= render 'shared/navigation/header_logo' %>
+        <%= render 'shared/navigation/main_header' %>
+      </div>
+    </header>
+
+    <div class="govuk-width-container">
+      <main class="govuk-main-wrapper govuk-!-padding-top-1 govuk-!-padding-bottom-1" role="main">
+        <p class="govuk-body">This service uses cookies to personalise your experience and to collect information about how you use the site. You can find out more about cookies by clicking the cookies settings button.</p>
+        <%= link_to 'Accept cookies', '#', id: 'accept-cookies', class: 'govuk-button govuk-button--secondary govuk-!-margin-right-5' %>
+        <%= link_to 'Cookie settings', cookies_policy_path, class: 'govuk-button govuk-button--secondary ' %>
```
tavy88: Extra space in `govuk-button govuk-button--secondary` class.

app/views/layouts/application.html.erb
smoothcontract: The wrapper markup is slightly different (an extra class is added). I initially had the whole header shared but had to split this out to allow different markup.

spec/features/home_spec.rb
smoothcontract: Not sure that’s necessary really. We don’t have that pattern in the specs for e.g. privacy policy page and the content only appears on the one correct page.

app/webpacker/styles/_cookies.scss
```diff
@@ -0,0 +1,20 @@
...
+  background-color: rgba(0,0,0,0.6);
```
cwrw: gov uk linter isn't happy with the double background colour unfortunately even though its a fallback from what I can see here: https://www.w3schools.com/howto/howto_css_modals.asp

I think maybe if we address the variable issue the linter is flagging up then see if it still isn't happy, we can remove the fallback

app/webpacker/packs/cookies-banner.js
```diff
@@ -0,0 +1,44 @@
...
+    var v = document.cookie.match('(^|;) ?' + name + '=([^;]*)(;|$)');
+
+    return v ? v[2] : null;
+  }
+
+  function setCookie(name, value, days) {
+    var date = new Date;
```
tavy88: New line after?

app/webpacker/packs/cookies-banner.js
cwrw: I couldn't find anything for cookies, and on their backlog it says to do, so I'm happy to keep our own implementation for now till we move to a more permanent solution

smoothcontract: okay.

spec/features/home_spec.rb
smoothcontract: Why would anyone add "Cookies are used" on a different page? 😂 
I sort of take your point, but I can only add one expectation so it seems fine to me to check the content for the page this lands on. That also checks that the destination page renders correctly as well as just being the right url - so actually I think checking content is better. @cwrw thoughts?

tavy88: It could be a footer e.g note: 'Cookies are used on this site.' It's not impossible. And I was not suggesting to add an extra expectation, just replace the have_text one with have_current_path. But in our case, it doesn't matter that much. Not precious about it.

cwrw: So we've got a mix of specs with text checking vs current path. I think from my side I wouldn't make any rules regarding  either as it's a per spec requirement, as some things can be duplicate across different pages and text is a better test while some are on only one page so the path is more relevant. In this case I'm leaning towards the `have_current_path(cookies_policy_path)` as we want to make sure we are on the right cookie page.

PR #256, opened 2019-10-16
Update job profile page content

tavy88:
Changes:

Remove Valuable skills section and training CTA
Remove Further help to change jobs section and CTA
Add CTA at the bottom that won’t actually go anywhere until
the questions pages are built

The feature is flagged.

Screen Shot 2019-10-17 at 08 38 05

https://dfedigital.atlassian.net/browse/GET-597


app/views/job_profiles/v2/_relevant_sections.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<%= @job_profile.section(xpath: JobProfileDecorator::WWYD_TASKS_SECTION_XPATH).html_safe %>
+<%= @job_profile.section(xpath: JobProfileDecorator::WWYD_WORK_SECTION_XPATH).html_safe %>
+<%= @job_profile.section(xpath: JobProfileDecorator::DIRECT_APPLICATION_SECTION_XPATH).html_safe %>
+<%= render 'apprenticeship' %>
+<%= @job_profile.section(xpath: JobProfileDecorator::CAREER_PATH_SECTION_XPATH).html_safe %>
+<%= @job_profile.section(xpath: JobProfileDecorator::OTHER_ROUTES_SECTION_XPATH).html_safe %>
+<%= @job_profile.section(xpath: JobProfileDecorator::CAREER_TIPS_SECTION_XPATH).html_safe %>
+<%= @job_profile.section(xpath: JobProfileDecorator::WORK_SECTION_XPATH).html_safe %>
+<%= @job_profile.section(xpath: JobProfileDecorator::RESTRICTIONS_AND_REQUESTS_SECTION_XPATH).html_safe %>
+<%= render 'jobs_in_your_area' %>
```
smoothcontract: Hmm.. I wonder if this is the best place to split the two different views. Looking at `_relevant_sections.html.erb` v1 this is the list of elements currently:

```ruby
<%= @job_profile.section(xpath: JobProfileDecorator::SKILLS_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::WWYD_TASKS_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::WWYD_WORK_SECTION_XPATH).html_safe %>
<%= render 'valuable_skills' %>
<%= @job_profile.section(xpath: JobProfileDecorator::DIRECT_APPLICATION_SECTION_XPATH).html_safe %>
<%= render 'apprenticeship' %>
<%= @job_profile.section(xpath: JobProfileDecorator::CAREER_PATH_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::OTHER_ROUTES_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::CAREER_TIPS_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::WORK_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::RESTRICTIONS_AND_REQUESTS_SECTION_XPATH).html_safe %>
<%= render 'jobs_in_your_area' %>
<%= render 'further_help' %>
```

smoothcontract: I know during planning we said this should use a v1 and v2 job profile view, but it seems the similarities are too great and it feels like overkill to split this view into two. I'd suggest instead retaining one version of this view as follows:

```ruby
<%= @job_profile.section(xpath: JobProfileDecorator::SKILLS_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::WWYD_TASKS_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::WWYD_WORK_SECTION_XPATH).html_safe %>
<%= render 'valuable_skills' unless Flipflop.job_profile_v2? %>
<%= @job_profile.section(xpath: JobProfileDecorator::DIRECT_APPLICATION_SECTION_XPATH).html_safe %>
<%= render 'apprenticeship' %>
<%= @job_profile.section(xpath: JobProfileDecorator::CAREER_PATH_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::OTHER_ROUTES_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::CAREER_TIPS_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::WORK_SECTION_XPATH).html_safe %>
<%= @job_profile.section(xpath: JobProfileDecorator::RESTRICTIONS_AND_REQUESTS_SECTION_XPATH).html_safe %>
<%= render 'jobs_in_your_area' %>
<%= render 'further_help' unless Flipflop.job_profile_v2? %>
```

This feels like a less invasive change and is easier to unpick when the feature flags are removed later (just delete the two relevant lines and associated views).

tavy88: The implementation was done according to what we discussed in planning and what the AC of the ticket says :)

Ideally, I would like to stick to that unless you have strong reasons to push for your approach. I find the separation cleaner and less headache as we remove the full file version when we do the cleanup.

spec/features/job_profile_spec.rb
```diff
@@ -115,6 +127,52 @@
...
+    enable_feature! :job_profile_v2
+
+    job_profile = create(:job_profile, :with_html_content)
+
+    visit(job_profile_path(job_profile.slug))
+
+    expect(page).not_to have_content('Further help to change jobs')
+  end
+
+  scenario 'User can see number of job vacancies near them with CTA: Target this job on V2' do
```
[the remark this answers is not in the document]
tavy88: So the initial requirement was the button to depend on the jobs section (which has different content depending on whether the jobs are available or not. However, as per the last discussion with Lois, she wants the button present regardless of the section.

app/views/job_profiles/_jobs_in_your_area.html.erb
```diff
@@ -5,4 +5,6 @@
 <% else %>
   <p class="govuk-body">At present, there are no jobs like this one being advertised near you on the Find a Job service. You may be able to find other jobs like this advertised on other job sites.</p>
 <% end %>
-<hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+<% if Flipflop.job_profile_v2? %>
```
cwrw: I think this should be unless rather than if, but I think it might be better moving this to further help this way we don't need to do this check here

tavy88: Let's discuss on a call. It's easier.

app/views/job_profiles/_jobs_in_your_area.html.erb
```diff
@@ -6,3 +6,4 @@
   <p class="govuk-body">At present, there are no jobs like this one being advertised near you on the Find a Job service. You may be able to find other jobs like this advertised on other job sites.</p>
 <% end %>
 <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
```
cwrw: 🦂

app/views/job_profiles/v2/_relevant_sections.html.erb
smoothcontract: Thanks @tavy88 and apologies for the incorrect assumption made during planning that a new view would be the best approach.

PR #263, opened 2019-10-17
User authentication sidebar partial update

tavy88:
The following changes are done:

Removed textbox from return to saved results partial
Changed sentence in partial
CTA button takes user to return-to-saved-results page

https://dfedigital.atlassian.net/browse/GET-537


app/views/shared/_return_to_saved_results.html.erb
```diff
@@ -2,12 +2,5 @@
...
   <p class="govuk-body"><%= t('return_to_saved_results.sub_title') %></p>
-  <%= form_with local: true, url: sign_in_path do |form| %>
-    <%= form_group_tag user_with_validation, :email do %>
-      <%= form.label(:email, class: 'govuk-label') %>
-      <%= errors_tag user_with_validation, :email %>
-      <%= form.text_field(:email, value: params[:email], class: css_classes_for_input(user_with_validation, :email, " govuk-!-width-full")) %>
-    <% end %>
-    <%= button_tag('Send a link', class: 'govuk-button govuk-button--secondary govuk-!-width-full govuk-!-margin-top-1 govuk-!-margin-bottom-2') %>
-  <% end %>
+  <%= link_to(t('return_to_saved_results.title'), return_to_saved_results_path, class: 'govuk-button govuk-button--secondary govuk-!-width-full govuk-!-margin-bottom-2') %>
```
cwrw: since the margin is the same as the other save your results partial, can we remove that from there as well?

config/locales/en.yml
```diff
@@ -142,7 +142,7 @@ en-GB:
       title: Link sent again
   return_to_saved_results:
     title: Return to saved results
-    sub_title: If you've already used this service before and saved your results, enter your email address. We'll send you a link you can use to return to your saved results.
+    sub_title: Return to any results you’ve previously saved in the service.
```
cwrw: can we use `'` instead of `’`

spec/features/user_authentication_sidebar_spec.rb
```diff
@@ -38,8 +38,6 @@
   def unlock_tasklist_steps(job_profile: job_profile1)
     visit(job_profile_skills_path(job_profile_id: job_profile.slug))
     click_on('Select these skills')
-
-    click_on('Find out what you can do with these skills')
```
cwrw: 👍

app/views/shared/_return_to_saved_results.html.erb
tavy88: Sorry not sure I follow. Remove from where and which margin?

config/locales/en.yml
tavy88: We have a mixture as you know. It won't make a difference. I would keep as is.

spec/features/user_authentication_sidebar_spec.rb
```diff
@@ -112,6 +110,14 @@ def register_user
       end
     end
 
+    scenario 'user gets to Return to saved results page when clicking Return to saved results button' do
```
cwrw: can we move this test to the sign in spec, as this spec file is for checking the partials rather than where they lead to

spec/features/user_sign_in_spec.rb
```diff
@@ -156,25 +114,13 @@ def sign_in_user(reset_sessions: true)
...
 
-  scenario 'User does not receives sign in email if email valid and user does not exists' do
+  scenario 'User does not receive sign in email if email valid and user does not exists' do
     send_sign_in_email
 
     expect(client).not_to have_received(:send_email)
       .with(sign_in_email)
   end
 
-  scenario 'User receives sign in email if they enter their same email but different case' do
```
cwrw: can we keep this test but change the visit to return to saved results?

spec/features/user_authentication_sidebar_spec.rb
tavy88: 👍

app/controllers/users_controller.rb
```diff
@@ -4,6 +4,14 @@ def new
     set_redirect_path_for_registration
   end
 
+  def return_to_saved_results
+    if params.key?(:email)
```
cwrw: ah I think you missed the place that handles the validation which is here: https://github.com/DFE-Digital/get-help-to-retrain/blob/master/app/controllers/users_controller.rb#L59

So instead this method remains as so: 
```
def return_to_saved_results
  @user = User.new
end
```
and we now render the partial instead on this line: https://github.com/DFE-Digital/get-help-to-retrain/blob/master/app/controllers/users_controller.rb#L64
so the method looks like this in the end: 
```
def sign_in_with(path: nil, partial: nil)
    if user.valid?
      sign_in_user unless user.new_record?
      path ? redirect_to(path) : render(partial)
    else
      render(:return_to_saved_results)
    end
  rescue NotifyService::NotifyAPIError
    # TODO: show user an error page, for now render link sent page
    path ? redirect_to(path) : render(partial)
  end
```
This way we can get rid of the last thing regarding the flash message and reduce the complexity, as well as clean up that method.

You can also get rid of this method as it's no longer used after that: https://github.com/DFE-Digital/get-help-to-retrain/blob/master/app/controllers/users_controller.rb#L48

tavy88: I didn't miss it I just didn't want to mess around that method as it's being used by several methods.

PR #265, opened 2019-10-17
Allow job profile to be targeted and shown in action plan

smoothcontract:
### Context
Allow job profile to be targeted and shown in action plan

### Guidance to review
WIP, but this is working so would welcome any initial feedback. Test coverage will be added in subsequent commits.


app/controllers/job_profiles_controller.rb
```diff
@@ -4,6 +4,12 @@ def show
     @job_vacancy_count = job_vacancy_count if user_session.postcode.present?
   end
 
+  def update
+    user_session.target_job_id = resource.id
+
+    redirect_to action_plan_path
+  end
```
tavy88: I don't think `update` is the right action here as you are not updating a JobProfile record?

app/controllers/application_controller.rb
```diff
@@ -7,12 +7,15 @@ def user_session
...
 
-  # The authenticate method comes from the passwordless gem.
-  #  This is authenticating the passwordless session.
+  # The authenticate method comes from the passwordless gem (note this refers to Passwordless::Session)
   def current_user
     @current_user ||= authenticate_by_session(User)
   end
 
+  def target_job
+    @target_job ||= helpers.target_job
```
tavy88: Where is this helper defined? And does it need to be in the parent? The target job is that persisted job_profile from my understanding? So storing that id on the session at the questions page controller level is enough. Not sure why we need a target_job object?

Also according to the prototype, when the user clicks on Target this job, one lands to the first question page, and I think that's where we should store the id on the session.

app/controllers/job_profiles_controller.rb
cwrw: oh interesting didn't think about it that way, maybe we can rename the method

app/views/job_profiles/_relevant_sections.html.erb
```diff
@@ -12,7 +12,7 @@
 <%= render 'jobs_in_your_area' %>
 <% if Flipflop.action_plan? %>
   <p class="govuk-body">Select this as your target job and we’ll help you prepare to apply.</p>
-  <%= link_to('Target this job', '#', class: 'govuk-button govuk-!-margin-top-1 govuk-!-margin-bottom-6') %>
+  <%= button_to('Target this job', job_profile_path(@job_profile.slug), method: :patch, class: 'govuk-button govuk-!-margin-top-1 govuk-!-margin-bottom-6') %>
```
tavy88: I think this should be built with the controller `TrainingPlanController` and the route should be `training_plan_controller#index` for now as per prototype.

```ruby
 link_to('Target this job', training_plan_path, class: 'govuk-button govuk-!-margin-top-1 govuk-!-margin-bottom-6')
```

app/controllers/job_profiles_controller.rb
smoothcontract: I don't want to introduce a new ActionPlan controller at this stage; that will naturally evolve as the additional features are addressed. As ever, I'm following the simplest implementation possible for the current requirements. That said, it is perhaps pushing the semantic boundary a little to use the update action here - we are doing something with a job profile, so it can sit within the job profiles controller, but it's only selecting rather than updating the job profile. So I'll go with a new name.

app/controllers/application_controller.rb
smoothcontract: The prototype represents the future state of all the Phase 4 features being implemented - we're not building this all at once so for now once a job is targetted the user sees that job in the action plan. In later sprints when the action plan questions are added we'll likely introduce a new controller for the question responses.

The session handling was added in a previous PR to unblock other stories in the sprint: https://github.com/DFE-Digital/get-help-to-retrain/pull/259

If you have a look at that you can see that the id of the targetted job is indeed stored in the session. The method here is a convenience method for use in controllers (will be needed for action plan and job vacancies pages in this sprint) to resolve that session entry to a job profile instance.

app/views/job_profiles/_relevant_sections.html.erb
smoothcontract: See first comment above - it's appropriate to be targetting a single job, so that can be within the job profiles controller for now.

app/controllers/job_profiles_controller.rb
tavy88: Ok but this entire thing will be deleted next sprint really. I was hoping we lay a stronger foundation as we know from the prototype where the Target this job will take you and having that controller now would make things easier next sprint. All your tests and implementation at this level will be removed so not sure if it's worth the effort really. 

All you would need is a simple controller with an index action for now and that would take away all the work you are doing at the JobProfile level, as this logic will not stay here.

smoothcontract: Sorry, I completely disagree. Until we explore the further features of the action plan we won't know how best to implement it. Guessing for future things is hard, whereas it's easier to build the things we know about now. It'll be a matter of moments to **possibly** move this method and associated specs to a new controller **if** that turns out to be the best implementation when we move forward.

Given our current understanding of the action plan, it still feels semantically correct to "target a job profile" - which means an operation involving a job profile, hence an action within the job profiles controller.

tavy88: Since the prototype is https://get-help-to-retrain-prototype.herokuapp.com/main/release4_e2e_v1/training_plan, creating a simple TrainingPlanController with an index action that does this:

```ruby
class ActionPlanController < ApplicationController
  def index
      session[:targetted_job_id] = resource.id

      redirect_to action_plan_path
  end
end
``` 

Is really all that is needed, and you don't need to have any update or similar action at the JobProfile level - which should not be there in the first place anyway. The advantage of this approach is that it offers a strong foundation as next sprint all we have to worry about is evolving this controller.

tavy88: My suggestion is optional and I did present my arguments for it. Up to you.

spec/features/action_plan_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    user_targets_a_job
+
+    # TODO: link this to the DWP job vacancies feature when implemented
+    expect(page).to have_link('Show jobs near me', href: 'https://findajob.dwp.gov.uk/')
+  end
+
+  private
+
+  def user_targets_a_job
+    create(:job_profile, :with_html_content, name: 'Fluffer', slug: 'fluffer').tap do |job_profile|
```
cwrw: 😂

smoothcontract: Shh!

app/controllers/job_profiles_controller.rb
smoothcontract: Your suggestion doesn't quite make sense to me - the `index` action of a training plan controller should probably just display the training plan. So now we're into considering what the name of that action should be if we were to go with that approach. It's easier to let this evolve as our understanding of the action plan develops - and is too early to try now and predict which direction it will go in. Hence, the best semantic fit for the current functionality is within the job profiles controller.

Let's put this to bed for now and revisit when we start to explore the next round of features.

config/routes.rb
```diff
@@ -31,6 +32,9 @@
   end
 
   resources :job_profiles, path: 'job-profiles', only: %i[show destroy] do
+    member do
+      post :target, constraints: ->(_req) { Flipflop.action_plan? }
```
cwrw: since we only have one might be cleaner to write it as so:
`post :target, , on: :member, constraints: ->(_req) { Flipflop.action_plan? }`

spec/routing/action_plan_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe 'routes for action plan', type: :routing do
+  context 'with action_plan feature enabled' do
+    before do
+      enable_feature! :action_plan
+    end
+
+    it 'successfully routes to pages#action_plan' do
+      expect(get('/action-plan')).to route_to(controller: 'pages', action: 'action_plan')
```
tavy88: I think we should use the rails helper method here? `action_plan_path`

spec/routing/action_plan_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+  end
+
+  context 'with action_plan feature disabled' do
+    before do
+      disable_feature! :action_plan
+    end
+
+    it 'does not root to pages#action_plan' do
+      expect(get('/action-plan')).not_to be_routable
```
tavy88: I think we should use the rails helper method here? action_plan_path

app/views/pages/action_plan.html.erb
```diff
@@ -0,0 +1,162 @@
...
+      </div>
+    </div>
+  </div>
+
+  <div class="govuk-grid-column-one-third">
+    <%= render "shared/contact_us" %>
+    <%= render 'shared/save_or_return_to_your_results' %>
+    <%= render 'shared/help_improve_this_service' %>
+  </div>
+</div>
```
tavy88: Think this could be broken is smaller partials per each section in part? Would make it easier to deal with imo.

app/views/pages/action_plan.html.erb
```diff
@@ -0,0 +1,162 @@
...
+        </div>
+      </div>
+
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-one-half">
+          <h3 class="govuk-heading-s">Your cover letter</h3>
+        </div>
+        <div class="govuk-grid-column-one-half">
+          <%= link_to 'Write cover letter', 'https://nationalcareers.service.gov.uk/get-a-job/covering-letter', class: 'govuk-button govuk-button--secondary action-plan-action' %>
+        </div>
```
tavy88: Because this is an external link I think it should open in a new tab? We do that for all external link atm.

app/views/pages/action_plan.html.erb
```diff
@@ -0,0 +1,162 @@
...
+      </div>
+      <p class="govuk-body">Prepare your CV and cover letter and brush up your interview skills so you’re all set to apply for jobs.</p>
+
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-one-half">
+          <h3 class="govuk-heading-s">Your CV</h3>
+        </div>
+        <div class="govuk-grid-column-one-half">
+          <%= link_to 'Update CV', 'https://nationalcareers.service.gov.uk/get-a-job/cv-sections', class: 'govuk-button govuk-button--secondary action-plan-action' %>
```
tavy88: Because this is an external link I think it should open in a new tab? We do that for all external link atm.

app/views/pages/action_plan.html.erb
```diff
@@ -0,0 +1,162 @@
...
+          <%= link_to 'Write cover letter', 'https://nationalcareers.service.gov.uk/get-a-job/covering-letter', class: 'govuk-button govuk-button--secondary action-plan-action' %>
+        </div>
+      </div>
+
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-one-half">
+          <h3 class="govuk-heading-s">Your interview skills</h3>
+        </div>
+        <div class="govuk-grid-column-one-half">
+          <%= link_to 'Prepare for interview', 'https://nationalcareers.service.gov.uk/get-a-job/interview-advice', class: 'govuk-button govuk-button--secondary action-plan-action' %>
```
tavy88: Because this is an external link I think it should open in a new tab? We do that for all external link atm.

spec/routing/action_plan_spec.rb
smoothcontract: This is a routing spec, so the literal path is important - it's testing the routes file. So if I changed the route to be like this, but still used `action_plan_path` above then the routing spec would pass:

```
get 'pants', to: 'pages#action_plan'
```

It makes sense to use url helpers everywhere but in routing specs. Using them within routing specs means the test is incomplete. See: https://relishapp.com/rspec/rspec-rails/docs/routing-specs

spec/routing/action_plan_spec.rb
smoothcontract: as above

app/views/pages/action_plan.html.erb
smoothcontract: This will happen naturally as we add more logic to the page to show/hide each relevant section. No need to do that yet IMO.

app/views/pages/action_plan.html.erb
smoothcontract: There is a new feature planned where external links from the action plan will go via a kind of doorway page that shows "You are leaving this service", so we don't ultimately want to open in a new tab. The idea of the doorway page is to explicitly prevent opening in a new tab, but also show users that they are moving to another service outside GHTR.

The key thing at this point is to have the correct URL really, the behaviour will be enhanced in a later story.

app/controllers/application_controller.rb
```diff
@@ -7,12 +7,15 @@ def user_session
...
-  # The authenticate method comes from the passwordless gem.
-  #  This is authenticating the passwordless session.
+  # The authenticate method comes from the passwordless gem (note this refers to Passwordless::Session)
   def current_user
     @current_user ||= authenticate_by_session(User)
   end
 
+  def target_job
+    helpers.target_job
+  end
```
tavy88: Could this not just be a method defined at the UserSession level?

```ruby
def target_job?
   session[:targetted_job].present?
end
```

Maybe I'm misunderstanding the need for this method here?

app/controllers/job_profiles_controller.rb
tavy88: Ok, the `index` was just a placeholder for now btw. It will not be `index` but for now, it's enough.

cwrw: Seeing how volatile and how quickly things change in this project, especially after user testing sessions, I would lean towards always implementing things gradually and when we need them. As the specs for the new additions have not been confirmed we should hold off but keep them in mind, it much easier to make informed decisions at the time when everything is confirmed.

The important thing is for the current implementation to make it easy for us to transition to the new one if we need it, and I think in this case it will allow us to do that.

app/controllers/application_controller.rb
tavy88: Then you can just call

```ruby
user_session.target_job?
```
anywhere you need to use it?

spec/routing/action_plan_spec.rb
tavy88: So basically you're stating that all of our other routing specs are wrong? 🤔

app/controllers/application_controller.rb
smoothcontract: The target job is needed in a couple of places - both within the controller (to decide whether to redirect or not) and the corresponding view. It doesn't feel like the responsibility of the `UserSession` to resolve to a `JobProfile` instance - that just holds a reference to the id. So the `application_helper` has a method to lookup and cache the job profile:

```ruby
  def target_job
    @target_job ||= JobProfile.find_by(id: user_session.target_job_id)
  end
```

That helper can be (and is) used in the view to render details of the target job. However helper methods are not available in controllers normally. It's possible to include the entire module within a controller but for Rails 5 and above the recommended approach is to use the `helper.` syntax I've followed here.  See: https://blog.bigbinary.com/2016/06/26/rails-add-helpers-method-to-ease-usage-of-helper-modules-in-controllers.html

The alternate approach would be to add a controller method and expose that as a helper. Really not much difference to choose between those two approaches, except that we have an established pattern and code coverage for `application_helper` and very little of that within `ApplicationController`.

Finally, the method is defined at application controller level because it will be needed for multiple controllers - notably also the job vacancies feature in the current sprint.

app/controllers/job_profiles_controller.rb
tavy88: My decision is not random. There will be at least one-page intermediary which will need a new controller. But let's move on for now.

spec/routing/action_plan_spec.rb
smoothcontract: Yep

smoothcontract: Well, there is a mixture of approaches we have now, but best practice is to use literal text strings for URL's within routing specs.

cwrw: So this is interesting, I always push for `action_plan_path` in other specs especially feature ones. Here I wasn't too bothered and saw them as interchangeable as this is a routing spec. After reading the docs it seems the preference is towards explicit specs. I wouldn't call our previous ones wrong, and happy to keep them as is. Also happy to keep this one as is also, as long as we only use it in this context as those are the routing specs and we understand the reasoning behind it being explicit

tavy88: https://relishapp.com/rspec/rspec-rails/v/3-9/docs/routing-specs/named-routes

app/views/pages/action_plan.html.erb
tavy88: I'm not convinced of this argument, I'm afraid. Even if we don't know how things will look like in the feature, I think we can still refactor where we can. The size of this view is probably the biggest we currently have and breaking it down in smaller chunks can easy our lives. There is no unknown here preventing this refactor. Someone will have to do it anyway next sprint most likely.

smoothcontract: I've just pushed another commit which removes a fair bit of markup so this is a little clearer now. In future sprints we'll be adding stories to ask questions about the users needs so this page will evolve to be more dynamic, showing sections only as needed. It will be much clearer when we work on those stories which sections are dynamic, and add the logic at that time.

If we split the page into partials now I'd argue it would actually make the future work harder - in any case it's unlikely that we would guess a half way house implementation now that would be exactly right for the future requirements.

I really want to get this point across - we are not building for future requirements now, just the ones we have in the current sprint. Experience has shown this approach involves the least initial work, least complexity, least rework and best productivity.

spec/routing/action_plan_spec.rb
smoothcontract: After writing this I decided to follow my own advice and simplify the routing specs for job profiles controller - see the last commit. This is now easier to read, more explicit and does not require any job profiles to be created.

PR #268, opened 2019-10-21
GET-605 Job Vacancies Near Me

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-605

* Add new Jobs near me page
* Update API integration to include jobs as well as count while maintaining compatibility with job profiles page vacancy count
* Pagination currently mapped to 50 to match API (commit where it maps 10 included just incase we revert back to it)
* Logic for handling API errors is pending to another ticket. Now we catch errors and return an empty response


app/controllers/job_profiles_controller.rb
```diff
@@ -37,10 +37,9 @@ def query_params
...
 
   def job_vacancy_count
-    FindAJobService.new.job_vacancy_count(
+    JobVacancySearch.new(
       name: resource.name,
-      postcode: user_session.postcode,
-      distance: 20
-    )
+      postcode: user_session.postcode
+    ).count
```
tavy88: I think you can just do 

```ruby
def job_vacancy_count
    @job_vacancy_count = JobVacancySearch.new(...)
end
```

And then you no longer need the assignment `@job_vacancy_count = job_vacancy_count`

app/controllers/job_vacancies_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  def postcode
+    @postcode ||= job_vacancies_params[:postcode] || user_session.postcode
+  end
+
+  def job_vacancies
+    Kaminari
+      .paginate_array(jobs, total_count: @job_vacancy_search.count)
+      .page(job_vacancies_params[:page])
+      .per(50)
+  end
```
tavy88: I think you can just do

```ruby
def job_vacancies
    @job_vacancies = Kaminari.paginate....
end
```

And then you no longer need the assignment `@job_vacancies = job_vacanci`e`s

app/models/job_vacancy.rb
```diff
@@ -0,0 +1,39 @@
...
+    body['location'].presence
+  end
+
+  def salary
+    body['salary'].presence
+  end
+
+  def description
+    body['description']
+      &.truncate(260)
```
tavy88: Could this be a constant?

spec/features/jobs_near_me_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+  end
+
+  scenario 'User gets relevant messaging if no address is entered' do
+    user_targets_a_job
+    find('.search-button-results').click
+
+    expect(page).to have_text(/Enter a postcode/)
+  end
+
+  xscenario 'User gets relevant messaging if there is an API error' do
```
tavy88: Can you add a TODO here with the potential ticket no that will tackle this? Not very clear why it's disabled.

app/models/job_vacancy_search.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def count
+    return unless valid?
+
+    job_vacancies_response.dig('pager', 'total_entries')
+  end
+
+  private
+
+  def job_vacancies_response
```
tavy88: This is optional but I think `job_openings` here would be enough as a name. It's also shorter.

spec/features/jobs_near_me_spec.rb
```diff
@@ -0,0 +1,151 @@
+require 'rails_helper'
+
+RSpec.feature 'Jobs near me', type: :feature do
+  before do
```
smoothcontract: I think it's preferable to use `background` rather than `before` with a feature spec.

spec/features/jobs_near_me_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+  xscenario 'User gets relevant messaging if there is an API error' do
+    user_targets_a_job
+    fill_in('postcode', with: 'NW6 9ET')
+    find('.search-button-results').click
+
+    expect(page).to have_text(/Sorry, there is a problem with this service/)
+  end
+
+  def user_targets_a_job
+    create(:job_profile, :with_html_content, name: 'Admin assistant').tap do |job_profile|
```
smoothcontract: Disappointing choice of job title 😛

app/views/job_vacancies/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    <p class="govuk-body">These jobs are currently being advertised in the <%= link_to('Find a job', 'https://findajob.dwp.gov.uk/', class: 'govuk-link', target: '_blank', data: { tracked_event: "Jobs near me - Clicked 'Find a job'", tracked_event_props: { url: 'https://findajob.dwp.gov.uk/' } }) %> service from the Department for Work and Pensions. You may be able to find more jobs advertised elsewhere by checking other sources of information.</p>
+    <div class="govuk-grid-column-full govuk-!-padding-0">
+      <%= form_with model: @job_vacancy_search, local: true, url: jobs_near_me_path, method: 'get' do |form| %>
+        <%= form_group_tag @job_vacancy_search, :postcode do %>
+          <%= form.label :postcode, class: 'govuk-label' %>
+          <%= errors_tag @job_vacancy_search, :postcode %>
+          <%= text_field_tag(:postcode, @postcode, class: css_classes_for_input(@job_vacancy_search, :postcode, "course-search-input govuk-input govuk-!-width-one-half")) %>
+          <%= button_tag('', class: 'search-button-results', name: nil, aria: { label: 'Search button' }) %>
```
tavy88: Why no button name here?

spec/routing/job_vacancies_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe 'routes for job vacancies', type: :routing do
+  context 'with action_plan feature enabled' do
+    before do
+      enable_feature! :action_plan
+    end
+
+    it 'successfully routes to job_vacancies#index' do
+      expect(get('/jobs-near-me')).to route_to(controller: 'job_vacancies', action: 'index')
```
smoothcontract: Yay 👍

app/controllers/job_profiles_controller.rb
tavy88: Then you can use `helper_method : job_vacancy_count` to access it in the view.

spec/features/jobs_near_me_spec.rb
smoothcontract: Or even better, a `pending` line so it shows up in the Rspec output.

app/controllers/job_vacancies_controller.rb
tavy88: Then you can use helper_method : job_vacancies to access it in the view.

app/views/job_vacancies/index.html.erb
smoothcontract: Good question 🤔

app/controllers/job_profiles_controller.rb
cwrw: The current implementation is simple enough, its defining an instance variable and is explicit, also this is happening in one place, rather than spreading the logic and not seeing the instance variables needed for a view we have them in one place.

app/controllers/job_vacancies_controller.rb
cwrw: same here

app/models/job_vacancy.rb
cwrw: I'm not sure that's necessary, it's just a view concern

app/models/job_vacancy_search.rb
cwrw: We've not used the word openings in this context, so I would rather avoid that

app/views/job_vacancies/index.html.erb
cwrw: When we add the name to this, it will submit a value called button and append it to the url, which we want to avoid doing

app/controllers/job_vacancies_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  private
+
+  def postcode
+    @postcode ||= job_vacancies_params[:postcode] || user_session.postcode
+  end
+
+  def job_vacancies
+    Kaminari
+      .paginate_array(jobs, total_count: @job_vacancy_search.count)
+      .page(job_vacancies_params[:page])
```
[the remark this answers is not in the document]
cwrw: I think in this case id rather leave as is, as I want to be clear that Im using `job_vacancies_params` for pages

app/decorators/job_vacancy_decorator.rb
```diff
@@ -0,0 +1,18 @@
...
+      .join(' - ')
+      .html_safe
+  end
+
+  private
+
+  def formatted_company
+    return unless company
+
+    content_tag(:b, company)
```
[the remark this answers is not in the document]
cwrw: tried that and unfortunately that didn't work, so will have to keep as is

app/models/job_vacancy_search.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def initialize(postcode:, name:, page: 1, distance: 20)
+    @postcode = postcode
+    @name = name
+    @page = page
+    @distance = distance
+  end
+
+  def jobs
+    return [] unless valid? && job_vacancies_response['jobs'].present?
```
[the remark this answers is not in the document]
cwrw: This was intentional as this does fail when we have empty jobs on API errors. I'll use the last implementation as that's nicer

app/models/job_vacancy_search.rb
```diff
@@ -0,0 +1,39 @@
+class JobVacancySearch
+  include ActiveModel::Validations
+
+  attr_reader :postcode, :name, :page, :distance
+  validates :postcode, presence: { message: I18n.t('courses.no_postcode_error') }
```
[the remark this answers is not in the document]
cwrw: I've added a comment to amend when we do get the correct error messages as we haven't yet, and if we will reuse will move it there then

app/controllers/job_vacancies_controller.rb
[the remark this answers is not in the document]
smoothcontract: Cool 👍

app/decorators/job_vacancy_decorator.rb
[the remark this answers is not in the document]
smoothcontract: Okay. Thanks for trying 👍

app/controllers/job_profiles_controller.rb
tavy88: Fair oki

app/models/job_vacancy.rb
smoothcontract: I'm pretty lukewarm about making these constants to be honest. It's always a tradeoff between readability and alleged performance. In practice I doubt we'd be able to measure the performance difference, and it's nice to be able to read these in situ - plus each string is only used in one place so on balance I'd say there's a stronger argument in this case for *not* using constants.

app/views/job_vacancies/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+  end
+%>
+
+<div class="govuk-grid-row govuk-!-margin-top-7">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= target_job.name %> jobs near me</h1>
+    <p class="govuk-body">These <%= target_job.name.downcase %> jobs are being advertised within 20 miles of your postcode.</p>
+    <p class="govuk-body">These jobs are currently being advertised in the <%= link_to('Find a job', 'https://findajob.dwp.gov.uk/', class: 'govuk-link', target: '_blank', data: { tracked_event: "Jobs near me - Clicked 'Find a job'", tracked_event_props: { url: 'https://findajob.dwp.gov.uk/' } }) %> service from the Department for Work and Pensions. You may be able to find more jobs advertised elsewhere by checking other sources of information.</p>
+    <div class="govuk-grid-column-full govuk-!-padding-0">
+      <%= form_with model: @job_vacancy_search, local: true, url: jobs_near_me_path, method: 'get' do |form| %>
```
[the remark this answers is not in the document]
cwrw: By default its false: https://guides.rubyonrails.org/form_helpers.html
 I've gone back and forth a lot on the get vs post, as there was the param in the url as well etc. discussion. I think I want to hold off on this for now, as I would rather amend the paths together in a full sweep across all form submissions if we are to do it. Maybe we can discuss this as a team later this week and make a decision

app/models/job_vacancy_search.rb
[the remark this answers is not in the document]
smoothcontract: Thanks 👍

app/models/job_vacancy_search.rb
tavy88: How about just `job_vacancies` then? Anyway, it's just a personal preference.

app/views/job_vacancies/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    </div>
+    <div class="govuk-grid-column-full govuk-!-padding-0">
+      <% if @job_vacancies.empty? %>
+        <p class="govuk-body-m">0 job vacancies found</p>
+        <p class="govuk-body-m">There aren't any job vacancies in your local area yet.</p>
+        <h2 class="govuk-heading-s">Try again using a different postcode.</h2>
+        <p class="govuk-body">You could try:</p>
+        <ul class="govuk-list govuk-list--bullet">
+          <li>double checking your postcode</li>
+          <li>using a different postcode</li>
```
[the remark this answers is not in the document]
cwrw: This is the same copy as the courses and not the real copy, flagged and pending update in another ticket

spec/features/jobs_near_me_spec.rb
cwrw: haha ;)

app/models/job_vacancy_search.rb
smoothcontract: Current name seems fine

app/views/job_vacancies/index.html.erb
smoothcontract: The button text is also blank, so presumably that just shows `Submit` or something? Admittedly I should probably test this on heroku 😋

app/views/job_vacancies/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+<% content_for :breadcrumb do
+    generate_breadcrumbs(
+      t('breadcrumb.jobs_near_me'),
+      [
+        [t('breadcrumb.task_list_home'), task_list_path],
+        [t('breadcrumb.action_plan'), action_plan_path]
+      ]
+    )
+  end
+%>
```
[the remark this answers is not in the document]
cwrw: Yea I think they are too tightly coupled to be separated, also you can only access this page via the action plan, so maybe it's safe to keep as is for now using the same flag

app/views/job_vacancies/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+          <%= paginate @job_vacancies %>
+        </div>
+      <% end %>
+    </div>
+  </div>
+
+  <div class="govuk-grid-column-one-third">
+    <%= render 'shared/contact_us' %>
+    <%= render 'shared/save_or_return_to_your_results' %>
+    <%= render 'shared/what_you_can_do_next' %>
```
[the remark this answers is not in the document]
cwrw: sure, removing 👍

app/views/job_vacancies/index.html.erb
```diff
@@ -0,0 +1,56 @@
...
+        <p class="govuk-body-m">There aren't any job vacancies in your local area yet.</p>
+        <h2 class="govuk-heading-s">Try again using a different postcode.</h2>
+        <p class="govuk-body">You could try:</p>
+        <ul class="govuk-list govuk-list--bullet">
+          <li>double checking your postcode</li>
+          <li>using a different postcode</li>
+        </ul>
+      <% else %>
+        <ul class="govuk-list">
+          <%= render partial: 'job_vacancy', collection: @job_vacancies %>
```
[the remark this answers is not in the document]
cwrw: alas `JobVacancy` is  not an `ActiveModel-compatible` object. It must implement `:to_partial_path` :(

app/views/job_vacancies/index.html.erb
cwrw: This is our fancy magnifier icon search bar, so no text is needed :)

app/views/job_vacancies/index.html.erb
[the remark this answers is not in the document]
smoothcontract: Agreed. One flag to rule them all 👍

app/views/job_vacancies/index.html.erb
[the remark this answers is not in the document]
smoothcontract: Okay. Onboard with all that 👍

app/views/job_vacancies/index.html.erb
smoothcontract: Of course, D'oh!

Also, just an aside from the DAC stuff - apparently this reads as "Search button button" on a screen reader (the element type is appended to the label). We can fix all of those issues in one pass but this will no doubt have the same accessibility tweaks as the other search buttons.

app/views/job_vacancies/index.html.erb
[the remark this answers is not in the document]
smoothcontract: Okay 👍

app/views/job_vacancies/index.html.erb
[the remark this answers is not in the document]
smoothcontract: That's a shame, but no worries.

PR #269, opened 2019-10-21
Update flow for task-list sections

tavy88:
Since we are introducing the Target this job
functionality, the following changes will happen:

Section 3 has been changed entirely and now links to action plan
Section 4 is now the survey
Section 5 is gone

Can't start yet logic:

- must have skills in session before you can do step 2
- must click "target this job" on a job profile before you can do 3
- once you've done step 2, you can do all of them

If a user removes ones skills, but one has already targetted a job,
then one should see section 2 locked.

Feature is flagged under action_plan feature.


app/views/pages/task_list.html.erb
```diff
@@ -18,9 +18,14 @@
...
-      <%= render 'pages/task_list_items/item_4' %>
-      <%= render 'pages/task_list_items/item_5' %>
+      <% if Flipflop.action_plan? %>
+        <%= render 'pages/task_list_items/v2/item_3' %>
+        <%= render 'pages/task_list_items/v2/item_4' %>
+      <% else %>
+        <%= render 'pages/task_list_items/v1/item_3' %>
+        <%= render 'pages/task_list_items/v1/item_4' %>
+        <%= render 'pages/task_list_items/v1/item_5' %>
+      <% end %>
```
smoothcontract: Nice approach - makes it easy to unpick later 👍

spec/features/task_list_v2_spec.rb
```diff
@@ -0,0 +1,160 @@
+require 'rails_helper'
```
smoothcontract: When we promote the `action_plan` feature, what happens to the `task_list_spec.rb` file - does it get deleted entirely (and this file renamed)? If so might be worth a `TODO` note at the top of each file to make that explicit.

tavy88: The file will get deleted entirely yes and _v2 will be renamed to the one being deleted. Can add the TODO sure.

smoothcontract: Thanks 👍

tavy88: Updated!

smoothcontract: Perfect 👍

spec/features/task_list_v2_spec.rb
```diff
@@ -132,7 +132,7 @@ def ensure_target_job_on_session
     expect(page).to have_current_path(action_plan_path)
   end
 
-  scenario 'User sets a target job on the session, but then removes all skills -> Section 2 becomes locked again' do
+  scenario 'User sets a target job on the session, but then removes all skills -> Sections 2,3,4 becomes locked again' do
```
tavy88: Ah grammar.. this should be `Sections 2,3,4 become`

PR #271, opened 2019-10-22
Use windows-2019 agent for deployments

mafonso:


azure-pipelines-deploy-template.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     environment: '${{parameters.environment}}'
 
     pool:
-      vmImage: 'vs2017-win2016'
+      vmImage: 'windows-2019'
```
mafonso: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops

PR #272, opened 2019-10-22
Auth0 Spike

tavy88:
### Context

This PR is to explore the Auth0 capabilities. We want to see if it can't be used as user authentication solution within our context,


config/routes.rb
```diff
@@ -58,6 +58,9 @@
 
   get '/sign-in/:token', to: 'passwordless/sessions#show', authenticatable: 'user', as: :token_sign_in
 
+  get 'auth/auth0/callback', to: 'auth0#callback'
+  get 'auth/failure', to: 'auth0#failure'
```
smoothcontract: Does the failure route ever get invoked? I can't see how that's mapped to the omniauth authentication process. Or is the path itself some convention within the omniauth gem?

app/controllers/auth0_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Auth0Controller < ApplicationController
+  def callback
+    # This stores all the user information that came from Auth0
+    # and the IdP
+    session[:userinfo] = request.env['omniauth.auth']
```
smoothcontract: Presumably in order to logout this key could just be removed from the session?

tavy88: I haven't implemented the logout function because it's properly documented.

Sample:

```ruby
# app/controllers/logout_controller.rb

class LogoutController < ApplicationController
  include LogoutHelper
  def logout
    reset_session
    redirect_to logout_url.to_s
  end
end

# app/helpers/logout_helper.rb

module LogoutHelper
  def logout_url
    domain = Rails.application.secrets.auth0_domain
    client_id = Rails.application.secrets.auth0_client_id
    request_params = {
      returnTo: root_url,
      client_id: client_id
    }

    URI::HTTPS.build(host: domain, path: '/v2/logout', query: to_query(request_params))
  end

  private

  def to_query(hash)
    hash.map { |k, v| "#{k}=#{URI.escape(v)}" unless v.nil? }.reject(&:nil?).join('&')
  end
end
```

config/routes.rb
tavy88: Same here. Yeah it's a convention and it would require the following configuration:

```ruby
# config/environments/production.rb

OmniAuth.config.on_failure = Proc.new { |env|
  message_key = env['omniauth.error.type']
  error_description = Rack::Utils.escape(env['omniauth.error'].error_reason)
  new_path = "#{env['SCRIPT_NAME']}#{OmniAuth.config.path_prefix}/failure?message=#{message_key}&error_description=#{error_description}"
  Rack::Response.new(['302 Moved'], 302, 'Location' => new_path).finish
}
```
Again I didn't feel the need to add it as it's properly documented.

smoothcontract: Fabulous, thankyou 👍

PR #273, opened 2019-10-22
GET-557 Add caching for health checks

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-557

Add staggered caching for health checks, all should be less than a minute

### DIFF
--- a/app/services/health_check/check_base.rb
+++ b/app/services/health_check/check_base.rb
@@ -15,13 +15,15 @@ module HealthCheck
     end
 
     def detail
-      {
-        status: status,
-        time: time
-      }.tap do |details|
-        details[:metricValue] = value if respond_to?(:value)
-        details[:metricUnit] = unit if respond_to?(:unit)
-        details[:output] = output if output.present?
+      Rails.cache.fetch(self.class.name.humanize, expires_in: self.class::CACHE_EXPIRY) do
+        {
+          status: status,
+          time: time
+        }.tap do |details|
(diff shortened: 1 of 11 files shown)


app/services/health_check/check_base.rb
```diff
@@ -15,13 +15,15 @@ def time
...
-        details[:output] = output if output.present?
+      Rails.cache.fetch(self.class::CACHE_KEY_NAME, expires_in: self.class::CACHE_EXPIRY) do
+        {
+          status: status,
+          time: time
+        }.tap do |details|
+          details[:metricValue] = value if respond_to?(:value)
+          details[:metricUnit] = unit if respond_to?(:unit)
+          details[:output] = output if output.present?
+        end
```
smoothcontract: This is pretty much the exact implementation that I had in mind, however as we already have a unique `name` attribute for each health check class, we can use that as the cache key rather than inventing a new constant. Maybe programmatically add a `_health_check` suffix... Or was there a reason for adding this? Let's discuss anyway.

e.g.

```ruby
Rails.cache.fetch("#{name}_health_check", expires_in: self.class::CACHE_EXPIRY) do
```

Or perhaps

```ruby
Rails.cache.fetch(class.name.humanize, expires_in: self.class::CACHE_EXPIRY) do
```

smoothcontract: Also Rails caching is sooo nice ❤️

PR #275, opened 2019-10-23
GET-490 Avoid session creation on homepage

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-490

Sessions are lazy loaded. If we don't touch teh session on
the homepage we can avoid creating it unnecessarily.
The three places the session is created and avoided are:
1) csrf token: skip this tag if its the homepage
2) start link: render the information page if the cookie
  is missing. The cookie is created when we have a session.
  Default to normal behaviour if the cookie exists
3) authentication partial: default to return to saved results
  unless there is a cookie, where then we default to previous
  behaviour

### DIFF
--- a/spec/features/home_spec.rb
+++ b/spec/features/home_spec.rb
@@ -58,4 +58,10 @@ RSpec.feature 'Navigation from home page' do
 
     expect(page).to have_text('Privacy policy')
   end
+
+  scenario 'Sessions not created when user has no previous session' do
+    visit(root_path)
+
+    expect(Session.count).to be_zero
+  end
 end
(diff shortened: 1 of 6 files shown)


spec/features/home_spec.rb
```diff
@@ -58,4 +58,10 @@
 
     expect(page).to have_text('Privacy policy')
   end
+
+  scenario 'Sessions not created when user has no previous session' do
+    visit(root_path)
+
+    expect(Session.count).to be_zero
```
tavy88: If you remove the condition from `csrf_meta_tags` (https://github.com/DFE-Digital/get-help-to-retrain/pull/275/files#diff-9599427925097c3c66f26ac1e0de5cadL5) the specs will still pass.

https://blog.tomoyukikashiro.me/post/test-csrf-in-feature-test-using-capybara/

That said, although this test is not a full test, I do think it's enough for our case.

PR #276, opened 2019-10-23
Get 569 azure auth2

tavy88:
### Context

SPIKE authentication with Azure AD using Microsoft Graph Gateway.


tavy88:
This is no longer needed.

PR #277, opened 2019-10-24
GET-581 Add smoke tests to QA

mafonso:
### Context
https://dfedigital.atlassian.net/browse/GET-581

Adds a stage dependent on QA for the smoke-tests

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -191,7 +191,7 @@ stages:
           path: '$(build.artifactstagingdirectory)'
           ArtifactName: 'ARMTemplates'
 
-- stage: Deployment
+- stage: QADeployment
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -203,3 +203,40 @@ stages:
...
+    pool:
+      vmImage: 'ubuntu-16.04'
+
+    steps:
+      - script: |
+          git clone https://github.com/DFE-Digital/get-help-to-retrain-smoke-tests.git
+        displayName: Checkout and install test suite
+
+      - task: CmdLine@2
+        continueOnError: 'true'
```
cwrw: 👍

mafonso: Just in case :)

PR #278, opened 2019-10-25
GET- 620 Avoid sending sensitive data as params and general tidy up of routes

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-620

We are currently sending postcodes/emails in params in a few places, change the routes to POST to avoid that.
Another change is to post emails in a hidden form for resending.
Also I've cleaned up the users routes as it was a bit messy and inconsistent.


app/views/users/registration_results_saved.html.erb
```diff
@@ -14,7 +14,10 @@
     <p class="govuk-body">We've sent an email to</p>
     <h3 class="govuk-heading-m"><%= @user.email %></h3>
     <p class="govuk-body">Check your email inbox to make sure you've received the email. You may need to verify your email address by clicking a link in the email.</p>
-    <p class="govuk-body">If you haven't received this email you can <%= link_to 'send it again', email_sent_again_path(email: @user.email), method: :post, class: 'govuk-link' %>.</p>
+    <%= form_with local: true, url: email_sent_again_path do |form| %>
```
cwrw: @smoothcontract I've tried your cleaner suggestion which was to use this instead:
```
<p class="govuk-body">If you haven't received this email you can <%= button_to 'send it again', link_sent_again_path, params: { email: @user.email }, method: :post, class: 'unbutton govuk-body govuk-link' %>.</p>
```
But unfortunately the markup generated looks like the following:
```
<p class="govuk-body">If you haven't received this email you can </p>
<form class="button_to" method="post" action="/link-sent-again">
  <input class="unbutton govuk-body govuk-link" type="submit" value="send it again">
  <input type="hidden" name="authenticity_token" value="5JAXixzo8WeH21nq5cRf8R2nKZ53Ck4sIT94HpoOAMuGr6Y+lkKdRCj/okGlDXM21q3vi4dLPLh22yvOChhJFg==">
  <input type="hidden" name="email" value="email@email.email">
</form>
"."
<p></p>
```
This looks quite strange on the page, but can be fixed with making everything `display: inline` or using spans and more css fixes. But the generated html isn't great as our use case (embedding link in <p> tags) isn't playing well with it. I'm thinking although the `button_to` is quite nice, in this case where we are wrapping the form around our values, I prefer to keep the implementation as so, to avoid the messy html. What are your thoughts?

smoothcontract: If the form needs some styling you can pass a class for that in the `button_to` options, so should be possible to at least have that in one CSS class for all of the forms that need inline linky buttons.

cwrw: I think the issue is mainly with the markup itself, although we can work around it to make it look good on the page, in the source it looks strange and will cause issues with accessibility. Instead I think the custom form will be much better as it wraps everything where it should be. I do like the `button_to` helper, but we can't use it in this use-case unfortunately

smoothcontract: Okay 👍

app/webpacker/styles/_user.scss
```diff
@@ -0,0 +1,9 @@
+.unbutton {
```
tavy88: Hmm not sure if this suggestive enough? 🤔 What do you think?

cwrw: The name `unbutton` was used across many write-ups to remove the css from buttons I came across, I also felt it was quite descriptive of what we're doing here

PR #287, opened 2019-10-31
GET-656 Add autocomplete to all forms across app

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-656

To coform to WCAG guidelines add autcomplete attribute to all inputs
in forms.

The value of attributes has been taken from:
https://www.w3.org/TR/html52/sec-forms.html#sec-autofill
with a table explaining the what the expected values map to
Other good links include:
https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html
and the gov uk example:
https://design-system.service.gov.uk/components/text-input/

### DIFF
--- a/app/views/user_personal_data/_form.html.erb
+++ b/app/views/user_personal_data/_form.html.erb
@@ -2,17 +2,17 @@
   <%= form_group_tag @user_personal_data, :first_name do %>
     <%= f.label :first_name, class: 'govuk-label' %>
     <%= errors_tag @user_personal_data, :first_name %>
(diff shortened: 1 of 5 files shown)


app/views/user_personal_data/_form.html.erb
```diff
@@ -52,15 +52,15 @@
       </h3>
       <%= errors_tag @user_personal_data, :gender %>
       <div class="govuk-radios__item">
-        <%= f.radio_button :gender, :female, class: 'govuk-radios__input', data: { cy: 'pid-gender-female-radio-btn' } %>
+        <%= f.radio_button :gender, :female, class: 'govuk-radios__input', autocomplete: 'sex', data: { cy: 'pid-gender-female-radio-btn' } %>
```
cwrw: so I don't think radio buttons pick up autocomplete, but will keep and see what DAC says about this if anything

PR #288, opened 2019-10-31
Inform user on which roles are removed

tavy88:
We want to inform the users which role they remove
by displaying a flash message.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
https://www.w3.org/TR/WCAG20-TECHS/ARIA19.html

Design info
https://github.com/alphagov/govuk-design-system-backlog/issues/2

Screen Shot 2019-10-31 at 13 59 39

https://dfedigital.atlassian.net/browse/GET-658

### DIFF
--- a/app/controllers/job_profiles_controller.rb
+++ b/app/controllers/job_profiles_controller.rb
@@ -14,7 +14,7 @@ class JobProfilesController < ApplicationController
     user_session.remove_job_profile(resource.id)
 
     redirect_to(
-      skills_path(query_params), flash: { notice: "The #{job_profile.name.downcase} role has been removed." }
+      skills_path(query_params), notice: t('.notice', name: job_profile.name.downcase)
     )
   end
 
--- a/app/controllers/job_profiles_controller.rb
+++ b/app/controllers/job_profiles_controller.rb
@@ -14,7 +14,7 @@ class JobProfilesController < ApplicationController
     user_session.remove_job_profile(resource.id)
 
     redirect_to(
-      skills_path(query_params)
+      skills_path(query_params), flash: { notice: "The #{job_profile.name.downcase} role has been removed." }
     )
   end
 
@@ -41,5 +41,9 @@ class JobProfilesController < ApplicationController
       name: resource.name,
       postcode: user_session.postcode
     ).count
+  end
+
+  def job_profile
+    JobProfile.find_by(slug: params[:id])
   end
 end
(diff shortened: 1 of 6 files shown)


app/views/skills/index.html.erb
```diff
@@ -13,6 +13,9 @@
   <div class="govuk-grid-column-two-thirds">
     <div class="govuk-grid-column-full govuk-!-padding-0">
       <h1 class="govuk-heading-xl">Your skills</h1>
+      <% if flash[:notice].present? %>
+        <div role="alert" class="alert-info" aria-live="polite"><%= flash[:notice] %></div>
```
tavy88: Please check the attached resources.

app/views/skills/index.html.erb
```diff
@@ -13,6 +13,9 @@
   <div class="govuk-grid-column-two-thirds">
     <div class="govuk-grid-column-full govuk-!-padding-0">
       <h1 class="govuk-heading-xl">Your skills</h1>
+      <% if flash[:notice].present? %>
+        <div role="region" class="alert-info" aria-live="polite"><%= flash[:notice] %></div>
+      <% end %>
```
tavy88: Please check the attached links in the PR's description.

app/controllers/job_profiles_controller.rb
```diff
@@ -14,7 +14,7 @@ def destroy
     user_session.remove_job_profile(resource.id)
 
     redirect_to(
-      skills_path(query_params)
+      skills_path(query_params), flash: { notice: "The #{job_profile.name.downcase} role has been removed." }
```
smoothcontract: I like to use locale entries for flash messages if possible - it keeps the controller looking cleaner. Could you add an entry for this please?

tavy88: @smoothcontract updated!

app/views/skills/index.html.erb
smoothcontract: If we end up using this flash pattern on other pages then this markup could be moved to a partial, but I'm happy for it to be inline on this page for now 👍

config/locales/en.yml
```diff
@@ -190,3 +190,7 @@ en-GB:
               blank: Enter a postcode
             dob:
               invalid: Enter a valid date of birth
+
+  job_profiles:
+    destroy:
+      notice: The %{name} role has been removed.
```
smoothcontract: 👍

PR #289, opened 2019-10-31
GET-563 Improve job profile search

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-563

*  Add Specialism column to Job Profiles
* Add specialism column in admin interface to csv import, list, view, update and search
* Update existing rake task for hidden titles from same spreadsheet to now include specialism. Rename the task to import additional data to allow us to import more columns in the future
* Add specialism to job profile query search. Something to note is that we don't order anything past the names so shouldn't rely on the order of the vector now, and revisit the ranking query instead


spec/models/job_profile_search_spec.rb
```diff
@@ -127,14 +134,16 @@
       cook = create(:job_profile, alternative_titles: 'Cook', name: 'name', description: nil)
```
cwrw: This test doesn't make sense anymore after adding all the extra queries. We have to revisit the ranking query as we don't sort anything past the names, and will sort by id after that meaning the relevance of where the column query comes in is irrelevant

app/services/job_profile_import_service.rb
```diff
@@ -62,6 +65,7 @@ def update_job_profile(data, key)
     value = data.delete(key).strip
     job_profile = JobProfile.find_by(key => value)
     if job_profile.present?
+      data = data.reject { |_k, v| v.blank? }
```
smoothcontract: Perhaps here it would be better to replace blank values with nil, that then allows values to be removed from the spreadsheet and the import would correctly set those values in the model to nil. With the code as you have it now there would be no way to remove values in future. It’s definitely nicer to write nil rather than blank values though - presumably this improves text query performance.

spec/services/job_profile_import_service_spec.rb
```diff
@@ -40,29 +40,50 @@
...
+      importer.import_additional_data(additional_data_path)
+      expect(ceo).to have_attributes(hidden_titles: 'Boss man,Big cheese', specialism: 'ufc,wrestler,wwe')
+    end
+
+    it 'ignores empty values when updating job profiles' do
+      acupuncturist = create(:job_profile, name: 'Acupuncturist', slug: 'acupuncturist')
+      importer.import_additional_data(additional_data_path)
+
+      expect(acupuncturist.reload.specialism).to be_nil
+    end
```
smoothcontract: I think you could probably remove this spec now - the one below seems to cover off the same functionality nicely.

PR #290, opened 2019-11-01
Update message for no jobs found

tavy88:
If the user has 0 results when searching for jobs
nearby one should see more details on what to do next.

https://dfedigital.atlassian.net/browse/GET-622

Screen Shot 2019-11-01 at 11 51 37

### DIFF
--- a/app/views/job_vacancies/index.html.erb
+++ b/app/views/job_vacancies/index.html.erb
@@ -13,7 +13,7 @@
 
 <div class="govuk-grid-row govuk-!-margin-top-7">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl"><%= t('.title') %></h1>
+    <h1 class="govuk-heading-xl"><%= t('.title', name: target_job.name) %></h1>
     <p class="govuk-body">These <%= target_job.name.downcase %> jobs are being advertised within 20 miles of your postcode.</p>
     <p class="govuk-body">These jobs are currently being advertised in the <%= link_to('Find a job', 'https://findajob.dwp.gov.uk/', class: 'govuk-link', target: '_blank', data: { tracked_event: "Jobs near me - Clicked 'Find a job'", tracked_event_props: { url: 'https://findajob.dwp.gov.uk/' } }) %> service from the Department for Work and Pensions. You may be able to find more jobs advertised elsewhere by checking other sources of information.</p>
     <div class="govuk-grid-column-full govuk-!-padding-0">
--- a/app/views/job_vacancies/index.html.erb
+++ b/app/views/job_vacancies/index.html.erb
@@ -13,7 +13,7 @@
 
 <div class="govuk-grid-row govuk-!-margin-top-7">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl"><%= target_job.name %> jobs near me</h1>
+    <h1 class="govuk-heading-xl"><%= t('.title') %></h1>
     <p class="govuk-body">These <%= target_job.name.downcase %> jobs are being advertised within 20 miles of your postcode.</p>
     <p class="govuk-body">These jobs are currently being advertised in the <%= link_to('Find a job', 'https://findajob.dwp.gov.uk/', class: 'govuk-link', target: '_blank', data: { tracked_event: "Jobs near me - Clicked 'Find a job'", tracked_event_props: { url: 'https://findajob.dwp.gov.uk/' } }) %> service from the Department for Work and Pensions. You may be able to find more jobs advertised elsewhere by checking other sources of information.</p>
     <div class="govuk-grid-column-full govuk-!-padding-0">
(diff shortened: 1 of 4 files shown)


app/views/job_vacancies/index.html.erb
```diff
@@ -13,7 +13,7 @@
 
 <div class="govuk-grid-row govuk-!-margin-top-7">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl"><%= target_job.name %> jobs near me</h1>
```
cwrw: I'm not sure about this change, it's not very clear from the ticket and I'd double check with Lois

spec/features/jobs_near_me_spec.rb
```diff
@@ -96,8 +96,23 @@
...
     user_targets_a_job
 
-    # TODO: real content pending
-    expect(page).to have_text('0 job vacancies found')
+    ['0 jobs found', 'There are no job vacancies with this job title in your area on the Find a job service from the Department for Work and Pensions.'].each do |text|
+      expect(page).to have_text(text)
+    end
+  end
+
+  scenario 'User can still explore DWP jobs portal without a postcode on the session' do
```
cwrw: I don't think we are doing this level of testing i.e. testing outside links on pages for this test and the one below, I know its a mix right now but I'd avoid doing this as these tests don't add much value

app/views/job_vacancies/index.html.erb
smoothcontract: +1

tavy88: Please check the Slack conversation with Andrew on our channel.

spec/features/jobs_near_me_spec.rb
tavy88: I think it should be covered as the content is expecting that link to existing. Why do you feel it's not valuable? We need to make sure we have that link added.

tavy88:
Yeah, it's also a lot of content duplication with the first paragraph.

app/views/job_vacancies/index.html.erb
cwrw: yea I've gone through it, but I don't believe there should be a heading change, and I'd double check

spec/features/jobs_near_me_spec.rb
cwrw: I don't think it's valuable to test content itself as that's the level of testing observed here. 
For instance the links on the action page make sense to test as they are actionable buttons. This one is just part of the content, and we aren't following this pattern anywhere, for instance next steps page there are many links but we don't test their existence as they don't add much to the journey and are part of the content, and will be covered in manual testing. If they were part of the journey then I would test them

app/views/job_vacancies/index.html.erb
tavy88: Yeah, this review is from a tech point. The content QA is done by Lois. So I think this hint should not be part of our tech reviews as we know things are run by Lois anyway by default as part of the process.

spec/features/jobs_near_me_spec.rb
tavy88: The only reason I added it is because I see it as part of the functionality which translates that user can still keep searching for jobs even if no results are returned. Anyway, I don't wish to spend too much time on such a small thing really.

app/views/job_vacancies/index.html.erb
cwrw: A review encompasses both technical and specification adherence and guidance. As a reviewer we should look at the whole picture, starting with the technical side as well as anything that might have been missed by POs or we find strange and are not clear on, then the PR owner will double check and amend or confirm is correct

spec/features/jobs_near_me_spec.rb
cwrw: I agree this isn't a blocker, but I'd like us to agree on something now so it doesn't keep coming up in later PRs

app/views/job_vacancies/index.html.erb
tavy88: Agreed but she can only do that when the ticket is ready for her QA. If she wants amendments the ticket will return back to DEV state as our process is. The ticket clearly says Jobs near me and this has also been checked with Andrew prior implementation.

cwrw: If a specification is not clear, we should make sure to get clarity on it before starting development or during PR review if something is flagged, not wait till PO review as that might be missed even then which was my main worry, in this case it's a trivial change in other cases it might not be. 

I understand that you got clarity earlier on in slack, but Andrew has +1 my comment about checking earlier in this conversation also, so clarity wasn't confirmed. Anyway I think this conversation can be put to rest now as it seems the conversation has been kicked off in the Jira ticket now (as per my request to Lois to double check the title change).

tavy88: It was Andrew who suggested the Jobs near me thing even though he +1'd your comment :) He was in refinement and I went with his advice. Lois btw has had a look (had time to think as well) and changes are being done. She still needs Murray to simplify the text. So all good and ticket is now under Murray's QA. Once he's happy with the new content the ticket will return to dev and I will do the changes. All good.

tavy88: > If a specification is not clear, we should make sure to get clarity on it before starting development or during PR review if something is flagged, not wait till PO review as that might be missed even then which was my main worry, in this case it's a trivial change in other cases it might not be.
> 
> I understand that you got clarity earlier on in slack, but Andrew has +1 my comment about checking earlier in this conversation also, so clarity wasn't confirmed. Anyway I think this conversation can be put to rest now as it seems the conversation has been kicked off in the Jira ticket now (as per my request to Lois to double-check the title change).

I think the conversation on JIRA has been kicked off by my comment/.note to Lois :) if you check the activity on it.

tavy88:
Will open a new PR as the content has changed completely and starting clean again.

PR #293, opened 2019-11-04
Amendments to skills builder page

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-649

Now instead of listing individual skills, summarises as "3 skills selected" for each job profile. This is less repetitive for screen readers. Cleaned up existing markup slightly and removed redundant styles. Updated feature specs to reflect.

### Guidance to review
I left out the centre align styles from the prototype markup - this looks really weird at narrow device widths. As a result the links to 'edit' and 'remove' are not in quite the same horizontal position as the prototype but responsiveness is improved.

### DIFF
--- a/app/views/skills/_job_profiles_with_skills.html.erb
+++ b/app/views/skills/_job_profiles_with_skills.html.erb
@@ -1,31 +1,29 @@
-  <table class="govuk-table">
-    <thead class="govuk-table__head">
-      <hr class="govuk-section-break govuk-section-break--visible">
-      <% job_profiles_with_skills.each do |job_profile_hash| %>
-      <tr class="govuk-table__row">
-        <th class="govuk-table__header govuk-section-break__l govuk-!-padding-0" scope="col">
-          <div class="govuk-grid-row govuk-!-margin-top-3">
-            <div class="govuk-grid-column-one-third govuk-!-padding-right-0">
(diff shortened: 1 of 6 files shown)


app/views/skills/_job_profiles_with_skills.html.erb
```diff
@@ -1,28 +1,28 @@
-<% job_profiles_with_skills.each do |job_profile_hash| %>
   <table class="govuk-table">
-    <h2 class="govuk-heading-m govuk-!-margin-bottom-1 govuk-!-margin-top-8"><%= job_profile_hash[:name] %></h2>
     <thead class="govuk-table__head">
```
cwrw: I'm wondering about the table, and the fact that it doesn't have body, maybe there is some value in restructuring the html to get a more traditional table?

tavy88: Yeah was thinking the same when I glanced over...

smoothcontract: I did play with that for a while, but it involved a lot of fighting with the default gov.uk table styles. I'm not convinced it needs a table at all really but opted for the option of minimum disturbance. I can have another crack at it if you think it's worthwhile, but I'm kind of inclined to leave it.

PR #294, opened 2019-11-04
GET-667 multicontainer application

mafonso:
### Context
https://dfedigital.atlassian.net/browse/GET-667

This PR changes the deployment from a single container WebApp to a multi-container WebApp.
This will allow us to run another container side-by-side with the app using the same config to start asynchronous tasks.

### DIFF
--- a/azure-pipelines-deploy-template.yml
+++ b/azure-pipelines-deploy-template.yml
@@ -24,6 +24,7 @@ jobs:
                 $EncodedText =[Convert]::ToBase64String($Bytes)
                 echo "##vso[task.setvariable variable=dockerCompose]$EncodedText"
                 $EncodedText
+              displayName: Base64 encode composite file into variable
 
             - task: AzureResourceGroupDeployment@2
               displayName: 'Azure Deployment:Create Or Update Resource Group action on $(resourceEnvironmentName)-$(serviceName)'
--- a/azure-pipelines-deploy-template.yml
+++ b/azure-pipelines-deploy-template.yml
@@ -18,6 +18,13 @@ jobs:
       runOnce:
         deploy:
           steps:
+            - pwsh: |
+                $DockerCompose = Get-Content -Path $(Pipeline.Workspace)\ARMTemplates\azure\docker-compose-azure.yml -Raw
+                $Bytes = [System.Text.Encoding]::ASCII.GetBytes($DockerCompose)
+                $EncodedText =[Convert]::ToBase64String($Bytes)
+                echo "##vso[task.setvariable variable=dockerCompose]$EncodedText"
+                $EncodedText
+
             - task: AzureResourceGroupDeployment@2
               displayName: 'Azure Deployment:Create Or Update Resource Group action on $(resourceEnvironmentName)-$(serviceName)'
               inputs:
@@ -27,7 +34,7 @@ jobs:
                 resourceGroupName: '${{parameters.resourceGroupName}}'
                 csmFile: '$(Pipeline.Workspace)\ARMTemplates\azure\template.json'
                 csmParametersFile: '$(Pipeline.Workspace)\ARMTemplates\azure\config\${{parameters.environment}}.parameters.json'
-                overrideParameters: '-containerImageReference "${{parameters.containerImageReference}}" -supportEmailAddresses ${{parameters.supportEmailAddresses}}'
+                overrideParameters: '-containerImageReference "${{parameters.containerImageReference}}" -dockerComposeFile "$(dockerCompose)"  -supportEmailAddresses ${{parameters.supportEmailAddresses}}'
                 deploymentOutputs: DeploymentOutput
 
             - task: AzureAppServiceManage@0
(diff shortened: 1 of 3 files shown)


azure-pipelines-deploy-template.yml
```diff
@@ -18,6 +18,13 @@ jobs:
       runOnce:
         deploy:
           steps:
+            - pwsh: |
+                $DockerCompose = Get-Content -Path $(Pipeline.Workspace)\ARMTemplates\azure\docker-compose-azure.yml -Raw
+                $Bytes = [System.Text.Encoding]::ASCII.GetBytes($DockerCompose)
+                $EncodedText =[Convert]::ToBase64String($Bytes)
```
mafonso: This is to work around new line handling by the echo command two lines below.

azure-pipelines-deploy-template.yml
```diff
@@ -18,6 +18,13 @@ jobs:
       runOnce:
         deploy:
           steps:
+            - pwsh: |
+                $DockerCompose = Get-Content -Path $(Pipeline.Workspace)\ARMTemplates\azure\docker-compose-azure.yml -Raw
+                $Bytes = [System.Text.Encoding]::ASCII.GetBytes($DockerCompose)
```
mafonso: `ASCII.GetBytes($DockerCompose)` to avoid getting a UTF16 file passed to the WebApp and ending with this magnificent result.

image

azure/template.json
```diff
@@ -215,7 +233,7 @@
     "domainDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-', parameters('domainDatabaseName'),'Psql')]",
     "restrictedDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-', parameters('restrictedDatabaseName'),'Psql')]",
     "actionGroupName": "[concat(variables('resourceNamePrefix'), '-ag')]",
-    "appServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]",
+    "appServiceRuntimeStack": "[concat('COMPOSE|', base64(replace(base64ToString(parameters('dockerComposeFile')),'__CONTAINER_REFERENCE__',parameters('containerImageReference'))))]",
```
mafonso: Decode the original string, replace the gitsha on it, and reenconde.. because Webapps want a base64 encoded string.

azure-pipelines-deploy-template.yml
```diff
@@ -18,6 +18,13 @@ jobs:
...
+                $EncodedText =[Convert]::ToBase64String($Bytes)
+                echo "##vso[task.setvariable variable=dockerCompose]$EncodedText"
```
mafonso: If we don't base64 encode before, this would render the newlines and we end up with the first line of the compose file in the variable, because that's what echo does.

azure-pipelines-deploy-template.yml
```diff
@@ -27,7 +34,7 @@ jobs:
                 resourceGroupName: '${{parameters.resourceGroupName}}'
                 csmFile: '$(Pipeline.Workspace)\ARMTemplates\azure\template.json'
                 csmParametersFile: '$(Pipeline.Workspace)\ARMTemplates\azure\config\${{parameters.environment}}.parameters.json'
-                overrideParameters: '-containerImageReference "${{parameters.containerImageReference}}" -supportEmailAddresses ${{parameters.supportEmailAddresses}}'
+                overrideParameters: '-containerImageReference "${{parameters.containerImageReference}}" -dockerComposeFile "$(dockerCompose)"  -supportEmailAddresses ${{parameters.supportEmailAddresses}}'
```
mafonso: `dockerCompose`variable should have a safe base64 ASCII string by this point.

azure/template.json
```diff
@@ -468,10 +486,13 @@
           "uri": "[concat(variables('deploymentUrlBase'), 'app-service-plan.json')]",
           "contentVersion": "1.0.0.0"
         },
-        "parameters": {
+          "parameters": {
```
cwrw: ✂️ extra space

azure-pipelines-deploy-template.yml
```diff
@@ -18,6 +18,13 @@ jobs:
       runOnce:
         deploy:
           steps:
+            - pwsh: |
+                $DockerCompose = Get-Content -Path $(Pipeline.Workspace)\ARMTemplates\azure\docker-compose-azure.yml -Raw
+                $Bytes = [System.Text.Encoding]::ASCII.GetBytes($DockerCompose)
+                $EncodedText =[Convert]::ToBase64String($Bytes)
+                echo "##vso[task.setvariable variable=dockerCompose]$EncodedText"
+                $EncodedText
```
cwrw: why do we need this line, shouldn't setting the variable above be enough?

mafonso: That was just debug. I'll leave it as it might be handy in future cases. 
It's not worth calling extra command just to format it, it's also the only user output of the task.
But I did set a display name on the task for clarity.

PR #295, opened 2019-11-04
Content updates for Action Plan section

tavy88:
This PR introduces the following changes:

Amendment of 1st paragrapg (type of work)
Target job: change the word ‘job’ to ‘type of work’ throughout this section
Your skills section: 2nd para change ‘You can see different types of work ideas if you edit your skills.’
CV section: change heading to ‘Get ready to change types of work’
Apply for a job section: change link to ‘Change target types of work’

as per prototype

https://get-help-to-retrain-prototype.herokuapp.com/main/release4_e2e_v2/action_plan?task-2=complete

### DIFF
--- a/app/webpacker/styles/_action-plan.scss
+++ b/app/webpacker/styles/_action-plan.scss
@@ -7,6 +7,6 @@
 
   &.govuk-button {
     display: block;
-    width: 210px;
+    width: 100%;
   }
 }
--- a/app/webpacker/styles/_action-plan.scss
+++ b/app/webpacker/styles/_action-plan.scss
@@ -7,6 +7,6 @@
 
   &.govuk-button {
     display: block;
-    width: 100%;
+    width: 210px;
   }
 }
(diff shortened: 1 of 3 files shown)


app/webpacker/styles/_action-plan.scss
```diff
@@ -7,6 +7,6 @@
 
   &.govuk-button {
     display: block;
-    width: 100%;
+    width: 210px;
```
smoothcontract: 🤔 This looks fine at a wide display size, but the responsive behaviour is way worse with this change. When I added the action plan I did make this 100% width deliberately, and Steve was happy with that change. It's also outside the scope of this story. Perhaps share a link to the prototype app with Steve and see what he thinks; my personal view is that this should be reverted.

tavy88: Ok, fair I thought it was a new design for a reduced width button.

app/webpacker/styles/_action-plan.scss
```diff
@@ -10,3 +10,4 @@
     width: 100%;
   }
 }
+
```
cwrw: 🦂

PR #300, opened 2019-11-05
Inform users if the Notify Service is not working

tavy88:
### Description

Add a Return to results: Error page
when notify fails to send magic link email

Screen Shot 2019-11-07 at 17 06 55

https://dfedigital.atlassian.net/browse/GET-496

### Testing
Try altering the API key on HEROKU to get to this page naturally.

### DIFF
--- a/app/views/errors/return_to_saved_results_error.html.erb
+++ b/app/views/errors/return_to_saved_results_error.html.erb
@@ -1,0 +1,25 @@
+<% page_title :errors_return_to_saved_results_error %>
+
+<% content_for :breadcrumb do
+    generate_breadcrumbs(
+      t('breadcrumb.link_not_sent'),
+      [
+        [t('breadcrumb.task_list_home'), task_list_path]
+      ]
+    )
+  end
+%>
+
+<main role="main" class="govuk-main-wrapper" id="main-content">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">Sorry, there is a problem with the service</h1>
+       <p class="govuk-body">Unfortunately we weren’t able to send you a link to your saved results at this time.</p>
+      <p class="govuk-body">You could try again later.</p>
+      <%= link_to 'Return to previous page', return_to_saved_results_path, class: 'govuk-link govuk-body', data: { module: 'govuk-button' } %>
+    </div>
+    <div class="govuk-grid-column-one-third">
+      <%= render 'shared/contact_us' %>
+    </div>
+  </div>
+</main>
--- a/app/views/errors/return_to_saved_results_error.html.erb
+++ b/app/views/errors/return_to_saved_results_error.html.erb
@@ -1,25 +1,13 @@
 <% page_title :errors_return_to_saved_results_error %>
 
-<% content_for :breadcrumb do
-    generate_breadcrumbs(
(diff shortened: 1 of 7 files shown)


spec/features/user_sign_in_spec.rb
```diff
@@ -70,6 +70,36 @@ def sign_in_user(reset_sessions: true)
...
+    visit(return_to_saved_results_path)
+
+    fill_in('email', with: 'test@test.test')
+
+    click_on('Send link')
+
+    expect(page).to have_current_path(return_to_saved_results_error_path)
+  end
+
+  scenario 'User redirected to error page if there is the Notify Service is using an invalid API KEY' do
```
cwrw: I don't believe this test should be here, but rather at the unit level for the notify service. This is due to us only rescuing from `NotifyAPIError` at the controller level.

tavy88: Yeah, can move it there.

app/views/errors/return_to_saved_results_error.html.erb
```diff
@@ -0,0 +1,12 @@
+<% page_title :errors_return_to_saved_results_error %>
+
+<main role="main" class="govuk-main-wrapper" id="main-content">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">Sorry, there is a problem with the service</h1>
+       <p class="govuk-body">Unfortunately we weren’t able to send you a link to your saved results at this time.</p>
+      <p class="govuk-body">You could try again later.</p>
+      <%= link_to 'Return to previous page', return_to_saved_results_path, class: 'govuk-link govuk-body', data: { module: 'govuk-button' } %>
```
cwrw: since this isn't a button can you remove the `data-module`, also not sure we need the `govuk-body` either class

config/locales/en.yml
```diff
@@ -21,6 +21,7 @@ en-GB:
     save_your_results: Save your results
     return_to_saved_results: Return to saved results
     link_sent: Link sent
+    link_not_sent: Link not sent
```
cwrw: this is no longer needed

smoothcontract: good catch 👍

tavy88: true

app/views/errors/return_to_saved_results_error.html.erb
tavy88: It's not 100% sure - will need Steve's 👁

spec/services/notify_service_spec.rb
```diff
@@ -17,7 +17,15 @@
...
+      it 'raises NotifyAPIError' do
+        expect {
+          service.send_email(email_address: email, template_id: template_id)
+        }.to raise_error(described_class::NotifyAPIError)
+      end
+    end
+
+    context 'when the NOTIFY_API_KEY is wrong' do
+      it 'raises NotifyAPIError' do
         expect {
```
cwrw: I'm not sure I understand how this is passing? We're not stubbing the client here to raise an error?

tavy88: Yeah

cwrw: We shouldn't be doing actual calls in tests, and should mock the client instead

tavy88: Should be stubbed.

PR #307, opened 2019-11-06
GET-682: Fix IE11 compatibility issue 

tavy88:
### Context

Fix IE11 compatibility issue

We currently have for event tracking that is using
forEach function. However, IE11 throws the following
error:

Object doesn't support property or method 'forEach'

Because of this the behaviour our accordion component
is broken on IE11 giving bad user experience.

Decided to go with the simplest solution of using the
classic for loop.

Screen Shot 2019-11-06 at 12 25 29

### Test

Please use: https://dev1.nrs-ghtr.org.uk/task-list
And browserling.com

### DIFF
--- a/app/webpacker/packs/track-events.js
+++ b/app/webpacker/packs/track-events.js
@@ -1,9 +1,11 @@
 function TrackEvents () {
   this.start = function () {
     if (typeof(appInsights) !== 'undefined') {
-      document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
-        trackComponentOnClick(element);
-      });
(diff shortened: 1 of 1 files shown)


app/webpacker/packs/track-events.js
```diff
@@ -1,9 +1,11 @@
 function TrackEvents () {
   this.start = function () {
     if (typeof(appInsights) !== 'undefined') {
-      document.querySelectorAll('[data-tracked-event]').forEach(function (element) {
-        trackComponentOnClick(element);
-      });
+      var elements = document.querySelectorAll('[data-tracked-event]')
```
cwrw: 🦂 `;`

PR #309, opened 2019-11-06
GET-621 Show error page if Find a Job API is down and cleanup exceptions for service

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-621

The errors for find a job service were a bit of a mess, so now made them
consistent to the following:

1) If there is an internal API response error we raise a Response Error
  which we catch and for outside services this is seen as an API error
2) If there is a NET::HTTP error, the best way is to save it through standard
   error see: https://stackoverflow.com/questions/5370697/what-s-the-best-way-to-handle-exceptions-from-nethttp
   rescue that and raise also API error for outside services. Since we are
   now doing that in the find a job service itself, remove the need to
  do that from the healthcheck service and rescue from API error instead
3) For job vacancies on the search page, show error page and rescue from
   new API error
4) For job vacancy count on job profiles page, rescue from API error but return
   nil to hide that section. This has been moved to the controller instead now

For testing purposes, I went with the following scenarios locally:
1) Job vacancy search: shows error if `Net::HTTP` exception or `Response` exception are thrown
2) Job profile count: shows nothing if `Net::HTTP` exception or `Response` exception are thrown. Shows nothing if no postcode entered, shows number over 0, and a different text for 0 otherwise
3) Health check: shows error if `Net::HTTP` exception or `Response` exception are thrown with **correct** message, show pass otherwise

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ restricted_erd.pdf
 *.log
 coverage
 **/local-parameters.json
+.DS_Store
--- a/app/controllers/job_profiles_controller.rb
+++ b/app/controllers/job_profiles_controller.rb
@@ -41,5 +41,7 @@ class JobProfilesController < ApplicationController
       name: resource.name,
       postcode: user_session.postcode
     ).count
+  rescue FindAJobService::APIError
+    nil
   end
 end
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -35,3 +35,4 @@ restricted_erd.pdf
 *.log
 coverage
 **/local-parameters.json
+.DS_Store
```
tavy88: Is this related to this PR?

cwrw: yes, the `.DS_Store` file was appearing in git changes and I wanted it to be ignored

PR #314, opened 2019-11-08
Add service error page for user registration

tavy88:
### Context

If notify service is down, users will get an error
page explaining what is happening.

Users can still resume their journey.

Screen Shot 2019-11-11 at 15 59 41

### Testing

1. Please check you can resume the journey (try saving results at different stages)
2. Deeplinking to the error page should redirect to task-list page.

https://dfedigital.atlassian.net/browse/GET-526

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,4 +1,4 @@
-class ErrorsController < ActionController::Base
+class ErrorsController < ApplicationController
   layout 'application'
 
   def not_found
--- a/app/views/errors/save_results_error.html.erb
+++ b/app/views/errors/save_results_error.html.erb
@@ -1,0 +1,24 @@
+<% page_title :errors_save_results_error %>
+
+<% content_for :breadcrumb do
+    generate_breadcrumbs(
+      t('breadcrumb.results_saved'),
+      [
+        [t('breadcrumb.task_list_home'), task_list_path]
+      ]
+    )
+  end
+%>
+
+<div class="govuk-grid-row govuk-!-margin-top-7">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Your results have been saved</h1>
+    <p class="govuk-body">We have saved your results using the email address you have provided.</p>
+    <p class="govuk-body">We cannot send you a confirmation email right now because of an error with our system. Your results are still saved.</p>
+    <p class="govuk-body">To see your saved skills and job matches in future, just return to the service and enter your email address. You will then be sent a link so you can return to your saved results.</p>
+    <%= link_to 'Continue', user_session.registration_triggered_path || task_list_path, class: 'govuk-button', data: { module: 'govuk-button' } %>
+  </div>
+  <div class="govuk-grid-column-one-third">
+    <%= render "shared/contact_us" %>
+  </div>
+</div>
--- a/spec/features/user_registration_spec.rb
+++ b/spec/features/user_registration_spec.rb
@@ -120,7 +120,19 @@ RSpec.feature 'User registration' do
     expect(page).to have_current_path(next_steps_path)
   end
 
-  scenario 'User can resume the journey from the point one saves results but gets the error page (NotifyService is down)' do
+  scenario 'When NotifyService is down, the user sees the correct error page when one tries to save results' do
+    allow(Notifications::Client).to receive(:new).and_raise(ArgumentError)
+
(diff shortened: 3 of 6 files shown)


app/controllers/errors_controller.rb
```diff
@@ -1,4 +1,4 @@
-class ErrorsController < ActionController::Base
+class ErrorsController < ApplicationController
```
smoothcontract: Not sure why you made this change as we don't need any of the application controller included stuff for these static error pages. Seems slightly unrelated to the PR? If it works without this change then I'd prefer to revert this.

app/views/errors/save_results_error.html.erb
```diff
@@ -0,0 +1,24 @@
...
+%>
+
+<div class="govuk-grid-row govuk-!-margin-top-7">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Your results have been saved</h1>
+    <p class="govuk-body">We have saved your results using the email address you have provided.</p>
+    <p class="govuk-body">We cannot send you a confirmation email right now because of an error with our system. Your results are still saved.</p>
+    <p class="govuk-body">To see your saved skills and job matches in future, just return to the service and enter your email address. You will then be sent a link so you can return to your saved results.</p>
+    <%= link_to 'Continue', user_session.registration_triggered_path || task_list_path, class: 'govuk-button', data: { module: 'govuk-button' } %>
+  </div>
```
smoothcontract: Hmm.. I guess this negates my earlier comment as you're referring to the session here. I'm not entirely convinced about adding the email error page here, although it seems to follow the pattern of other things. Maybe something to review in future, perhaps grouping this with the other save and return stuff instead.

spec/features/user_registration_spec.rb
```diff
@@ -120,6 +120,33 @@ def register_user
     expect(page).to have_current_path(next_steps_path)
   end
 
+  scenario 'User sees the error page when one tries to save results NotifyService is down' do
```
smoothcontract: `when` NotifyService is down - might be better wording?

spec/features/user_registration_spec.rb
```diff
@@ -120,6 +120,33 @@ def register_user
...
+    visit(job_profile_skills_path(job_profile_id: job_profile.slug))
+    click_on('Select these skills')
+    click_on('Save my results')
+    fill_in('email', with: 'test@test.test')
+    click_on('Save your results')
+
+    expect(page).to have_current_path(save_results_error_path)
+  end
+
+  scenario 'User can resume the journey from the point one saves results but gets the error page (NotifyService is down)' do
```
smoothcontract: This scenario seems to be exactly the same as the one before, but with an extra step. I'd suggest either removing the previous test completely, or perhaps streamline the setup steps for this test by directly visiting the `save_results_error_path` and then clicking `Continue`. Seems no sense in repeating the same steps as above as it increases test execution time but doesn't increase coverage.

smoothcontract: Does this work? If so it seems to demonstrate the scenario more clearly:

```ruby
visit(save_results_error_path)
click_on('Continue')
expect(page).to have_current_path(training_hub_path)
```

app/views/errors/save_results_error.html.erb
tavy88: Exactly we need to access `user_session`

app/controllers/errors_controller.rb
tavy88: Yep check my second comment.

spec/features/user_registration_spec.rb
tavy88: Yep, it was supposed to be in brackets.

spec/features/user_registration_spec.rb
tavy88: I'm not sure this would cover both scenarios:

1. One scenario is checking that you see the error page (hence the assert for the correct URL) (have_current_path(error-page-url)
2. The second scenario is testing that the user can actually resume the journey from where they left it.

I think both are relevant.

What you are suggesting would be a third scenario: Deeplinking should take the user back to task-page only which I might add as well.

app/views/errors/save_results_error.html.erb
smoothcontract: Cool 👍

spec/features/user_registration_spec.rb
smoothcontract: Okay, you convinced me 😀

PR #315, opened 2019-11-11
Fix small typo on action plan page

tavy88:
### Context

Fix small type from **Change type of work** to **Change types of work**

### Ticket

https://dfedigital.atlassian.net/browse/GET-647


tavy88:
Not needed as the correct way is singular type of work.

PR #317, opened 2019-11-11
Many tweaks to job profile page and associated decorator

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-646

Found some interesting quirks with the JobProfileDecorator:

* Firstly, the use of `.children` causes some strange behaviour when removing nodes. Although Nokogiri allows nodes to be removed, they don't get removed from the document (or they do sometimes, if the nodes are removed multiple times!). Removing `.children` fixes that.
* Secondly, many of the mutation methods were using global xpath expressions for many things, which meant that each time a single section is parsed, the content is mutated across the whole document. This is slow and introduces a fun side effect where every heading eventually ends up as an h2 (we resize headings to be one size larger, but calling mutate repeatedly for multiple sections eventually upsizes all headings to the largest size). To fix that I've used relative rather than global xpath expressions. This did mean that the "More Info" section looked different, as it includes some `<h4>` tags which don't look right in the job profile page as `<h3>` tags - so I've updated the relevant `mutate_h4_tags` method to convert those to `<h2>` which is what it did previously. It might not be the correct intended appearance, but at least matches the existing page.
* Thirdly, I found the approach of passing xpath arguments to the `section` method to be confusing, and harder to test. The specs also used a different xpath expression to the one in the `relevant_sections` partial - it's easy to introduce inconsistencies with the original approach, so I refactored to use a hash of supported sections, and pass the relevant `key` for the section of interest. This gives a cleaner interface when using the decorator and avoids bleeding the xpath expressions outside of the decorator class.

Other stuff that's changed:
* Rearranged order of sentences and changed heading within job profile brief partial
* Removed typical hours and you could work sections from job profile, along with associated styles and images
* Skills are now rendered using the associated skills model for the job profile, rather than parsing HTML
* Apprenticeship content now strips out everything after and including the `Entry requirements` heading - but only within that section. (This was a massive pain to get working, compounded by the issues identified above with the decorator).
* Changed 'job' to 'vacancy` for local jobs display via DWP API
* Changed 'job' to 'type of work' throughout the job profile page (this broke a few feature tests that clicked "Target this job")
* Updated job profile decorator specs to be consistent with actual behaviour within the decorator (e.g. the decorator strips out the h2 heading for apprenticeship section, but the decorator used a different xpath expression that includes the heading, also `<a>` was missing a closing tag that messed up the link substitution).

### Guidance to review
Easiest way to test is to compare two job profiles, using this branch against QA environment.


app/decorators/job_profile_decorator.rb
```diff
@@ -111,6 +114,18 @@ def html_body
     @html_body ||= Nokogiri::HTML(content)
   end
 
+  def mutate_content
+    remove_siblings @doc.xpath("h4[.='Entry requirements']").first
```
tavy88: Should this method be parametrized? Or the name changed to something less generic?

tavy88: Also please: https://github.com/airbnb/ruby#method-calls

app/decorators/job_profile_decorator.rb
```diff
@@ -111,6 +114,18 @@ def html_body
...
+  end
+
+  def remove_siblings(node)
+    while node
+      node_to_remove = node
+      node = node.next
+      node_to_remove.remove
+    end
```
tavy88: I think you can do something like:

```ruby
node.xpath('following-sibling::*').remove
```

E.g

```ruby
irb(main):001:0> my_html = '<body>
irb(main):002:0'   <div id="a">
irb(main):003:0'     <h2>My Section</h2>
irb(main):004:0'     <p class="backtotop">Back to Top</p>
irb(main):005:0'     <p>More Content</p>
irb(main):006:0'     <p>Even More Content</p>
irb(main):007:0'   </div>
irb(main):008:0' </body>'
=> "<body>\n  <div id=\"a\">\n    <h2>My Section</h2>\n    <p class=\"backtotop\">Back to Top</p>\n    <p>More Content</p>\n    <p>Even More Content</p>\n  </div>\n</body>"
irb(main):009:0> doc = Nokogiri::HTML(my_html)
=> #<Nokogiri::HTML::Document:0x3fe1b33ca24c name="document" children=[#<Nokogiri::XML::DTD:0x3fe1b33ebf28 name="html">, #<Nokogiri::XML::Element:0x3fe1b33ef830 name="html" children=[#<Nokogiri::XML::Element:0x3fe1b33ef5b0 name="body" children=[#<Nokogiri::XML::Text:0x3fe1b33ef344 "\n  ">, #<Nokogiri::XML::Element:0x3fe1b33ef27c name="div" attributes=[#<Nokogiri::XML::Attr:0x3fe1b33ef1dc name="id" value="a">] children=[#<Nokogiri::XML::Text:0x3fe1b33ee96c "\n    ">, #<Nokogiri::XML::Element:0x3fe1b33ee8a4 name="h2" children=[#<Nokogiri::XML::Text:0x3fe1b33ee6ec "My Section">]>, #<Nokogiri::XML::Text:0x3fe1b33ee50c "\n    ">, #<Nokogiri::XML::Element:0x3fe1b33ee458 name="p" attributes=[#<Nokogiri::XML::Attr:0x3fe1b33ee3f4 name="class" value="backtotop">] children=[#<Nokogiri::XML::Text:0x3fe1b33f9f24 "Back to Top">]>, #<Nokogiri::XML::Text:0x3fe1b33f9b28 "\n    ">, #<Nokogiri::XML::Element:0x3fe1b33f9a74 name="p" children=[#<Nokogiri::XML::Text:0x3fe1b33f98bc "More Content">]>, #<Nokogiri::XML::Text:0x3fe1b33f9704 "\n    ">, #<Nokogiri::XML::Element:0x3fe1b33f9650 name="p" children=[#<Nokogiri::XML::Text:0x3fe1b33f9498 "Even More Content">]>, #<Nokogiri::XML::Text:0x3fe1b33f92f4 "\n  ">]>, #<Nokogiri::XML::Text:0x3fe1b33f9150 "\n">]>]>]>
irb(main):010:0> div = doc.at('#a')
=> #<Nokogiri::XML::Element:0x3fe1b33ef27c name="div" attributes=[#<Nokogiri::XML::Attr:0x3fe1b33ef1dc name="id" value="a">] children=[#<Nokogiri::XML::Text:0x3fe1b33ee96c "\n    ">, #<Nokogiri::XML::Element:0x3fe1b33ee8a4 name="h2" children=[#<Nokogiri::XML::Text:0x3fe1b33ee6ec "My Section">]>, #<Nokogiri::XML::Text:0x3fe1b33ee50c "\n    ">, #<Nokogiri::XML::Element:0x3fe1b33ee458 name="p" attributes=[#<Nokogiri::XML::Attr:0x3fe1b33ee3f4 name="class" value="backtotop">] children=[#<Nokogiri::XML::Text:0x3fe1b33f9f24 "Back to Top">]>, #<Nokogiri::XML::Text:0x3fe1b33f9b28 "\n    ">, #<Nokogiri::XML::Element:0x3fe1b33f9a74 name="p" children=[#<Nokogiri::XML::Text:0x3fe1b33f98bc "More Content">]>, #<Nokogiri::XML::Text:0x3fe1b33f9704 "\n    ">, #<Nokogiri::XML::Element:0x3fe1b33f9650 name="p" children=[#<Nokogiri::XML::Text:0x3fe1b33f9498 "Even More Content">]>, #<Nokogiri::XML::Text:0x3fe1b33f92f4 "\n  ">]>
irb(main):011:0> div.at('.backtotop').xpath('following-sibling::*').remove
=> [#<Nokogiri::XML::Element:0x3fe1b33f9a74 name="p" children=[#<Nokogiri::XML::Text:0x3fe1b33f98bc "More Content">]>, #<Nokogiri::XML::Element:0x3fe1b33f9650 name="p" children=[#<Nokogiri::XML::Text:0x3fe1b33f9498 "Even More Content">]>]
irb(main):012:0> puts div
<div id="a">
    <h2>My Section</h2>
    <p class="backtotop">Back to Top</p>


  </div>
=> nil
irb(main):013:0>
```

app/decorators/job_profile_decorator.rb
tavy88: Also, this method seems to be doing too many things in one go, violating a bit the single responsibility principle. Would you consider handling the node search in its own method?

```ruby
def headline_node(context)
  @doc.xpath("h4[.='#{context}']").first
end

....
remove_siblings(headline_node('Entry requirements'))
....
```

Seems a bit more readable this way too.

app/webpacker/styles/_job-profile.scss
```diff
@@ -124,22 +124,6 @@
...
-
-.time-icon {
-  @extend %responsive-icon;
-
-  background: url('../images/time.svg');
-  background-position: center;
-  background-repeat: no-repeat;
-}
-
 .arrow-falling-icon {
```
tavy88: Nice!

app/decorators/job_profile_decorator.rb
smoothcontract: I will amend this, wanted to address one of your points specifically. We only care about the rubocop style guide - that's the GDS standard. I'm not interested in aligning to the recommendations from another third party style guide from AirBNB. Hence calling methods without brackets is absolutely fine.

app/decorators/job_profile_decorator.rb
smoothcontract: I was looking for something like this - much more elegant so will amend. 👍

app/decorators/job_profile_decorator.rb
tavy88: Oki fair :)

PR #318, opened 2019-11-12
GET-600 Add Questions for training and job hunting

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-600

* Add pages for training and job hunting
* If pages answered skip from targeted jobs
* Add different logic for action plan to show what you selected

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,6 +16,10 @@ class ApplicationController < ActionController::Base
     helpers.target_job
   end
 
+  def redirect_unless_target_job
+    redirect_to task_list_path unless target_job.present?
+  end
+
   helper_method :user_session, :current_user
 
   private
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -1,7 +1,11 @@
 class QuestionsController < ApplicationController
+  def training
+    @referer = request.referer && URI(request.referer).path == action_plan_path
+  end
+
   def training_answers
     user_session.training = questions_params[:training]
-    redirect_to helpers.job_hunting_questions || action_plan_path
+    redirect_to action_path_or_job_hunting_path
   end
 
   def job_hunting_answers
@@ -13,8 +17,15 @@ class QuestionsController < ApplicationController
 
   def questions_params
     params.permit(
+      :action_plan_referer,
       training: [],
       job_hunting: []
     )
+  end
+
+  def action_path_or_job_hunting_path
+    return action_plan_path if questions_params[:action_plan_referer] == 'true'
+
+    helpers.job_hunting_questions || action_plan_path
   end
 end
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -1,7 +1,7 @@
 class QuestionsController < ApplicationController
   def training_answers
     user_session.training = questions_params[:training]
-    redirect_to action_plan_path
+    redirect_to helpers.job_hunting_questions || action_plan_path
   end
 
   def job_hunting_answers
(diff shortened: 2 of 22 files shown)


app/controllers/questions_controller.rb
```diff
@@ -0,0 +1,31 @@
+class QuestionsController < ApplicationController
+  def training
+    @referer = request.referer && URI(request.referer).path == action_plan_path
```
smoothcontract: Maybe rename this to something more specific, e.g. `@referred_from_action_plan`

smoothcontract: Or even `@action_plan_referer` to match the param name used below.

app/helpers/question_helper.rb
```diff
@@ -0,0 +1,9 @@
+module QuestionHelper
+  def job_hunting_questions
+    job_hunting_questions_path unless user_session.job_hunting.any?
+  end
+
+  def training_questions
+    training_questions_path unless user_session.training.any?
+  end
+end
```
smoothcontract: Have a vague sensation that these might be redundant - or at least there's a scenario where this won't work. If user goes through the two question pages, to action plan, and then back to task list then I think the link to action plan will instead direct them to the first question page?

cwrw: rejigged the journey after discussing in person

spec/features/questions_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+  scenario 'User sees training questions if only job hunting questions already answered' do
+    job_profile = create(:job_profile, :with_html_content)
+    visit(job_hunting_questions_path)
+    check('I want advice on creating or updating a CV', allow_label_click: true)
+    visit(job_profile_path(job_profile.slug))
+    click_on('Target this type of work')
+
+    expect(page).to have_current_path(training_questions_path)
+  end
```
smoothcontract: This is a slightly unrealistic scenario as user would have to deep link to the job hunting questions after targeting a job (i.e. manually skipping over the training questions page without submitting). I wonder if there's a cleaner way to address the deep linking scenario - I did test the review app and it's possible with a new session to go directly to the questions pages, submit both and the user is then redirected to task list (as there's no target job in the session).

Perhaps leave this test scenario in place, but add another check to ensure there's a target job in the session before rendering or submitting the questions pages. This is already defined in the `PagesController` so maybe something to DRY up:

```ruby
    redirect_to task_list_path unless target_job.present?
```

spec/routing/action_plan_spec.rb
```diff
@@ -35,5 +43,13 @@
     it 'does not root to errors#jobs_near_me_error' do
       expect(get('/jobs-near-me-error')).not_to be_routable
     end
+
+    it 'does not root to questions#training' do
```
smoothcontract: route

spec/routing/action_plan_spec.rb
```diff
@@ -35,5 +43,13 @@
...
+      expect(get('/training-questions')).not_to be_routable
+    end
+
+    it 'does not root to questions#job_hunting_questions' do
```
smoothcontract: route

app/controllers/application_controller.rb
```diff
@@ -16,7 +16,11 @@ def target_job
     helpers.target_job
   end
 
-  helper_method :user_session, :current_user
+  def redirect_unless_target_job
+    redirect_to task_list_path unless target_job.present?
+  end
+
+  helper_method :user_session, :current_user, :redirect_unless_target_job
```
smoothcontract: Don't think the new redirect method needs to be a helper method as it won't be called from a view.

PR #322, opened 2019-11-13
Semantic logger formatter

tavy88:
### Context
Format rails logs as JSON so we can nicely broadcast to DataDog.

### Screenshot

Screen Shot 2019-11-13 at 16 10 41

image

### Link to documentation
http://rocketjob.github.io/semantic_logger/rails


Gemfile
```diff
@@ -137,3 +134,8 @@ group :test do
   gem 'vcr', '~> 5.0', require: false
   gem 'webmock', '~> 3.6', require: false
 end
+
+group :production do
+  # Logs formatter
+  gem 'rails_semantic_logger'
+end
```
tavy88: We only want to enable this in PRD I think.

Gemfile
```diff
@@ -134,3 +134,8 @@ group :test do
   gem 'vcr', '~> 5.0', require: false
   gem 'webmock', '~> 3.6', require: false
 end
+
+group :production do
+  # Logs formatter
+  gem 'rails_semantic_logger', '~> 4.4.3'
```
cwrw: could you change this to `gem 'rails_semantic_logger', '~> 4.4'` as discussed in person

config/environments/production.rb
```diff
@@ -82,6 +82,21 @@
     stdout_logger.formatter = config.log_formatter
     config.logger           = ActiveSupport::TaggedLogging.new(stdout_logger)
 
+    config.rails_semantic_logger.started    = true
+    config.rails_semantic_logger.processing = true
+    config.rails_semantic_logger.rendered   = true
```
mafonso: What is the difference between this

config/environments/production.rb
```diff
@@ -82,6 +82,21 @@
...
+    config.rails_semantic_logger.quiet_assets = true
+    config.rails_semantic_logger.rendered = false
```
mafonso: and this?

config/environments/production.rb
mafonso: Is there anyway these three events could be aggregated, similarly to what LogRage does?
`method=GET path=/jobs/833552.json format=json controller=jobs action=show status=200 duration=58.33 view=40.43 db=15.26`

config/environments/production.rb
```diff
@@ -82,6 +82,21 @@
...
+    STDOUT.sync = true
+    config.rails_semantic_logger.add_file_appender = false
+    config.semantic_logger.add_appender(
+      io: STDOUT,
+      level: config.log_level,
+      formatter: config.rails_semantic_logger.format
+    )
+
     if ENV['LOGGLY_TOKEN'].present?
       loggly_url              = "https://logs-01.loggly.com/inputs/#{ENV['LOGGLY_TOKEN']}/tag/ruby/"
```
mafonso: You should be able to send them in this format to loggly as well, while we have it. The format will likely be very useful on it.

config/environments/production.rb
tavy88: If you check the docs link attached in the description of the PR - it will tell the difference. It's basically deciding which part of the whole process is logged: 

image

config/environments/production.rb
tavy88: Good spot - this came from an else statement I dropped eventually. Should be set to `true`.

config/environments/production.rb
tavy88: Sorry which 3 events are you talking about?

config/environments/production.rb
tavy88: I can't see a LOGGLY_TOKEN setup in PRD so it doesn't look like we're using it there. Let's review this in later refinements? Don't think it's a blocker for this PR.

config/environments/production.rb
mafonso: `started`, `processing` and `rendered`.  LogRage aggregates all these into a single message.  
It's a lot simpler to process and analyse if have all of these in the same log event.
Also, since we "pay" per message and not message size, it's a 30% savings just on this alone.

PR #324, opened 2019-11-14
GET-661 Enable GA tracking for outbound urls

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-661

We need to track clicks that send users to outbound urls, either through
the same window or through the a new tab. Add JS which does the following
taken from:
http://www.joehessert.com/seo/google-analytics-external-link-tracking
https://support.google.com/analytics/answer/7478520?hl=en

1) gets all links (our buttons that link to outside are just button styled
  links)
2) check the domain, and make sure its not a mailto/call link
3) Get the url target (same tab/different tab)
4) Log the click if its an outbound url using gtag event

Testing:
Use the heroku app to navigate to outside urls and look for events on realtime 
Tested on IE11 and seems reasonably ok, will confirm with Mak after his crossbrowser testing
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

### DIFF
--- a/app/webpacker/packs/outbound-link-tracking.js
+++ b/app/webpacker/packs/outbound-link-tracking.js
@@ -1,22 +1,29 @@
 function OutboundLinkTracking () {
   this.start = function () {
+    if (typeof(gtag) !== 'undefined') {
+      trackOutboundLinks();
+    }
+  }
+
+
+  function trackOutboundLinks () {
     if (document.getElementsByTagName) {
       var el = document.getElementsByTagName('a');
       var getDomain = document.domain.split('.').reverse()[1] + '.' + document.domain.split('.').reverse()[0];
 
-      // Look thru each a element
+      // Look through each a element
       for (var i = 0; i < el.length; i++) {
-        // Extract it's href attribute
+        // Extract its href attribute
         var href = (typeof(el[i].getAttribute('href')) == 'string' ) ? el[i].getAttribute('href') : '';
 
         // Query the href for the top level domain (xxxxx.com)
         var myDomain = href.match(getDomain);
 
         // If link is outbound and is not to this domain
-        if (typeof gtag === 'function' && (href.match(/^(https?:|\/\/)/i)  && !myDomain) || href.match(/^mailto\:/i)) {
+        if ((href.match(/^(https?:|\/\/)/i)  && !myDomain) || href.match(/^mailto\:/i)) {
 
           // Add an event to click
-          addEvent(el[i],'click', function(e) {
+          addEvent(el[i], 'click', function(e) {
             var url = this.getAttribute('href');
             var win = (typeof(this.getAttribute('target')) == 'string') ? this.getAttribute('target') : '';
 
--- a/app/webpacker/packs/outbound-link-tracking.js
+++ b/app/webpacker/packs/outbound-link-tracking.js
@@ -1,0 +1,57 @@
+function OutboundLinkTracking () {
+  this.start = function () {
+    if (document.getElementsByTagName) {
+      var el = document.getElementsByTagName('a');
+      var getDomain = document.domain.split('.').reverse()[1] + '.' + document.domain.split('.').reverse()[0];
+
+      // Look thru each a element
+      for (var i = 0; i < el.length; i++) {
+        // Extract it's href attribute
+        var href = (typeof(el[i].getAttribute('href')) == 'string' ) ? el[i].getAttribute('href') : '';
+
+        // Query the href for the top level domain (xxxxx.com)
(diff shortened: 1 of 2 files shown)


app/webpacker/packs/outbound-link-tracking.js
```diff
@@ -0,0 +1,57 @@
+function OutboundLinkTracking () {
+  this.start = function () {
+    if (document.getElementsByTagName) {
+      var el = document.getElementsByTagName('a');
+      var getDomain = document.domain.split('.').reverse()[1] + '.' + document.domain.split('.').reverse()[0];
+
+      // Look thru each a element
+      for (var i = 0; i < el.length; i++) {
+        // Extract it's href attribute
```
smoothcontract: its

app/webpacker/packs/outbound-link-tracking.js
```diff
@@ -0,0 +1,57 @@
...
+        var href = (typeof(el[i].getAttribute('href')) == 'string' ) ? el[i].getAttribute('href') : '';
+
+        // Query the href for the top level domain (xxxxx.com)
+        var myDomain = href.match(getDomain);
+
+        // If link is outbound and is not to this domain
+        if ((href.match(/^(https?:|\/\/)/i)  && !myDomain) || href.match(/^mailto\:/i)) {
+
+          // Add an event to click
+          addEvent(el[i],'click', function(e) {
```
smoothcontract: Sorry to be picky, extra space after the first comma?

app/webpacker/packs/outbound-link-tracking.js
```diff
@@ -0,0 +1,57 @@
...
+            var url = this.getAttribute('href');
+            var win = (typeof(this.getAttribute('target')) == 'string') ? this.getAttribute('target') : '';
```
smoothcontract: Could this not just get the attribute directly?

```
var win = this.getAttribute('target');
```

I tried this in the console, and if the attribute isn't present then `null` is returned. So that value will work fine in the later expression `if (win)` on L29 as expected.

app/webpacker/packs/outbound-link-tracking.js
```diff
@@ -0,0 +1,57 @@
...
+
+      // Look thru each a element
+      for (var i = 0; i < el.length; i++) {
+        // Extract it's href attribute
+        var href = (typeof(el[i].getAttribute('href')) == 'string' ) ? el[i].getAttribute('href') : '';
+
+        // Query the href for the top level domain (xxxxx.com)
+        var myDomain = href.match(getDomain);
+
+        // If link is outbound and is not to this domain
```
smoothcontract: Maybe amend this comment slightly: `If linking to another domain (except via email)` or similar.

app/webpacker/packs/outbound-link-tracking.js
```diff
@@ -0,0 +1,57 @@
+function OutboundLinkTracking () {
+  this.start = function () {
+    if (document.getElementsByTagName) {
+      var el = document.getElementsByTagName('a');
+      var getDomain = document.domain.split('.').reverse()[1] + '.' + document.domain.split('.').reverse()[0];
```
smoothcontract: Not sure I understand this line fully...

```
console.log(document.domain);
> github.com
```

vs

```
console.log(document.domain.split('.').reverse()[1] + '.' + document.domain.split('.').reverse()[0]);
> github.com
```

Why does it need the more complex expression?

app/webpacker/packs/outbound-link-tracking.js
```diff
@@ -0,0 +1,57 @@
...
+      // Look thru each a element
+      for (var i = 0; i < el.length; i++) {
+        // Extract it's href attribute
+        var href = (typeof(el[i].getAttribute('href')) == 'string' ) ? el[i].getAttribute('href') : '';
+
+        // Query the href for the top level domain (xxxxx.com)
+        var myDomain = href.match(getDomain);
+
+        // If link is outbound and is not to this domain
+        if (typeof gtag === 'function' && (href.match(/^(https?:|\/\/)/i)  && !myDomain) || href.match(/^mailto\:/i)) {
```
smoothcontract: This line is getting a little cumbersome now - maybe move the check for `gtag` being defined to somewhere higher in the function (maybe include on L3). I don't think it needs to be tested on every iteration of the loop for each link tag.

app/webpacker/packs/outbound-link-tracking.js
cwrw: yea this is definitely wrong, it yields `org.uk` on our domain, so not sure why its like this, copy pasta. Let me change

PR #325, opened 2019-11-15
Fix caching in Health status endpoint

cwrw:
We are currently caching details, but not the status, which makes internal calls to all endpoints again. Cache the status as well.

Also if an API is down, we are not memoizing the method result `value` as we are rescuing from exceptions, this causes the API to be called multiple times every time we invoke the method `value`. Change that so we only call it once

### DIFF
--- a/app/services/health_check/find_a_job_check.rb
+++ b/app/services/health_check/find_a_job_check.rb
@@ -7,10 +7,7 @@ module HealthCheck
     end
 
     def value
-      @value ||= FindAJobService.new.health_check
-    rescue FindAJobService::APIError => e
-      @output = "Find A Job API error: #{e.message}"
-      ''
+      @value ||= fetch_value
     end
 
     def unit
@@ -19,6 +16,15 @@ module HealthCheck
 
     def status
       value.present? ? :pass : :warn
+    end
+
+    private
+
+    def fetch_value
+      FindAJobService.new.health_check
+    rescue FindAJobService::APIError => e
+      @output = "Find A Job API error: #{e.message}"
+      ''
     end
   end
 end
--- a/app/services/health_check/check_base.rb
+++ b/app/services/health_check/check_base.rb
@@ -10,6 +10,10 @@ module HealthCheck
       raise NotImplementedError
     end
 
+    def cached_status
+      detail[:status]
+    end
+
     def time
       Time.now.iso8601
     end
--- a/app/services/health_check/features_check.rb
+++ b/app/services/health_check/features_check.rb
@@ -7,10 +7,7 @@ module HealthCheck
     end
 
     def value
-      @value ||= Flipflop.health_check?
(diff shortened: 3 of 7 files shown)


smoothcontract:
Actually, there might be an even simpler fix - the key thing is to avoid calling `status` method directly, so perhaps this is sufficient:

```ruby
module HealthCheck
  class CheckBase
    attr_reader :output

    def name
      raise NotImplementedError
    end

    def status
      raise NotImplementedError
    end

    def cached_status
      detail[:status]
    end

    def time
      Time.now.iso8601
    end

    def detail
      Rails.cache.fetch(self.class.name.humanize, expires_in: self.class::CACHE_EXPIRY) do
        {
          status: status,
          time: time
        }.tap do |details|
          details[:metricValue] = value if respond_to?(:value)
          details[:metricUnit] = unit if respond_to?(:unit)
          details[:output] = output if output.present?
        end
      end
    end
  end
end

module HealthCheck
  class ReportService
    def report
      # Build JSON structure according to https://tools.ietf.org/html/draft-inadarei-api-health-check-01
      {
        status: status,
        version: Rails.configuration.git_commit,
        details: details,
        description: 'Get help to retrain service health check'
      }
    end

    def status
      return :pass if statuses.all?(:pass)
      return :fail if statuses.any?(:fail)

      :warn
    end

    def details
      checks.map { |check| [check.name, [check.detail]] }.to_h
    end

    private

    def statuses
      # Key difference - use the cached status value here, not the realtime one
      @statuses ||= checks.map(&:cached_status)
    end

    def checks
      @checks ||= CheckBase.descendants.map(&:new)
    end
  end
end
```

I think I prefer this approach - less invasive, and easier to reason about the behaviour. Thoughts?

app/services/health_check/find_a_job_check.rb
```diff
@@ -20,5 +17,14 @@ def unit
     def status
       value.present? ? :pass : :warn
     end
+
+    private
+
+    def fetch_value
+      FindAJobService.new.health_check
```
tavy88: could this be

```ruby
def fetch_value
  service. health_check
end

def service
  @service ||= FindAJobService.new
end
```

tavy88: Same for the rest of the places.

smoothcontract: I don't think that makes sense, as it's not the instance of the service that we want to memoize, but the response from the health check method.

tavy88: You can have 

```ruby
def service
  FindAJobService.new
end
```

If you don't need memoizing.

smoothcontract: Moving that one line of code to a separate method doesn't feel like a worthwhile change to me - it's easy (perhaps easier) to understand the `fetch_value` implementation as it is now.

PR #328, opened 2019-11-15
Update content as follows:

tavy88:
1. All job results titles are underlined now
2. New subtitle paragraph
3. When a user has already input a job, when adding new ones
one should see Your previous job title on results page
4. Remove job-search-form-group-tag tag class from the form as
it doesn't seem to be needed anymore.

Screen Shot 2019-11-15 at 16 55 03

https://dfedigital.atlassian.net/browse/GET-642

### DIFF
--- a/app/views/shared/search/_results_form.html.erb
+++ b/app/views/shared/search/_results_form.html.erb
@@ -1,6 +1,6 @@
 <div class="govuk-grid-column-full govuk-!-padding-0">
   <%= form_tag search_path, method: 'get' do %>
-    <%= form_group_tag @job_profile_search, :term, tag_class: ['job-search-form-group-tag'] do %>
+    <%= form_group_tag @job_profile_search, :term do %>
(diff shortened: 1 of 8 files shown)


app/views/shared/search/_results_form.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="govuk-grid-column-full govuk-!-padding-0">
   <%= form_tag search_path, method: 'get' do %>
-    <%= form_group_tag @job_profile_search, :term, tag_class: ['job-search-form-group-tag'] do %>
+    <%= form_group_tag @job_profile_search, :term do %>
```
tavy88: The space between the search field and the text under seems to be reduced in the prototype. Awaiting Steve F's answer if that was intentional. For the time being following the prototype. If this is true I don't think we need the CSS class `job-search-form-group-tag`.

PR #330, opened 2019-11-18
GET-692 If sessions are zero we now issue a fail instead of a warn

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-692

### DIFF
--- a/spec/services/health_check/report_service_spec.rb
+++ b/spec/services/health_check/report_service_spec.rb
@@ -54,8 +54,8 @@ RSpec.describe HealthCheck::ReportService do
         create :job_profile
       end
 
-      it 'returns warn status' do
-        expect(service.status).to eq(:warn)
+      it 'returns fail status' do
+        expect(service.status).to eq(:fail)
       end
     end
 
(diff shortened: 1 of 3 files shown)


spec/services/health_check/report_service_spec.rb
```diff
@@ -55,7 +55,7 @@
       end
 
       it 'returns warn status' do
```
tavy88: Update the spec desc here, please.

spec/services/health_check/sessions_check_spec.rb
```diff
@@ -6,13 +6,13 @@
   describe '#status' do
     context 'with no sessions populated' do
       it 'returns warn' do
-        expect(check.status).to eq :warn
+        expect(check.status).to eq :fail
```
tavy88: Same update the desc here.

spec/services/health_check/sessions_check_spec.rb
```diff
@@ -6,13 +6,13 @@
...
       end
 
       it 'returns warn if database not available' do
         allow(Session).to receive(:count).and_raise(PG::ConnectionBad)
 
-        expect(check.status).to eq :warn
+        expect(check.status).to eq :fail
```
tavy88: Same update the desc here.

spec/services/health_check/report_service_spec.rb
```diff
@@ -55,7 +55,7 @@
...
-        expect(service.status).to eq(:warn)
+        expect(service.status).to eq(:fail)
```
tavy88: Update the spec desc here, please.

PR #331, opened 2019-11-18
All outbound links open in the same tab

tavy88:
Now that we have these links tracked in GA, we can drop
the tracked-event data attribute as we don't need to track
them in AppInsights.

NOTE:

We can not yet get rid of TrackEvents() JS function as
we still need to track the interraction on our inpage
smart survey(`shared/feedback/_survey.html.erb`) Once we implement GA backend tracking,
we can drop this script.

Screen Shot 2019-11-18 at 14 38 48

### Testing

Set GOOGLE_ANALYTICS_TRACKING_ID to the workspace you want to use (I used QA) and see the outbound events coming though in real-time.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -14,7 +14,7 @@
         {
           "type": "template",
           "name": "courses/index",
-          "line": 53,
+          "line": 43,
           "file": "app/views/courses/index.html.erb",
           "rendered": {
             "name": "courses/_course",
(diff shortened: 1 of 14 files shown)


config/brakeman.ignore
```diff
@@ -1,5 +1,35 @@
 {
   "ignored_warnings": [
+    {
```
cwrw: can you remove the now deprecated ignored warnings?

PR #333, opened 2019-11-18
Update copy for start page

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-640

Just removed `and` from last bullet point!

### DIFF
--- a/app/views/home/index.html.erb
+++ b/app/views/home/index.html.erb
@@ -9,7 +9,7 @@
       <li>you're employed</li>
       <li>you don't have a degree</li>
       <li>you're aged 24 or over</li>
-      <li>and you're earning below £35,000 a year</li>
+      <li>you're earning below £35,000 a year</li>
     </ul>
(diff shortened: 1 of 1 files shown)


app/views/home/index.html.erb
```diff
@@ -9,7 +9,7 @@
       <li>you're employed</li>
       <li>you don't have a degree</li>
       <li>you're aged 24 or over</li>
-      <li>and you're earning below £35,000 a year</li>
+      <li>you're earning below £35,000 a year</li>
     </ul>
```
cwrw: there is another change on the call line in the prototype, not sure though if that's intentional?

smoothcontract: Not mentioned in Jira ticket, but I'll double check that.

smoothcontract: Confirmed that the prototype is out of date, and should also state `free`. So this is indeed the only change needed for the start page.

PR #334, opened 2019-11-18
Upgrade deployment API version

mafonso:
### Context
https://dfedigital.atlassian.net/browse/GET-748

This PR updates the deployment API version on the service template, as well as updating the ARM schema version.

It is also making calls against the updated APIs on the generic resource templates [here](https://github.com/DFE-Digital/get-help-to-retrain-devops/commit/d414d835feaac0ceb14e61d18a33024358ae4ffd
)

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
     "keyVaultName": {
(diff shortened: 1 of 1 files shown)


azure/template.json
```diff
@@ -264,12 +237,15 @@
         "parameters": {
           "storageAccountName": {
             "value": "[variables('storageAccountName')]"
+          },
+          "storageKind": {
```
cwrw: shouldn't this be `kind`?

azure/template.json
```diff
@@ -224,37 +224,10 @@
...
-      },
-      "dependsOn": [
-        "app-service-plan"
-      ]
-    },
     {
       "name": "storage-account",
       "type": "Microsoft.Resources/deployments",
-      "apiVersion": "2017-05-10",
+      "apiVersion": "2019-08-01",
```
cwrw: isn't the latest `2019-04-01`? https://docs.microsoft.com/en-us/azure/templates/microsoft.storage

azure/template.json
cwrw: I guess the only question is why did this change?

mafonso: Before we were using the default kind which was the only one available. Now we are explicitly passing it (with a default value) so that we can override it to use the v2 kind.

https://docs.microsoft.com/en-us/azure/storage/common/storage-account-overview

PR #336, opened 2019-11-19
GET-637 Do not redirect from skills page when no skills available, also params cleanup

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-637

* If there are no skills stay on page
* If there is no params remain on page
* Cleanup params for search/job profile id as they are no longer required for breadcrumbs or page logic
* Cleanup button params

### DIFF
--- a/app/views/job_profiles/skills/index.html.erb
+++ b/app/views/job_profiles/skills/index.html.erb
@@ -40,7 +40,7 @@
             </div>
           </fieldset>
         <% end %>
-        <%= button_tag('Select these skills', class: 'govuk-button', aria: { label: 'Select these skills button' }, data: { cy: 'select-skills-btn', module: 'govuk-button' }) %>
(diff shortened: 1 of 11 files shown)


app/views/job_profiles/skills/index.html.erb
```diff
@@ -40,7 +40,7 @@
             </div>
           </fieldset>
         <% end %>
-        <%= button_tag('Select these skills', class: 'govuk-button', aria: { label: 'Select these skills button' }, data: { cy: 'select-skills-btn', module: 'govuk-button' }) %>
+        <%= button_tag('Select these skills', class: 'govuk-button', name: nil, aria: { label: 'Select these skills button' }, data: { cy: 'select-skills-btn', module: 'govuk-button' }) %>
```
tavy88: Setting name: nil so you avoid sending the button param. Cool!

PR #346, opened 2019-11-25
GET-473 Spike background jobs

cwrw:
* Use `Delayed Job` as well as `whenever` gem to schedule tasks. 
* Add extra containers for new jobs in docker compose

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -1,3 +1,5 @@
+# set :output, error: 'log/error.log', standard: 'log/standard.log'
+
 every 1.minute do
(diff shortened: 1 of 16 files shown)


config/schedule.rb
```diff
@@ -0,0 +1,5 @@
+# set :output, error: 'log/error.log', standard: 'log/standard.log'
+
+every 1.minute do
+  rake 'scheduled_task:log'
```
smoothcontract: Does this also work if the job is scheduled directly, i.e.

```ruby
TestJob.perform_later
```

PR #347, opened 2019-11-27
Get 725 track backend activity ga

tavy88:
### Context

Replace exisiting server side tracking

- Update logic for the current TrackingService service
to use GA instead of AppInsights
- Ensure backwards compatibility with the sections that
were using AppInsights
- Enable debugging mode for the new services, as the API
will not actually return validation errors (unless debug is ON).
This is intentional so it doesn't interrupt the main application
thread.
For more details please check:
https://developers.google.com/analytics/devguides/collection/protocol/v1/validating-hits

Docs: https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide

### Ticket

https://dfedigital.atlassian.net/browse/GET-793


app/controllers/application_controller.rb
```diff
@@ -24,9 +24,8 @@ def redirect_unless_target_job
 
   private
 
-  def track_event(event_key, properties = {})
-    event = I18n.t(event_key, scope: :events)
-    properties.present? ? TrackingService.track_event(event, properties) : TrackingService.track_event(event)
+  def track_event(properties = {})
```
smoothcontract: Just to recap what we discussed on the call, I'd expect this to change to:

```ruby
def track_event(event_key, value)
```

app/controllers/check_your_skills_controller.rb
```diff
@@ -10,7 +10,7 @@ def index
   def results
     redirect_to(skills_path) if user_session.job_profiles_cap_reached?
 
-    track_event(:check_your_skills_index_search, search: search) if search.present?
+    track_search_event
```
smoothcontract: This should just change to:

```ruby
track_event(:check_your_skills_index_search, search) if search.present?
```

Should be no need for a new method.

app/controllers/courses_controller.rb
```diff
@@ -1,6 +1,6 @@
 class CoursesController < ApplicationController
   def index
-    track_event(:courses_index_search, search: postcode) if postcode.present?
+    track_postcode_event
```
smoothcontract: This should just change to

```ruby
track_event(:courses_index_search, postcode) if postcode.present?
```

Again, no need for a new method.

app/controllers/user_personal_data_controller.rb
```diff
@@ -42,9 +42,12 @@ def check_location_eligibility
 
   def track_location_eligibility_search
```
smoothcontract: May not need this as a separate method as it could be moved to a single line (also the name is somewhat outdated):

```ruby
track_event(:pages_location_eligibility_search, postcode)
```

app/services/tracking_service.rb
```diff
@@ -1,45 +1,75 @@
...
+  attr_reader :response, :ga_tracking_id
+
+  def initialize(ga_tracking_id: Rails.configuration.google_analytics_tracking_id, debug: false)
+    @ga_tracking_id = ga_tracking_id
+    @debug = debug
   end
 
-  def self.key
-    @key ||= Rails.configuration.app_insights_instrumentation_key
+  def track_event(props = {})
```
smoothcontract: This is a leaky abstraction - the client consuming the service is required to know and understand the payload values that Google Analytics expects. Instead we should pass service agnostic parameters such as `event` and `value` that work irrespective of the fact we're currently using GA for analytics. The service then has responsibility to build the correct payload including those values.

smoothcontract: Also, when we get to track multiple values such as the checked skills, one idea is to simple accept an array `value` - if that datatype is passed then the service can invoke the batch endpoint multiple times, once for each entry in the array.

Assuming the batch endpoint isn't any slower, it might be simpler to use that same endpoint for everything - even if a single value is passed. Maybe something to review with the google documentation.

tavy88: Let's take the batching out from the scope of this PR. As agreed on the call, I would like this to just replace the existing tracking. Batching will come with its own complexity.

app/controllers/application_controller.rb
```diff
@@ -24,9 +24,12 @@ def redirect_unless_target_job
 
   private
 
-  def track_event(event_key, properties = {})
-    event = I18n.t(event_key, scope: :events)
-    properties.present? ? TrackingService.track_event(event, properties) : TrackingService.track_event(event)
+  def track_event(event_key, value = nil)
+    event_label = I18n.t(event_key, scope: :events)
+
+    TrackingService.new.track_event(key: event_key, label: event_label, value: value)
```
tavy88: I appreciate there is a bit of redundancy here for key, label, but I don't want to move the locale access at the service level. This way the service remains flexible for labels that are not necessarily localised. However, in our case this should not be allowed.

app/controllers/check_your_skills_controller.rb
tavy88: Yeah already did that change but you must have commented before I pushed.

app/controllers/courses_controller.rb
tavy88: Yeah already did that change but you must have commented before I pushed.

app/controllers/user_personal_data_controller.rb
tavy88: Yeah already did that change but you must have commented before I pushed.

app/controllers/application_controller.rb
tavy88: Yeah already did a similar change but you must have commented before I pushed.

app/services/tracking_service.rb
smoothcontract: Okay, happy to leave batching out for now. But I would prefer to see named parameters here of `key`, `label` and `value`. Looks from the implementation below that these are all mandatory, and so it's clearer if the interface specifies these explicitly; it's important documentation. These three values can then be instance variables or part of a props hash internally, that doesn't really matter so long as the interface is explicit.

app/services/tracking_service.rb
```diff
@@ -1,45 +1,78 @@
...
-      # send telemetry to the service in batches of 5
-      tc.channel.sender.send_buffer_size = 5
-      # the background worker thread will be active for 5 seconds before it shuts down. if
-      # during this time items are picked up from the queue, the timer is reset.
-      tc.channel.sender.send_time = 5
-      # the background worker thread will poll the queue every 0.5 seconds for new items
-      tc.channel.sender.send_interval = 0.5
-    end
+  def send_ga_event(props)
+    @response = Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == 'https') { |http|
```
smoothcontract: Don't think the optional `uri.scheme` stuff is needed, as both URL's use https. Also this should use block delimiters rather than braces as it's over multiple lines. If Rubocop disagrees then you might need to fiddle with the rules a bit. So something like:

```ruby
@response = Net::HTTP.start(uri.host, uri.port, use_ssl: true) do |http|
  request.set_form_data(build_payload(props))
  http.request(request)
end
```

app/services/tracking_service.rb
tavy88: The problem is that the `label` is not mandatory :) That's why I kept it flexible.

app/services/tracking_service.rb
tavy88: We've had this issue before a couple of times and as you remember it's Rubocop complaining. What tweak do you have in mind? Can you give me an example?

tavy88: Also in regards to the `uri.scheme` I think ti's good practice to enforce it. BingSpellCheck service is using the same approach and that got approved.

app/services/tracking_service.rb
smoothcontract: That shouldn't matter, in fact it makes an even stronger case for named parameters. So you could go with this - it's a minimal change but makes the code easier to understand in future:

```ruby
def track_event(key:, value:, label: nil)
  return unless ga_tracking_id.present?

  send_ga_event(key: key, value: value, label: label)
  ...
```

That documents the two mandatory parameters and the optional label quite nicely.

app/services/tracking_service.rb
smoothcontract: We do have code that uses multiline blocks with do/end syntax and rubocop is passing - so you'll need to do some investigation.

app/services/tracking_service.rb
```diff
@@ -1,45 +1,78 @@
...
+
+  def initialize(ga_tracking_id: Rails.configuration.google_analytics_tracking_id, debug: false)
+    @ga_tracking_id = ga_tracking_id
+    @debug = debug
   end
 
-  def self.key
-    @key ||= Rails.configuration.app_insights_instrumentation_key
+  def track_event(key:, label:, value:)
+    return unless ga_tracking_id && [key, label, value].all?
```
smoothcontract: I thought you said that label was optional? This is enforcing that all values are passed. In which case why not just make them mandatory parameters and simplify this line?

app/controllers/application_controller.rb
smoothcontract: This is fine, no issues here.

app/controllers/application_controller.rb
```diff
@@ -24,9 +24,12 @@ def redirect_unless_target_job
...
+  rescue TrackingService::TrackingServiceError
+    nil
```
smoothcontract: Is this line redundant?

app/services/tracking_service.rb
tavy88: I decided to make the label mandatory within our context.  We won't be sending unlabelled events as it would increase the difficulty for reporting. The label is not mandatory for the GA API, but it is for us.

app/controllers/application_controller.rb
tavy88: It's a pattern we seem to follow: 

https://github.com/DFE-Digital/get-help-to-retrain/blob/a14a394ad9105072c868e5007a4986accec1cf54/app/lib/flipflop/strategies/split_client_strategy.rb#L24

https://github.com/DFE-Digital/get-help-to-retrain/blob/a14a394ad9105072c868e5007a4986accec1cf54/app/controllers/job_profiles_controller.rb#L30

I assume it's the same situation?

app/services/tracking_service.rb
tavy88: You need at least 3 lines of code I think before Rubocop gets silenced. Is this really a blocker for the PR?

app/services/tracking_service.rb
tavy88: I decided to make the label mandatory within our context. We won't be sending unlabelled events as it would increase the difficulty for reporting. The label is not mandatory for the GA API, but it is for us.

app/controllers/application_controller.rb
smoothcontract: Not quite. We don't test the return value of `track_event` anywhere so don't really care what the return value is. For a `job_vacancy_count` we'd definitely want 0 or nil something defined.

app/services/tracking_service.rb
smoothcontract: Fair enough, so why not change the first couple of lines to make this super explicit (and simpler implementation):

```ruby
def track_event(:key, :label, :value)
  return unless ga_tracking_id.present?
```

This will then throw an ArgumentError if called without passing all three named arguments - which is preferable to checking within the method that all three were passed.

app/controllers/application_controller.rb
tavy88: Are you saying:

```def job_vacancy_count
    JobVacancySearch.new(
      name: resource.name,
      postcode: user_session.postcode
    ).count
  rescue FindAJobService::APIError
  end``` would still not return `nil` if there was an excpetion rescued?

File: app/services/tracking_service.rb (hunk 8 above)
Comment (tavy88) [reply app/services/tracking_service.rb]: Does it?

```ruby
SyntaxError:
  get-help-to-retrain/app/services/tracking_service.rb:14: syntax error, unexpected tSYMBEG, expecting ')'
    def track_event(:key, :label, :value)
```

app/controllers/application_controller.rb
tavy88: In addition, Rubocop will complain if code is left this way.

smoothcontract: okay

app/services/tracking_service.rb
smoothcontract: I had the syntax wrong, but the arguments are still required - hence checking that they are present is redundant. This should work:

```ruby
def track_event(key:, label:, value:)
  return unless ga_tracking_id.present?
```

tavy88: But if you pass

```ruby
track_event(key: nil, label: 1, value: 2)
```
I don't think it will.

cwrw: Hmm this is interesting, I'm not sure failing silently is going to help us here, we need to know something is wrong and if we are sending empty values

app/services/tracking_service.rb
```diff
@@ -1,45 +1,74 @@
...
+
+    send_event(key, label, value)
+  rescue StandardError => e
+    Rails.logger.error("Tracking service error: #{e.message}")
```
cwrw: can we not use `handle_api_errors` here instead?

cwrw: Also won't we be throwing an error twice at this point?

app/services/tracking_service.rb
```diff
@@ -1,45 +1,74 @@
...
+    @uri ||= begin
+      return URI.parse(DEBUG_API_ENDPOINT) if debugging_enabled?
+
+      URI.parse(API_ENDPOINT)
+    end
   end
 
-  def self.queue
-    @queue ||= ApplicationInsights::Channel::AsynchronousQueue.new(sender)
+  def request
```
cwrw: this doesn't seem to be used anywhere?

.env.test
```diff
@@ -1,4 +1,4 @@
 USER_FEEDBACK_SMART_SURVEY_LINK=https://www.smartsurvey.co.uk/s/8IBQ1/
 NOTIFY_API_KEY=test
-GOOGLE_ANALYTICS_TRACKING_ID=
 BING_SPELL_CHECK_API_KEY=test
+GOOGLE_ANALYTICS_TRACKING_ID=test
```
cwrw: I would suggest not setting this, as we will be making real requests in the tests unfortunately

spec/services/tracking_service_spec.rb
```diff
@@ -1,36 +1,131 @@
...
+      ea: 'Event action'
+    }
+  }
+
+  def fake_ga_request_with(payload, debug: false)
+    api_endpoint = debug ? TrackingService::DEBUG_API_ENDPOINT : TrackingService::API_ENDPOINT
+
+    stub_request(:post, api_endpoint)
+      .with(body: URI.encode_www_form(payload))
+      .to_return(body: ga_response, status: 200)
```
cwrw: can you add a test for a response that isn't a 200?

tavy88: As mentioned on the call we had yesterday, GA does not give any other response than 200 right now even if you have validation errors.

app/services/tracking_service.rb
tavy88: Yeah not anymore after my last commit.

app/controllers/user_personal_data_controller.rb
```diff
@@ -30,7 +30,8 @@ def personal_data_params
   end
 
   def check_location_eligibility
-    track_location_eligibility_search
+    track_event(:pages_location_eligibility_search, postcode)
```
cwrw: change name of tag as it's no longer valid

app/services/tracking_service.rb
cwrw: discussed in person, this is fine, just need a new type of error

app/controllers/user_personal_data_controller.rb
cwrw: discussed in person, keep same but flag to business

app/services/tracking_service.rb
tavy88: I will keep it as is now - will think about it when I do the batching.

app/services/tracking_service.rb
tavy88: I decided to remove this bit actually as GA is not throwing any errors.

PR #348, opened 2019-11-29
Add support for searching for a specific job profile in skills matcher

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-727

* Refactored existing functionality for job profile search to allow common code in a concern (DRY).
* Made 50 limit in skills matcher results optional, as otherwise we risk not showing match scores for searched job profiles.
* Refactored existing partials for skills builder search to allow use within skills matcher
* Refactored existing partials for skills matcher results to allow use for explicit job profile search (and made these more idiomatic).
* Cleaned up decoration of job profile collections with a new method in JobProfileDecorator
* Tweaked wording for no results page to be generic enough to be used in both skill builder and matcher scenarios
* Added feature coverage

### Guidance to review
I didn't go mad on testing every aspect of the skills match and job profile search functionality as those are already extensively tested; this is largely just reusing existing functionality. If any more test scenarios make sense without repeating what we already have then I'm open to suggestions.


app/models/skills_matcher.rb
```diff
@@ -1,8 +1,9 @@
 class SkillsMatcher
-  attr_reader :user_session, :job_profile_id
+  attr_reader :user_session, :options
```
smoothcontract: The `:job_profile_id` attribute was not used - looks like old code that didn't get cleaned up.

app/models/skills_matcher.rb
```diff
@@ -37,7 +38,7 @@ def build_query
         .from(Arel.sql("(#{skills_matched_query}) as ranked_job_profiles"))
         .joins('LEFT JOIN job_profiles ON job_profiles.id = ranked_job_profiles.job_profile_id')
         .order(skills_matched: :desc, growth: :desc, name: :asc)
-        .limit(50)
+        .limit(options[:limit])
```
smoothcontract: Interestingly `.limit(nil)` does not apply a limit, which is nice as it makes this syntax really simple :)

app/views/job_profiles/_job_profile.html.erb
```diff
@@ -0,0 +1,22 @@
+<li class="govuk-!-padding-bottom-1">
```
smoothcontract: This is essentially the previous job_matches partial from skills matcher (minus the wrapper `ul` tags)

app/views/job_profiles/results.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <div class="govuk-grid-column-two-thirds" >
+    <% if @job_profiles.blank? %>
+      <%= render '/shared/search/no_results', scope: 'job_profiles.results' %>
+    <% else %>
+      <%= render "search_form" %>
+      <ul class="govuk-list">
+        <%= render @job_profiles %>
+      </ul>
+      <div class="govuk-grid-column-full govuk-!-padding-top-6 govuk-!-margin-bottom-4 govuk-!-margin-top-6 pagination__section">
+        <%= paginate @results %>
```
smoothcontract: `@results` and `@job_profiles` have the same number of elements, but the decorated version doesn't include necessary keys for kaminari pagination. The same pattern is used for skills matcher.

app/views/shared/search/_no_results.html.erb
```diff
@@ -1,8 +1,8 @@
 <h3 class="govuk-heading-xl">No results found</h3>
-<%= render "shared/search/results_form", search_path: results_check_your_skills_path, scope: 'check_your_skills.results' %>
+<%= render "shared/search/results_form", search_path: { action: :results }, scope: scope %>
```
smoothcontract: Tweaked this a bit to make it more generic - although it's in the `shared` folder it was still pretty tightly coupled to skills builder.

spec/models/skills_matcher_spec.rb
```diff
@@ -10,19 +10,19 @@
       expect(matcher.match).to be_empty
     end
 
-    it 'limits maximum number of job matches to 50' do
+    it 'limits maximum number of job matches' do
```
smoothcontract: As the limit is now passed as an option I took the opportunity to create fewer test jobs in this spec.

app/views/job_profiles/_search_form.html.erb
```diff
@@ -0,0 +1,3 @@
+<h1 class="govuk-heading-xl govuk-!-margin-bottom-7"><%= t(:title, scope: 'job_profiles.results') %></h1>
+<%= render "shared/search/results_form", search_path: results_job_profiles_path, scope: 'job_profiles.results' %>
+<p class="govuk-body-l govuk-!-margin-bottom-2"><%= t(:body, scope: 'job_profiles.results') %></p>
```
cwrw: Need to amend this partial to follow the view as seen here: https://get-help-to-retrain-prototype.herokuapp.com/main/release5_e2e_v2/specific_job_search_results
where the body should be above the search form, and we should include `result found`

app/controllers/check_your_skills_controller.rb
```diff
@@ -1,33 +1,27 @@
 class CheckYourSkillsController < ApplicationController
+  include SearchableJobProfile
+
   def index
     redirect_to(skills_path) if user_session.job_profiles_cap_reached?
 
-    @job_profile_search = JobProfileSearch.new(term: search, profile_ids_to_exclude: profile_ids_to_exclude)
+    build_job_profile_search
```
cwrw: we talked about refactoring this logic a while ago to avoid the unexpected behaviour, do you think you could do this in this ticket or is it adding too much scope?

smoothcontract: Are you referring to the search results ordering? If so it's more appropriate to tackle that in this ticket, which I plan to work on next: https://dfedigital.atlassian.net/browse/GET-775

app/controllers/job_profiles_controller.rb
```diff
@@ -1,4 +1,21 @@
 class JobProfilesController < ApplicationController
+  include SearchableJobProfile
+
+  def index
+    build_job_profile_search
```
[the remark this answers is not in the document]
smoothcontract: Not entirely sure it's necessary as it works fine with no skills in the session - obviously no skills match in that scenario. I could be persuaded if you think it's worthwhile?

app/views/job_profiles/_search_form.html.erb
smoothcontract: Good catch, I was deep in the mindset of partial re-use 😁

app/controllers/job_profiles_controller.rb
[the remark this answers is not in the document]
cwrw: Although it won't break the page, I'm thinking since its part of the skills matcher journey after user skills have been selected might be better to guard against it

app/controllers/check_your_skills_controller.rb
cwrw: I was thinking more around: https://dfedigital.atlassian.net/browse/GET-788
But there is a ticket for it already if you would rather defer this

smoothcontract: Yeah, I'd rather not increase the scope of this further.

PR #349, opened 2019-11-29
GET-460 Add In page feedback survey form

cwrw:
https://dfedigital.atlassian.net/browse/GET-460

* Add a new table in restricted database to capture information
* Disable form if javascript disabled
* Add two ajax forms, one for the "yes" and one for the "no"
* Add accessibility for survey
* Remove smart survey


app/controllers/feedback_surveys_controller.rb
```diff
@@ -0,0 +1,12 @@
+class FeedbackSurveysController < ApplicationController
+  def create
+    @survey = FeedbackSurvey.new(feedback_survey_params)
+    render :show if @survey.save
```
cwrw: TODO: what happens when form fails

spec/features/user_feedback_survey_spec.rb
```diff
@@ -1,31 +1,62 @@
...
-  scenario 'Feedback form is not present when disabled in configuration' do
-    allow(Rails.configuration).to receive(:smart_survey_user_feedback_link).and_return(nil)
+  scenario 'User gets a thank you message when clicking yes', :js do
+    answer_yes
 
-    visit(root_path)
+    expect(page).to have_text('Thank you for your feedback')
+  end
+
+  scenario 'Page helpful survey is persisted when answering yes', :js do
```
cwrw: should this test be here or in request specs, I prefer here

app/controllers/feedback_surveys_controller.rb
smoothcontract: I don't see how the `save` operation could fail - there's only one validated parameter and no way for that to be omitted in the UI. So I think it's safe to replace this with

```ruby
  @survey.save
  render :show
```

app/views/shared/feedback/_survey.html.erb
```diff
@@ -1,36 +1,40 @@
...
-    <div id="feedback-prompt-success" class="feedback-prompt-content hidden">
-      Thank you for your feedback
-    </div>
+  <div id="feedback-not-useful-container" class="top-border__container  govuk-!-padding-top-2 hidden">
+    <%= link_to 'close', '#', class: 'govuk-body govuk-link close-feedback-survey', role: 'button', aria: { controls: 'page-is-not-useful', expanded: true } %>
+    <%= form_with url: feedback_surveys_path, id: 'page-is-not-useful-form', aria: { hidden: true } do |f| %>
+      <%= hidden_field_tag('page_useful', 'no') %>
+      <div class="govuk-form-group govuk-!-margin-top-6">
+        <%= label_tag :feedback_survey_message, 'How should we improve this page?', class: 'govuk-label govuk-!-margin-bottom-4' %>
+        <%= text_area_tag :message, nil, class: 'govuk-textarea govuk-!-width-two-thirds', rows: 3 %>
```
smoothcontract: Is it possible to restrict the maximum length of text here? I suspect a long enough text string would cause Postgres to break - alternatively perhaps add a length validation on the model.

app/controllers/feedback_surveys_controller.rb
smoothcontract: Actually it doesn't look like the instance variable is used in the views anywhere, so could simplify further:

```ruby
  FeedbackSurvey.new(feedback_survey_params).save
  render :show
```

Also I'm wondering why not just rename the JS view to `create.js.erb` and then this reduces to a single line:

```ruby
  FeedbackSurvey.new(feedback_survey_params).save
```

db/restricted_schema.rb
```diff
@@ -39,15 +46,6 @@
     t.index ["updated_at"], name: "index_sessions_on_updated_at"
   end
 
-  create_table "users", force: :cascade do |t|
```
smoothcontract: This change seems unrelated - could you revert please?

spec/features/user_feedback_survey_spec.rb
smoothcontract: This fits well as a feature spec

app/controllers/feedback_surveys_controller.rb
cwrw: I was in two minds on using `create` instead of `show`, as we are saving the survey and then rendering something indicating it's successful, but if we don't care about that then maybe we can use `create` instead like you suggest 🤔 will have a look

smoothcontract: It doesn’t actually show the survey, so I think create is a better fit for this specific example

app/views/feedback_surveys/create.js.erb
```diff
@@ -0,0 +1 @@
+document.querySelector('#feedback-survey-container').innerHTML = "<%= escape_javascript(render partial: 'success')%>";
```
tavy88: You can use `j(render partial: 'success')` here. https://api.rubyonrails.org/classes/ActionView/Helpers/JavaScriptHelper.html#method-i-j

app/views/shared/feedback/_survey.html.erb
```diff
@@ -1,36 +1,40 @@
...
-        data: {
-          tracked_event: 'User Feedback On Page Survey',
-          tracked_event_props: { 'answer': 'yes' }
-        },
-      } do %>
-        Yes
+      <%= form_with url: feedback_surveys_path, class: 'feedback-survey-form', id: 'page-is-useful-form', aria: { hidden: false } do |f| %>
+        <%= hidden_field_tag('page_useful', 'yes') %>
+        <%= submit_tag('Yes', role: 'button', class: 'unbutton-survey feedback-prompt-link', data: { tracked_event: 'User Feedback On Page Survey', tracked_event_props: { 'answer': 'yes' } }) %>
+        <span class="govuk-visually-hidden">this page is useful</span>
```
tavy88: Hmm, I was hoping we would be storing the answers as well yes/no in the same table so we won't have to do this tracking anymore? It would be a bit redundant to continue with server-side tracking now that we store everything on our db.

```ruby
so table structure:

answer    message
yes               nil
no                 Reason why...
```

spec/features/user_feedback_survey_spec.rb
```diff
@@ -1,31 +1,62 @@
...
+    answer_yes
+    click_on('Start now')
 
-    expect(page).not_to have_text('Is this page useful?')
+    expect(FeedbackSurvey.first.attributes).to include(
+      'page_useful' => 'yes'
+    )
```
tavy88: Hmm not sure about this one. I don't think we should be testing AR behaviour at this level?

I would just expect to see if the message is present or not on the page.

spec/features/user_feedback_survey_spec.rb
```diff
@@ -1,31 +1,62 @@
...
+    answer_no
+    fill_in('message', with: 'Page was not useful')
+    click_on('Send message')
+    click_on('Start now')
 
-    click_on('Accept cookies')
+    expect(FeedbackSurvey.first.attributes).to include(
+      'message' => 'Page was not useful',
+      'page_useful' => 'no'
+    )
```
tavy88: Hmm not sure about this one. I don't think we should be testing AR behaviour at this level?

I would just expect to see if the message is present or not on the page.

tavy88: I think this should be a unit test at the model level.

spec/features/user_feedback_survey_spec.rb
tavy88: I think this should be a unit test at the model level.

app/views/feedback_surveys/create.js.erb
smoothcontract: I think also you can skip the `partial` keyword too - in most cases Rails is smart enough to realise you meant the partial. Presumably same thing works with JS views too. So would have:

```ruby
"<%= j(render 'success') %>"
```

app/views/shared/feedback/_survey.html.erb
cwrw: I left the tracking here as was waiting for your pr to get merged, also I think removing this tracking will fall into the ticket that removes AAI, where we can have a think about where this fits in. Not sure I understand the comments about the table structure though, is it not already  looking like that?

smoothcontract: It does store the answers as yes/no in the same table. Here are some examples from my testing:

```
=> #<FeedbackSurvey id: 3, page_useful: "no", message: "", created_at: "2019-11-29 12:27:56", updated_at: "2019-11-29 12:27:56">
irb(main):002:0> FeedbackSurvey.last
=> #<FeedbackSurvey id: 4, page_useful: "no", message: "", created_at: "2019-12-02 09:23:54", updated_at: "2019-12-02 09:23:54">
irb(main):003:0> FeedbackSurvey.last
=> #<FeedbackSurvey id: 5, page_useful: "no", message: "xxxxxx", created_at: "2019-12-02 09:24:05", updated_at: "2019-12-02 09:24:05">
irb(main):004:0> FeedbackSurvey.last
=> #<FeedbackSurvey id: 6, page_useful: "yes", message: nil, created_at: "2019-12-02 09:24:12", updated_at: "2019-12-02 09:24:12">
irb(main):005:0> FeedbackSurvey.last
=> #<FeedbackSurvey id: 7, page_useful: "no", message: "01234567890123456789012345678901234567890123456789...", created_at: "2019-12-02 09:24:59", updated_at: "2019-12-02 09:24:59">
```

This is performing client side tracking of yes/no clicks so is a different use case to analysing the survey responses. I think it's useful to still have this in GA until we confirm otherwise. One caveat - I didn't test this aspect.

spec/features/user_feedback_survey_spec.rb
smoothcontract: I think a model test would be pretty superfluous, and prefer the approach taken here. We generally don't include unit tests for setting model attributes - that's part of the Rails framework and we can just expect that to work.

spec/features/user_feedback_survey_spec.rb
smoothcontract: Disagree - see above comment.

spec/features/user_feedback_survey_spec.rb
tavy88: Can you show me another feature spec where we test setting model attributes like this?

app/views/shared/feedback/_survey.html.erb
tavy88: > I left the tracking here as was waiting for your pr to get merged, also I think removing this tracking will fall into the ticket that removes AAI, where we can have a think about where this fits in. Not sure I understand the comments about the table structure though, is it not already looking like that?

Yep sorry it's all in there. Missed that bit.

tavy88: > It does store the answers as yes/no in the same table. Here are some examples from my testing:
> 
> ```
> => #<FeedbackSurvey id: 3, page_useful: "no", message: "", created_at: "2019-11-29 12:27:56", updated_at: "2019-11-29 12:27:56">
> irb(main):002:0> FeedbackSurvey.last
> => #<FeedbackSurvey id: 4, page_useful: "no", message: "", created_at: "2019-12-02 09:23:54", updated_at: "2019-12-02 09:23:54">
> irb(main):003:0> FeedbackSurvey.last
> => #<FeedbackSurvey id: 5, page_useful: "no", message: "xxxxxx", created_at: "2019-12-02 09:24:05", updated_at: "2019-12-02 09:24:05">
> irb(main):004:0> FeedbackSurvey.last
> => #<FeedbackSurvey id: 6, page_useful: "yes", message: nil, created_at: "2019-12-02 09:24:12", updated_at: "2019-12-02 09:24:12">
> irb(main):005:0> FeedbackSurvey.last
> => #<FeedbackSurvey id: 7, page_useful: "no", message: "01234567890123456789012345678901234567890123456789...", created_at: "2019-12-02 09:24:59", updated_at: "2019-12-02 09:24:59">
> ```
> 
> This is performing client side tracking of yes/no clicks so is a different use case to analysing the survey responses. I think it's useful to still have this in GA until we confirm otherwise. One caveat - I didn't test this aspect.

I saw the table structure and replied back. I missed that bit. I don't agree that we need GA tracking client-side anymore. I don't really see the point now that we track that server-side. We had to do client side tracking because we couldn't store the answers in SmartSurvey. That was the only reason.

spec/features/user_feedback_survey_spec.rb
tavy88: I did reply.

app/views/shared/feedback/_survey.html.erb
cwrw: discussed in person, will remove AAI, and use the table to get the numbers, will also discuss with Lois

spec/features/user_feedback_survey_spec.rb
cwrw: Discussed in person, leaving as is as this is a special usecase

app/views/shared/feedback/_survey.html.erb
tavy88: 👍

PR #351, opened 2019-12-02
Add batch processing to the TrackingService

tavy88:
### Description
Switch to GA server side tracking for user events

1) Replace AppInsights server side tracking with GA
server side tracking
2) Allow batch requests sending
3) Tracked sections:
- skills search
- postcode search
- courses search
- skills builder
- training questions
- job_hunting questions

Screen Shot 2019-12-04 at 09 51 31

### Approach

Decided to go with the following approach:

```ruby
track_events(:key, props), where props is an array of props defined by a (label, value) pair.
```
I do think it's really intuitive from a usage perspective. Can't think of a better way of doing it.

https://dfedigital.atlassian.net/browse/GET-794

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -29,9 +29,9 @@ class ApplicationController < ActionController::Base
     event_label = I18n.t(event_key, scope: :events)
 
     track_events(
-      event_key,
       [
         {
+          key: event_key,
           label: event_label,
           value: event_value
         }
@@ -39,8 +39,8 @@ class ApplicationController < ActionController::Base
     )
   end
 
-  def track_events(event_key, props = [])
-    TrackingService.new.track_events(key: event_key, props: props)
+  def track_events(props = [])
+    TrackingService.new.track_events(props: props)
   rescue TrackingService::TrackingServiceError
     nil
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   include Passwordless::ControllerHelpers
+  include GaTrackingHelper
 
   before_action :set_raven_context
 
@@ -24,10 +25,22 @@ class ApplicationController < ActionController::Base
 
   private
 
-  def track_event(event_key, value = nil)
+  def track_event(event_key, event_value)
     event_label = I18n.t(event_key, scope: :events)
 
-    TrackingService.new.track_event(key: event_key, label: event_label, value: value)
+    track_events(
+      event_key,
+      [
+        {
+          label: event_label,
+          value: event_value
+        }
+      ]
+    )
+  end
+
+  def track_events(event_key, props = [])
+    TrackingService.new.track_events(key: event_key, props: props)
   rescue TrackingService::TrackingServiceError
     nil
   end
--- a/app/controllers/job_profiles_skills_controller.rb
+++ b/app/controllers/job_profiles_skills_controller.rb
@@ -6,7 +6,7 @@ class JobProfilesSkillsController < ApplicationController
       skills_builder.build
 
       if skills_valid?
-        track_selections(:skills_builder, selected: selected_skills, unselected: unselected_skills)
+        track_selections(selected: selected_skills, unselected: unselected_skills)
 
         return redirect_to skills_path
       end
@@ -53,14 +53,16 @@ class JobProfilesSkillsController < ApplicationController
 
   def selected_skills
     {
-      label: "#{job_profile.name} - Ticked",
+      key: :skills_builder_ticked,
+      label: job_profile.name,
       values: available_skills.slice(*user_skill_ids).values
     }
   end
 
   def unselected_skills
     {
-      label: "#{job_profile.name} - Unticked",
+      key: :skills_builder_unticked,
+      label: job_profile.name,
       values: available_skills.except(*user_skill_ids).values
     }
   end
--- a/app/controllers/job_profiles_skills_controller.rb
+++ b/app/controllers/job_profiles_skills_controller.rb
@@ -5,7 +5,11 @@ class JobProfilesSkillsController < ApplicationController
     else
       skills_builder.build
 
-      return redirect_to skills_path if skills_valid?
+      if skills_valid?
+        track_selections(:skills_builder, selected: selected_skills, unselected: unselected_skills)
+
+        return redirect_to skills_path
+      end
 
       render('job_profiles/skills/index')
     end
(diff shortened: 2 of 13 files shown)


app/services/tracking_service.rb
```diff
@@ -11,27 +15,31 @@ def initialize(ga_tracking_id: Rails.configuration.google_analytics_tracking_id,
...
-      URI.parse(API_ENDPOINT)
-    end
+  def batch_uri
+    URI.parse(BATCH_API_ENDPOINT)
   end
 
-  def send_event(key, label, value)
-    ::Net::HTTP.post(uri, build_payload(key, label, value)).body
+  def send_events(key, props)
+    raise BatchTooBigError, "Batch size cannot be over #{MAX_BATCH_SIZE}" if props.size > MAX_BATCH_SIZE
```
cwrw: Instead of raising an error, if we get batches more than 20, we can stagger them to send 20. For instance if I receive 23 events, I can sent 20 then 3 and not throw an error

app/controllers/job_profiles_skills_controller.rb
```diff
@@ -33,7 +36,33 @@ def skills_params
...
+      events_props
+    )
+  end
+
   def profile_skills_not_editable?
     user_session.job_profile_ids.exclude?(job_profile.id)
   end
+
+  def user_skills
+    @user_skills ||= Skill.find(skills_params[:skill_ids].reject(&:empty?))
```
cwrw: instead of doing the extra query here for find skills again, we can build a hash of  ids from the `job_profile.skills` as so:
```
job_profile_skills = job_profile.skills.each_with_object({}) {|skill, hash| hash[skill.id] = skill.name }
```
then to diff them out we just delete them from the hash:
```
job_profile_skills.except!(skills_params[:skill_ids].reject(&:empty?))
```
After that we have the values we expect to be diffed, and we can use that to get the name as well for the event props

app/controllers/questions_controller.rb
```diff
@@ -1,22 +1,69 @@
 class QuestionsController < ApplicationController
   before_action :redirect_unless_target_job
 
+  AVAILABLE_TRAINING_OPTIONS = %w[english_skills math_skills].freeze
```
cwrw: we can have a hash here instead, mapping the training to the value, which means we don't need if/else statments but just do a lookup instead

app/controllers/questions_controller.rb
```diff
@@ -1,22 +1,69 @@
...
     user_session.job_hunting = questions_params[:job_hunting] || []
+
+    track_answers_for(:job_hunting)
+
     redirect_to action_plan_path
   end
 
   private
 
+  def track_answers_for(section)
```
cwrw: not sure this warrants a new method, we could call `track_events(section, events_props_for(section))` everywhere instead and avoid having another method call

app/controllers/application_controller.rb
```diff
@@ -24,10 +24,22 @@ def redirect_unless_target_job
...
+    }
+
+    track_events(
+      event_key,
+      [event_props]
+    )
+  end
+
+  def track_events(event_key, props = [])
+    TrackingService.new.track_events(key: event_key, props: props)
```
cwrw: there seems to be a theme we are doing in tracking everywhere which does this mapping:
```
some_array.map do |value|
      {
        label: I18n.t(state, event_label: 'label'),
        value: value
      }
    end
```
can we not move the mapping here instead, so we just pass arrays and allow the method to hide any tracking implementation?

app/controllers/questions_controller.rb
```diff
@@ -1,22 +1,69 @@
...
+    track_events(section, events_props_for(section))
+  end
+
+  def events_props_for(section)
+    selected_options_for(section) + rejected_options_for(section)
+  end
+
+  def selected_options_for(section)
```
cwrw: I found this tracking section hard to follow, I think moving the mapping to the `track_events` will help hide the tracking complexity and keep controllers cleaner

app/services/tracking_service.rb
tavy88: I would keep it this way for simplicity. Your suggestion adds unnecessary complication. Besides our current payload size is max 9 atm.

app/controllers/job_profiles_skills_controller.rb
tavy88: Sorry which extra query are you talking about?

app/controllers/questions_controller.rb
tavy88: I think it's cleaner this way. 

track_events(section, events_props_for(section)) exposes to much internal logic at that controller level.

app/controllers/application_controller.rb
tavy88: It's not always the same if you check the skills tracking.

app/controllers/job_profiles_skills_controller.rb
cwrw: `Skill.find`, the line this comment is on

app/controllers/questions_controller.rb
tavy88: Ok.

app/services/tracking_service.rb
cwrw: Since this is the first iteration, Im ok to wait till we can leverage `Delayed::Job` when added to handle multiple API requests as we don't want that now synchronously, but I think it's something we should keep in mind regardless. It just means we have to be careful not to send more than batches of 20 for now

app/controllers/job_profiles_skills_controller.rb
tavy88: Sorry, my question was extra to which other query?

cwrw: so currently we are doing two queries: `job_profile.skills` and `Skills.find`. If we build a hash map of the main skills set using `job_profile.skills`, we won't need to call `Skills.find`

app/controllers/application_controller.rb
tavy88: Anyway I've done a bit more optimisation.

app/controllers/job_profiles_skills_controller.rb
tavy88: I've done a bit more optimisation.

tavy88:
I think if you look at app/controllers/job_profiles_skills_controller.rb it pretty much does what you suggest. I prepare both arrays, then the track_events method is called with those 2 arrays summed up. The only difference is around `track_selections(event_key, selected_values, unselected_values)` which you are suggesting, mine doesn't take any params.

Which methods at this controller do you feel is too much? Can you list them? I feel they all have a clear responsibility.

The complication is more around questions_controller as there is the type of question param coming into play. Let's discuss tomorrow.

I would need a better explanation of why you feel the current implementation is wrong. If we can reduce the methods we will - I am happy to do so, but I would like to see which ones as I feel there is no overlapping in the methods responsibilities. Each controller builds its payload differently, and I've moved that method that actually builds the payload to the application_controller level.

app/controllers/job_profiles_skills_controller.rb
```diff
@@ -36,4 +40,30 @@ def skills_params
   def profile_skills_not_editable?
     user_session.job_profile_ids.exclude?(job_profile.id)
   end
+
+  def available_skills
```
cwrw: discussed in person: Im going to move this to job profile in the work Im doing now (after this gets merged) as there is duplication in the job profile controller

PR #352, opened 2019-12-02
GET-728 Add skills gap section to Job profiles

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-728

* Add existing skills by checking intersection with user session
* Add missing skills by diffing user session with job profile skills
* Only show section if there are any skills available

### DIFF
--- a/app/controllers/job_profiles_controller.rb
+++ b/app/controllers/job_profiles_controller.rb
@@ -2,6 +2,8 @@ class JobProfilesController < ApplicationController
   def show
     @job_profile = JobProfileDecorator.new(resource)
     @job_vacancy_count = job_vacancy_count
+    @skills_to_develop = @job_profile.skills - user_skills
+    @existing_skills = @job_profile.skills & user_skills
   end
 
   def target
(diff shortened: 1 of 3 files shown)


app/controllers/job_profiles_controller.rb
```diff
@@ -35,4 +37,8 @@ def job_vacancy_count
   rescue FindAJobService::APIError
     nil
   end
+
+  def user_skill_ids
```
smoothcontract: I think I'd prefer `user_skills` as a name here - this returns `Skill` instances, not id's.

app/controllers/job_profiles_controller.rb
```diff
@@ -2,6 +2,8 @@ class JobProfilesController < ApplicationController
   def show
     @job_profile = JobProfileDecorator.new(resource)
     @job_vacancy_count = job_vacancy_count
+    @skills_to_develop = @job_profile.skills - user_skill_ids
+    @existing_skills = @job_profile.skills & user_skill_ids
```
smoothcontract: So elegant! 🥇

PR #354, opened 2019-12-04
GET-769 job profile content tweaks

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-769

* Add expected/experienced to salary range. This required adding three custom columns for responsiveness as it wouldn't work with normal columns from the govuk styles, which collapsed and overlapped in strange ways. 
* Amend apprenticeships content
* Amend target job section
* Add new link to bottom of page to go back to job matches

### DIFF
--- a/app/decorators/job_profile_decorator.rb
+++ b/app/decorators/job_profile_decorator.rb
@@ -32,23 +32,13 @@ class JobProfileDecorator < SimpleDelegator # rubocop:disable Metrics/ClassLengt
     html_body.xpath(HOW_TO_BECOME_XPATH).text
   end
 
-  def salary_range # rubocop:disable Metrics/MethodLength
+  def salary_range
     return 'Variable' unless salary_min && salary_max
 
-    content_tag(:div, class: 'govuk-grid-row salary-container') {
-      column_data(
-        content: number_to_currency(salary_min, precision: 0) + tag(:br) + '(starter)',
-        content_classes: 'salary-range'
-      ) +
-        column_data(
-          content: ' – ',
-          column_classes: 'salary-separator-container',
-          content_classes: 'salary-separator'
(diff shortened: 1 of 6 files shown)


app/decorators/job_profile_decorator.rb
```diff
@@ -34,7 +35,11 @@ def how_to_become
   def salary_range
     return 'Variable' unless salary_min && salary_max
 
-    "#{number_to_currency(salary_min, precision: 0)} to #{number_to_currency(salary_max, precision: 0)}"
+    content_tag(:div, class: 'salary-columns') {
+      column_data(content: number_to_currency(salary_min, precision: 0) + tag(:br) + '(starter)') +
+        column_data(content: ' – ', content_classes: 'salary-separator') +
+        column_data(content: number_to_currency(salary_max, precision: 0) + tag(:br) + '(experienced)')
+    }.html_safe
```
smoothcontract: Hmm... I'm wondering if this would be easier to understand if it was just an ERB partial rather than using the `content_tag` helper like this. Or perhaps even use ERB directly within this method? This is kind of hard to read. How about something along these lines:

```ruby
ERB.new('
  <div class="salary-columns">
    <div class="salary-column"><%= number_to_currency(salary_min, precision: 0) %><br> (starter)</div>
    <div class="salary-column salary-separator"> - </div>
    <div class="salary-column"><%= number_to_currency(salary_max, precision: 0) %><br> (experienced)</div>
  </div>
').result(binding)
```

I'd prefer a partial TBH, but appreciate using a decorator function may be easier to test. Thoughts?

smoothcontract: Oops, I missed the `<p>` tags above, but you get the general idea :)

PR #356, opened 2019-12-04
GET-722 Update advisor text for xmas/new year

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-722


smoothcontract:
Looks fine but it seems odd to me that the same change isn't replicated on the start page too?

Screenshot 2019-12-04 at 11 08 59

cwrw:
checked with Lois and it's only on the partial

PR #358, opened 2019-12-04
GET-721 Add details to training questions, and amend p size

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-721


tavy88:
Also tested!

PR #362, opened 2019-12-05
Add support for sorting skill matcher results

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-775

Preferred sort order is stored in the user session to avoid having to pass this around as a parameter. Added support for another option to specify order in the skills matcher (we only need this for the main list of skill matches, not when searching for a job profile by name - but this could easily be added later if needed).

### Guidance to review
Add some skills, show match results and experiment with sort order. It's pretty hard to tell the order has changed unless the skills are chosen wisely, but the correct query can be observed in the console.

### DIFF
--- a/app/controllers/skills_matcher_controller.rb
+++ b/app/controllers/skills_matcher_controller.rb
@@ -4,19 +4,20 @@ class SkillsMatcherController < ApplicationController
   def index
     return redirect_to task_list_path unless user_session.job_profile_skills?
 
-    @order = preferred_sort_order
     @results = Kaminari.paginate_array(skills_matcher.match).page(params[:page])
     @scores = skills_matcher.job_profile_scores
     @job_profiles = JobProfileDecorator.decorate(@results)
-    @options = helpers.options_for_select(SORT_OPTIONS, @order)
+    @options = helpers.options_for_select(SORT_OPTIONS, order)
   end
 
   private
 
-  def preferred_sort_order
-    user_session.skills_matcher_sort = sort if %w[skills growth].include?(sort)
+  def order
+    @order ||= begin
+      user_session.skills_matcher_sort = sort if %w[skills growth].include?(sort)
 
-    (user_session.skills_matcher_sort || 'skills').to_sym
+      (user_session.skills_matcher_sort || 'skills').to_sym
+    end
   end
 
   def sort
@@ -24,10 +25,10 @@ class SkillsMatcherController < ApplicationController
   end
 
   def sort_params
-    params.permit(:sort)
+    params.permit(:sort, :page)
   end
 
   def skills_matcher
-    @skills_matcher ||= SkillsMatcher.new(user_session, limit: 50, order: @order)
+    @skills_matcher ||= SkillsMatcher.new(user_session, limit: 50, order: order)
   end
 end
--- a/app/controllers/skills_matcher_controller.rb
+++ b/app/controllers/skills_matcher_controller.rb
@@ -1,15 +1,33 @@
 class SkillsMatcherController < ApplicationController
+  SORT_OPTIONS = [['Skills match', :skills], ['Recent job growth', :growth]].freeze
+
   def index
     return redirect_to task_list_path unless user_session.job_profile_skills?
 
+    @order = preferred_sort_order
     @results = Kaminari.paginate_array(skills_matcher.match).page(params[:page])
     @scores = skills_matcher.job_profile_scores
     @job_profiles = JobProfileDecorator.decorate(@results)
+    @options = helpers.options_for_select(SORT_OPTIONS, @order)
   end
 
   private
 
+  def preferred_sort_order
+    user_session.skills_matcher_sort = sort if %w[skills growth].include?(sort)
+
+    (user_session.skills_matcher_sort || 'skills').to_sym
+  end
+
+  def sort
+    sort_params[:sort]
+  end
+
+  def sort_params
+    params.permit(:sort)
+  end
+
   def skills_matcher
-    @skills_matcher ||= SkillsMatcher.new(user_session, limit: 50)
+    @skills_matcher ||= SkillsMatcher.new(user_session, limit: 50, order: @order)
   end
 end
--- a/app/models/skills_matcher.rb
+++ b/app/models/skills_matcher.rb
@@ -37,8 +37,17 @@ class SkillsMatcher
         )
         .from(Arel.sql("(#{skills_matched_query}) as ranked_job_profiles"))
         .joins('LEFT JOIN job_profiles ON job_profiles.id = ranked_job_profiles.job_profile_id')
-        .order(skills_matched: :desc, growth: :desc, name: :asc)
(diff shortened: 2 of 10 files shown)


app/controllers/skills_matcher_controller.rb
```diff
@@ -1,15 +1,33 @@
...
+    @options = helpers.options_for_select(SORT_OPTIONS, @order)
   end
 
   private
 
+  def preferred_sort_order
+    user_session.skills_matcher_sort = sort if %w[skills growth].include?(sort)
+
+    (user_session.skills_matcher_sort || 'skills').to_sym
+  end
```
tavy88: why not use:

```ruby
def sort
  @sort ||= 
end
```
instead `preferred_sort_order`. This way you don't need `@sort = preferred_sort_order`

app/views/skills_matcher/_sort_form.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class="govuk-form-group hidden" id="sort-container">
+  <%= form_tag nil, method: :get, local: true do |f| %>
```
tavy88: Why is the `nil`needed here?

https://apidock.com/rails/ActionView/Helpers/FormTagHelper/form_tag

spec/models/skills_matcher_spec.rb
```diff
@@ -150,6 +150,33 @@
       )
     end
 
+    it 'arranges job profiles in job growth order as well as matching skills order' do
+      skill1 = create(:skill)
+      skill2 = create(:skill)
+      skill3 = create(:skill)
+      job_profile1 = create(:job_profile, skills: [skill1, skill2, skill3])
+      job_profile2 = create(:job_profile, growth: 60, name: 'Researcher', skills: [skill2])
```
tavy88: would more readable if we used `researcher` instead `job_profile2` and same for the rest.

smoothcontract: I’ve followed the same pattern as the other feature specs here. What’s important is the order of the results so assigning different names to each test job profile wouldn’t really add value.

smoothcontract: I’ve followed the same pattern as the other feature specs here. What’s important is the order of the results so assigning different names to each test job profile wouldn’t really add value.

app/views/skills_matcher/_sort_form.html.erb
smoothcontract: Otherwise this doesn’t generate the correct url path. I want a get request against the current path so this is a nice shortcut for that.

app/controllers/skills_matcher_controller.rb
smoothcontract: That would result in a circular call - sort method returns just the param passed into the controller. It’s a pattern we use extensively.

smoothcontract:
@tavy88 apologies for the double posting and inadvertent resolution of ongoing conversation - I was replying on my mobile and kept clicking on the wrong button coupled with poor connectivity while also trying to eat breakfast :)

spec/models/skills_matcher_spec.rb
tavy88: It would just make the test more readable.

app/controllers/skills_matcher_controller.rb
tavy88: Sorry I meant instead of

```ruby
@order = preferred_sort_order
```

assignment

to use

```ruby
def order
  @order||= begin
      user_session.skills_matcher_sort = sort if %w[skills growth].include?(sort)

     (user_session.skills_matcher_sort || 'skills').to_sym
  end
end
```
What circular call would this create?

I thought I saw `@sort = preferred_sort_order` initially but it's actually `@order`

app/webpacker/packs/sorting.js
```diff
@@ -0,0 +1,24 @@
...
+
+      if (typeof(select) !== 'undefined' && select != null) {
+        handleSortOrder(select);
+      }
+    }
+  }
+
+  function handleSortOrder(element) {
+    element.onchange = function(e) {
+      element.form.submit();
```
cwrw: Although this is simple, I was hoping to use the rails way of handling AJAX (see feedback form) unless there were any caveats?

app/controllers/skills_matcher_controller.rb
```diff
@@ -1,15 +1,33 @@
...
     @results = Kaminari.paginate_array(skills_matcher.match).page(params[:page])
     @scores = skills_matcher.job_profile_scores
     @job_profiles = JobProfileDecorator.decorate(@results)
+    @options = helpers.options_for_select(SORT_OPTIONS, @order)
   end
 
   private
 
+  def preferred_sort_order
+    user_session.skills_matcher_sort = sort if %w[skills growth].include?(sort)
```
cwrw: it would be nice if we memoize this method in place so we won't need `@order`

app/models/skills_matcher.rb
```diff
@@ -37,11 +37,20 @@ def build_query
...
         .joins('LEFT JOIN job_profiles ON job_profiles.id = ranked_job_profiles.job_profile_id')
-        .order(skills_matched: :desc, growth: :desc, name: :asc)
+        .order(order_options)
         .limit(options[:limit])
     end
   end
 
+  def order_options
+    case options[:order]
+    when :growth
```
cwrw: nice and simple 👍

spec/features/skills_matcher_spec.rb
```diff
@@ -120,6 +120,26 @@ def visit_skills_for_current_job_profile(js_enabled = false)
...
+  end
+
+  scenario 'automatically reloads results when sort order changed', :js do
+    visit_skills_for_current_job_profile(true)
+
+    find("option[value='growth']").click
+
+    expect(page).to have_current_path(skills_matcher_index_path(sort: 'growth'))
+  end
+
```
cwrw: could you add a test to show that we persist the selected sorting value, either by pagination and checking or by navigating away and coming back?

app/webpacker/packs/sorting.js
smoothcontract: Discussed offline. Explored this in a separate PR - it introduces more complexity as the displayed URL no longer matches what's shown on the page, so would confuse user if they refreshed. Feels like more work than originally intended for this story and it works well enough with the full page submission, so agreed to keep this without Ajax. PR for future reference: https://github.com/DFE-Digital/get-help-to-retrain/pull/370

spec/features/skills_matcher_spec.rb
smoothcontract: See below 👍

cwrw:
looks good, thanks for addressing! 👍

spec/models/skills_matcher_spec.rb
smoothcontract: If anything, it would make the test less readable as it's only the order that's significant in this specific test. I'm not planning to change this.

PR #363, opened 2019-12-06
Remove ApplicationInsights service

tavy88:
### Description
Now that we have switched to GA tracking
both frontend and backend, we no longer need
Application Insights.

###
https://dfedigital.atlassian.net/browse/GET-774


azure/template.json
```diff
@@ -526,14 +515,6 @@
...
-                "name": "APPINSIGHTS_INSTRUMENTATIONKEY",
-                "value": "[reference('app-insights').outputs.instrumentationKey.value]"
-              },
-              {
-                "name": "APPINSIGHTS_JAVASCRIPT_ENABLED",
-                "value": "[parameters('appInsightsJavaScriptEnabled')]"
-              },
               {
                 "name": "DB_DATABASE",
                 "value": "[parameters('domainDatabaseName')]"
```
cwrw: do we still need the resource at https://github.com/DFE-Digital/get-help-to-retrain/pull/363/files#diff-ac517c061ef2523e9ec4d0cca6624f1eL620 or no? (question for @mafonso)

tavy88: Already covered this with Miguel. He wants it in, still.

PR #364, opened 2019-12-06
Hotfix skills finder

cwrw:
Instead of doing an extra query, which was breaking the controller if no ids were found, build a hash and use mapping to extract skill names. This is used in two places, so moved to the skills model for reuse. 

I also had to read the name in the views rather than the skills model itself for job profiles. This means we don't need the partial anymore, and since it was simple I took it out and kept inline.


app/views/job_profiles/_skills_section.html.erb
```diff
@@ -3,13 +3,17 @@
 <% if @existing_skills.any? %>
   <h3 class="govuk-heading-s">Skills you have</h3>
   <ul class="govuk-list govuk-list--bullet">
-    <%= render @existing_skills %>
+    <% @existing_skills.each do |skill_name| %>
+      <li><%= skill_name %></li>
+    <% end %>
```
tavy88: I still liked the idea of this loop being a partial really.

app/views/job_profiles/_skills_section.html.erb
```diff
@@ -3,13 +3,17 @@
...
   </ul>
 <% end %>
 <% if @skills_to_develop.any? %>
   <h3 class="govuk-heading-s">Skills you may need to develop</h3>
   <ul class="govuk-list govuk-list--bullet">
-    <%= render @skills_to_develop %>
+    <% @skills_to_develop.each do |skill_name| %>
+      <li><%= skill_name %></li>
+    <% end %>
```
tavy88: I still liked the idea of this loop being a partial really.

app/models/skill.rb
```diff
@@ -7,4 +7,16 @@ class Skill < PrimaryActiveRecordBase
   def self.import(names)
     names.map { |name| find_or_create_by(name: name) }
   end
+
+  def self.names_that_include(skill_ids)
```
tavy88: How do you feel about renaming this to `self.inclusive_of(skill_ids)`

app/models/skill.rb
```diff
@@ -7,4 +7,16 @@ class Skill < PrimaryActiveRecordBase
...
+    mapping.slice(*skill_ids).values
+  end
+
+  def self.names_that_exclude(skill_ids)
```
tavy88: And this as `self.exclusive_of(skill_ids)`

spec/models/skill_spec.rb
```diff
@@ -13,4 +13,67 @@
...
+      create(:skill, name: 'Skill2')
+
+      expect(described_class.names_that_include(skill1.id.to_s)).to contain_exactly(
+        skill1.name
+      )
+    end
+
+    it 'returns nothing if no ids intersect' do
+      create(:skill, name: 'Skill1')
+      create(:skill, name: 'Skill2')
```
tavy88: You don't need to create 2 skills in this test. Also, I feel a description like:

`it returns nothing if there is no skill match` would make things clearer. Not sure what ids interest means in this context.

app/models/skill.rb
smoothcontract: I prefer the current name, otherwise it’s less clear that the method returns only id:name hash pairs. The alternate suggestions here make me think skill instances would be returned instead.

smoothcontract: I prefer the current name, otherwise it’s less clear that the method returns only id:name hash pairs. The alternate suggestions here make me think skill instances would be returned instead.

app/views/job_profiles/_skills_section.html.erb
smoothcontract: It would be nice but it needs to be a skill instance for that to work.

spec/models/skill_spec.rb
```diff
@@ -13,4 +13,67 @@
...
+      skill2 = create(:skill, name: 'Skill2')
+
+      expect(described_class.names_that_exclude(skill1.id.to_s)).to contain_exactly(
+        skill2.name
+      )
+    end
+
+    it 'returns all original names if no ids intersect' do
+      skill1 = create(:skill, name: 'Skill1')
+      skill2 = create(:skill, name: 'Skill2')
```
tavy88: You don't need to create 2 skills in this test. Also, I feel a description like:

`it returns all original names if none of the excluded skills is present` would make things clearer. Not sure what ids interest means in this context.

app/views/job_profiles/_skills_section.html.erb
cwrw: hmm going to do something like this, let me amend:
```
<%= render partial: "skills/skill", collection: @existing_skills, as: :skill_name %>
```

app/views/job_profiles/_skills_section.html.erb
tavy88: Surely it can be a partial like: `render: skills, locals: @skills_to_develop` ?

app/models/skill.rb
tavy88: I also believe that with `names_that_include` would expect a list of names, not ids though right? Do you not find this confusing as well?

cwrw: I prefer the current name as well, same below. We are returning a collection of names that include the skill ids, when you include the `skill` before the method name `names_that_include`, I think it reads nicely.

spec/models/skill_spec.rb
cwrw: amended the description but I'd like to keep all skills here to show we return more than one

app/views/job_profiles/_skills_section.html.erb
tavy88: You can also do `render: 'skills', locals: @skills_to_develop` @cwrw

tavy88:
Tested looks good!

PR #366, opened 2019-12-06
Cache all static assets

tavy88:
### Context

We set static cache-control to 'public, max-age=31536000'
so the browser caches all static assets for up to a year.
In production, every asset has a hash added to its name,
so whenever the file changes the browser requests
the latest version as the hash and therefore the whole
filename changes.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -76,6 +76,9 @@ Rails.application.configure do # rubocop:disable Metrics/BlockLength
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
 
+  # caches all static assets for up to a year
+  config.static_cache_control = 'public, max-age=31536000'
+
   # Use a different logger for distributed setups.
   if ENV['RAILS_LOG_TO_STDOUT'].present?
     stdout_logger           = ActiveSupport::Logger.new(STDOUT)


config.ru
```diff
@@ -2,4 +2,5 @@
 
 require_relative 'config/environment'
 
+use Rack::Deflater
```
cwrw: reading up on this, it seems this isn't a great place to add it, and this might be better: https://www.schneems.com/2017/11/08/80-smaller-rails-footprint-with-rack-deflate/
as there is discouragement in the comments: https://stackoverflow.com/questions/7236583/where-to-insert-rackdeflater-in-the-rack also regarding images

smoothcontract: Not convinced by this TBH - it should be the responsibility of Azure WAF to perform asset compression. This would add extra overhead to the Rails app when the WAF cache is cold.

config/environments/production.rb
```diff
@@ -76,6 +76,9 @@
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
 
+  # caches all static assets for up to a year
+  config.static_cache_control = 'public, max-age=31536000'
```
smoothcontract: All good here.

config.ru
tavy88: So this is what I'm looking at with Miguel but WAF does seem to handle compression for a limited amount of files.

tavy88:
Please check my comment.

config.ru
tavy88: None of the answers you provided is upvoted though. It's just different opinions.

smoothcontract: Sounds encouraging 😄

config/application.rb
```diff
@@ -28,6 +28,8 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    config.middleware.use Rack::Deflater
+
```
tavy88: https://thoughtbot.com/blog/content-compression-with-rack-deflater

PR #367, opened 2019-12-08
Update serialize javascript package

cwrw:
Since this is a sub dependency it seems we have to define it as a resolution:
https://yarnpkg.com/lang/en/docs/selective-version-resolutions/
https://til.hashrocket.com/posts/unsfxg7dmw-force-upgrade-of-transitive-dependencies-in-yarn

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,5 +11,8 @@
   },
   "devDependencies": {
     "webpack-dev-server": "^3.9.0"
+  },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


tavy88:
@cwrw Which bit (if any) could be affected by this upgrade on the service?

cwrw:
this seems to be used in webpack libraries, I've walked through the whole application and nothing seems to have been affected

PR #368, opened 2019-12-08
Fix salary range

cwrw:
* Align salary separtor on certain break points to the left
* Change width to 40% to align with responsive section break

Walked through this with Steve fox (2) and he's happy with it

### DIFF
--- a/app/webpacker/styles/_job-profile.scss
+++ b/app/webpacker/styles/_job-profile.scss
@@ -178,16 +178,24 @@
   display: flex;
   flex-flow: row wrap;
 
-  @include govuk-media-query($until: tablet) {
-    width: 30%;
+  @include govuk-media-query($from: mobile, $until: desktop) {
+    width: 40%;
   }
 }
 
 .salary-separator {
   padding: 5px;
+  text-align: center;
+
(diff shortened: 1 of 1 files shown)


app/webpacker/styles/_job-profile.scss
```diff
@@ -178,16 +178,24 @@
...
   padding: 5px;
+  text-align: center;
+
+  @include govuk-media-query($from: tablet, $until: 965px) {
+    text-align: left;
+  }
+
+  @include govuk-media-query($from: mobile, $until: 480px) {
+    text-align: left;
+  }
```
tavy88: Hmm is this within the GDS guidelines? I think we should only use the 3 breakpoints they provide? Otherwise, this could become messy really quick if we start adding custom breakpoints.

cwrw: Although we usually stick to that and I usually think we should adhere to the columns the salary range is a special edge case where we couldn't use the columns from the Gov-uk frontend library, and need to add some extra points to align the text to the left for one column. So I think it is ok in this case alone

tavy88: That's how it always starts: Just this time! Ha ha. Ok, will approve.

PR #372, opened 2019-12-10
Remove Loggly gem and configuration

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-836

We no longer need to use Loggly as all logging is now handled through Datadog.

### Guidance to review
No more Loggly!


smoothcontract:
Deployed to dev1 and ready for QA - key thing is to check that the Datadog logging is unaffected.

PR #376, opened 2019-12-12
Add semantic logger formatter for data dog

cwrw:
this is taken from:
https://github.com/flipgroup/data-dog-semantic-logger/

The amendments:
1) log does not have method `method_name`, I adapted it to the data dog
recommendations (added custom name for logger name and the log.nam was
actually the class name)
2) Rubocop fixes
3) Remove try so that we can see the errors for anything missing

### DIFF
--- a/app/lib/semantic_logger/formatters/data_dog.rb
+++ b/app/lib/semantic_logger/formatters/data_dog.rb
@@ -1,7 +1,3 @@
-require 'json'
-require 'useragent'
-require 'uri'
-
 # This can be attributed to https://github.com/flipgroup/data-dog-semantic-logger/
 # Log message structure:
 # https://docs.datadoghq.com/logs/processing/attributes_naming_convention/
@@ -31,17 +27,6 @@ require 'uri'
 #       path: <string>,
 #       queryString: <object>,
 #       scheme: <string>
-#     },
-#     useragent_details: {
-#       os: {
-#         family: <string>
-#       },
-#       browser: {
-#         family: <string>
-#       },
-#       device: {
-#         family: <string>
-#       }
 #     }
 #   },
 #   logger: {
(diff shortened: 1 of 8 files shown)


app/lib/semantic_logger/formatters/data_dog.rb
```diff
@@ -0,0 +1,298 @@
...
+      def time?
+        !time.nil?
+      end
+
+      def appname
+        ENV['APP_NAME'].to_s
+      end
+
+      def appname?
+        !appname.nil?
```
tavy88: Can this not be `appname.present?` instead?

cwrw: not nil is not synonymous to present, example:
```
irb(main):002:0> !{}.nil?
=> true
irb(main):003:0> {}.present?
=> false
```

tavy88: Surely your `appname` can be NIL or have value though. Why would it have a hash value?

```ruby
irb(main):001:0> !nil.nil?
=> false
irb(main):002:0> nil.present?
=> false
```

app/lib/semantic_logger/formatters/data_dog.rb
```diff
@@ -0,0 +1,299 @@
...
+
+        (local_time.to_time.utc.to_f * 1000).to_i
+      end
+
+      def time?
+        !time.nil?
+      end
+
+      def appname
+        ENV['APP_NAME'].to_s
```
tavy88: This can be `Rails.application.class.parent_name`. No need to pass it in as env var.

tavy88: You don't need to pass the app name as ENV:

You can use `Rails.application.class.parent_name`.

PR #378, opened 2019-12-12
Add support for viewing and editing user personal data via active admin

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-815

Needed to tweak validation for `UserPersonalData` slightly to allow ActiveAdmin to edit the `dob` field directly. Other than that this just adds a new Admin class and includes a link on the dashboard. 

### Guidance to review
Worked with Lois to iron out the columns, filters etc. for this and is happy. Will need to add support for the `ADMIN_MODE` flag to get this deployed on a review app, so will cover that in the next story. For now this is still only exposed in development mode so will need to be tested locally.

### DIFF
--- a/app/admin/user_personal_data.rb
+++ b/app/admin/user_personal_data.rb
@@ -1,0 +1,33 @@
+if defined?(ActiveAdmin)
+  ActiveAdmin.register UserPersonalData do
+    GENDER_OPTIONS = {
+      'Male' => :male,
+      'Female' => :female,
+      'Prefer not to say' => :not_mentioned
+    }.freeze
+
+    actions :all, except: :new
+
+    permit_params :first_name, :last_name, :postcode, :dob, :gender
+
+    filter :first_name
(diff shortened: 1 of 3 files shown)


app/admin/user_personal_data.rb
```diff
@@ -0,0 +1,33 @@
...
+        f.input :postcode
+        f.input :dob
+        f.input :gender, as: :select, collection: GENDER_OPTIONS
+      end
+      f.actions
+    end
+
+    config.sort_order = 'id_asc'
+  end
+end
```
tavy88: I would like to see some tests - at least that check the page is accessed without errors. Thoughts?

We are now moving this to production and it's no longer a dev only tool.

smoothcontract: That will come in follow up PR's. This needs to be merged to unblock subsequent work. At this point it still is a dev tool and only exposed for local development mode.

PR #379, opened 2019-12-12
Add support for `ADMIN_MODE` environment variable

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-804

If this variable is set then only the ActiveAdmin routes are exposed (not yet authenticated). If not set then only the user facing routes are exposed. I left the Feature flag UI out of this for now as we might want to use that in either mode, and it's only available for local development.

With admin mode on, site root directs to the admin pages and no customer facing routes are available. With the admin mode flag off the site behaves in customer facing mode only, no admin routes are available.

Routes are not evaluated dynamically so the proposed `if / then / else` approach didn't work. Adding custom constraint classes proved to be a nice solution though. The diff is a little messy but essentially this just adds a `constraints` block around the two sets of routes.

### Guidance to review
Note that this does not yet introduce authentication. To test locally, set an environment variable (perhaps in `.env.development.local`):

```
ADMIN_MODE=true
```

One the heroku review app I've enabled admin mode and also setup some sample test data for feedback surveys and user personal data.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -206,3 +206,9 @@ en-GB:
               blank: Enter a postcode
             dob:
               invalid: Enter a valid date of birth
+
+  active_admin:
+    filters:
+      predicates:
+        from: "From"
+        to: "To"
--- a/spec/routing/admin_spec.rb
+++ b/spec/routing/admin_spec.rb
@@ -1,0 +1,94 @@
+require 'rails_helper'
+
+RSpec.describe 'routes for admin pages', type: :routing do
+  context 'when admin mode is enabled' do
+    around do |example|
+      Rails.configuration.admin_mode = true
+      example.run
+    ensure
+      Rails.configuration.admin_mode = false
+    end
+
+    it 'successfully routes site root to admin/dashboard#index' do
+      expect(get('/')).to route_to(controller: 'admin/dashboard', action: 'index')
+    end
+
+    it 'successfully routes /admin path to admin/dashboard#index' do
+      expect(get('/admin')).to route_to(controller: 'admin/dashboard', action: 'index')
+    end
+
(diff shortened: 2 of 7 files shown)


spec/routing/admin_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'rails_helper'
+
+RSpec.describe 'routes for admin pages', type: :routing do
+  context 'when admin mode is enabled' do
+    around do |example|
+      Rails.configuration.admin_mode = true
+      example.run
+    ensure
+      Rails.configuration.admin_mode = false
+    end
```
tavy88: I don't think we change the config for tests like this.

Why not:

```ruby
allow(Rails.application.config).to receive(: admin_mode).and_return(true)
```

and 

```ruby
allow(Rails.application.config).to receive(: admin_mode).and_return(false)
```

in before blocks for respective sections?

config/locales/en.yml
```diff
@@ -206,3 +206,9 @@ en-GB:
               blank: Enter a postcode
             dob:
               invalid: Enter a valid date of birth
+
+  active_admin:
```
cwrw: 👍

spec/routing/admin_spec.rb
cwrw: hmm interesting, I've seen both implementation so Im fine with either.

smoothcontract:
> looks good, tested and all works as expected, and thanks for fixing the previous bug. I was wondering about the cookies, and if we could include the change here as well?

I think it makes sense to cover session changes as part of the authentication story.

spec/routing/admin_spec.rb
smoothcontract: Yes, either approach works fine so no real benefit in changing this.

PR #381, opened 2019-12-13
Add a new rake task to support setting up improved test data for admin area

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-815

This creates random feedback surveys and user personal data. Also needed to upgrade Faker to avoid some deprecation warnings. Plan is to run this rake task against Heroku parent review app so that we have better test data to use for working on the admin stories.

### Guidance to review
Probably best to run locally for testing.


spec/factories/user_personal_data.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :user_personal_data do
+    first_name { Faker::Name.first_name }
+    last_name { Faker::Name.last_name }
+    postcode { ['NW11 8QE', 'NW11 7PT', 'NW11 7HB', 'NW10 7SF', 'NW10 7NZ', 'NW10 6HJ'].sample }
+    gender { %w[male female not_mentioned].sample }
+    dob { Faker::Date.between(from: 99.years.ago, to: 18.years.ago) }
```
tavy88: I think these should be traits rather than randomised. This could create flakiness in further tests that use this factory.

E.g `gender`, `postcode`

smoothcontract: The factory isn't used anywhere other than this new rake task at present, so that's not a problem we have. I suppose they could all be moved under a `:sample` trait if that feels cleaner. @cwrw thoughts?

tavy88: But what's the point of randomizing (when we're not using that data anyway). I also assume this will be used when we write tests for the admin/user_personal_data section.

spec/factories/feedback_survey.rb
```diff
@@ -2,5 +2,9 @@
   factory :feedback_survey do
     page_useful { true }
     url { 'https://some-url.com' }
+
+    trait :with_message do
+      message { Faker::Movies::HarryPotter.quote }
```
cwrw: bahaha excellent!
`"We could all have been killed - or worse, expelled."` 👍

spec/factories/user_personal_data.rb
cwrw: A factory has to be valid at its most basic setup, after that we can add specific traits if reuse is needed, so I think this implementation is ok. I think this won't be an issue even in testing as you can override whichever fields you want really

tavy88: We are generating more data than we need at this point.

Why do we need  a sample of this`['NW11 8QE', 'NW11 7PT', 'NW11 7HB', 'NW10 7SF', 'NW10 7NZ', 'NW10 6HJ']` ?

smoothcontract: I copied the same approach for postcodes we use in other factories. And I'm not convinced performance is much of a concern here. So I'm going to go ahead and merge this as is.

PR #382, opened 2019-12-13
Make Faker and FactoryBot gems available in production mode

smoothcontract:
This allows rake tasks to setup admin test data on Heroku review apps. I could alternatively work around this by attaching to the heroku DB instead...


smoothcontract:
Abandoning this idea - found a workaround instead. No need to increase the production bundle size for a one-off data setup.

PR #383, opened 2019-12-13
Customize event category for outbound links

tavy88:
When users click on different course providers,
we want to track which topic these course
providers fall under.

Screen Shot 2019-12-13 at 15 05 20

https://dfedigital.atlassian.net/browse/GET-846

### DIFF
--- a/app/views/courses/_course.html.erb
+++ b/app/views/courses/_course.html.erb
@@ -4,6 +4,6 @@
   <p class="govuk-body govuk-!-margin-bottom-0">Distance: <b><%= course.distance.round(1) %> miles</b></p>
   <p class="govuk-body-s"><%= course.full_address %></p>
   <p class="govuk-body-s">Tel:<%= link_to(course.phone_number, "tel:#{course.phone_number}", class: 'govuk-!-padding-left-1 govuk-link' ) %></p>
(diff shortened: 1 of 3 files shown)


app/views/courses/_course.html.erb
```diff
@@ -4,6 +4,6 @@
   <p class="govuk-body govuk-!-margin-bottom-0">Distance: <b><%= course.distance.round(1) %> miles</b></p>
   <p class="govuk-body-s"><%= course.full_address %></p>
   <p class="govuk-body-s">Tel:<%= link_to(course.phone_number, "tel:#{course.phone_number}", class: 'govuk-!-padding-left-1 govuk-link' ) %></p>
-  <%= link_to 'Visit website', course.url, class: 'govuk-!-margin-bottom-6 govuk-button', aria: { label: "Visit website of #{course.provider}" }, data: { module: 'govuk-button' } %>
+  <%= link_to 'Visit website', course.url, class: 'govuk-!-margin-bottom-6 govuk-button', aria: { label: "Visit website of #{course.provider}" }, data: { module: 'govuk-button', ga_event_category: course.topic } %>
```
cwrw: talked in person but for posterity: we are currently only sending the course name to tracking. It would be good to have a notion of that this is a type of course. I think this is worth looking into as we will eventually be adding more courses, and they would fall into this category.

tavy88:
Noelyn is happy with the current convention.

PR #385, opened 2019-12-16
Generate ActiveAdmin assets for deployment

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-856

This is a temporary measure until ActiveAdmin adds support for Webpacker. There is an ongoing PR to add Webpacker support to ActiveAdmin in a future release, but we can't depend on this being released soon. In the meantime, adding `assets:precompile` step uses the sprockets based pipeline to generate the ActiveAdmin related assets only.

### Guidance to review
Deployed to Dev1, with `ADMIN_MODE` flag on, and it's all working nicely.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,8 @@ RUN bundle install --clean --force --without "development test" \
 COPY package.json yarn.lock ./
 RUN yarn install
 COPY . ./
-RUN bundle exec rails webpacker:compile SECRET_KEY_BASE=stubbed
+# assets:precompile is necessary only for ActiveAdmin assets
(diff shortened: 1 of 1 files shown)


smoothcontract:
FYI have a suspicion that this is actually building the webpacker packers twice, so trying just the second command on a different branch. Will deploy this to Azure and see if the user facing site renders correctly: https://github.com/DFE-Digital/get-help-to-retrain/pull/386

smoothcontract:
Replaced by https://github.com/DFE-Digital/get-help-to-retrain/pull/386

PR #387, opened 2019-12-17
Improve event category outbound links

tavy88:
For courses, use a better event category name when
tracking the outbound links.

### DIFF
--- a/app/views/courses/_course.html.erb
+++ b/app/views/courses/_course.html.erb
@@ -4,6 +4,6 @@
   <p class="govuk-body govuk-!-margin-bottom-0">Distance: <b><%= course.distance.round(1) %> miles</b></p>
   <p class="govuk-body-s"><%= course.full_address %></p>
   <p class="govuk-body-s">Tel:<%= link_to(course.phone_number, "tel:#{course.phone_number}", class: 'govuk-!-padding-left-1 govuk-link' ) %></p>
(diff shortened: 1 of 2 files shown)


app/views/courses/_course.html.erb
```diff
@@ -4,6 +4,6 @@
   <p class="govuk-body govuk-!-margin-bottom-0">Distance: <b><%= course.distance.round(1) %> miles</b></p>
   <p class="govuk-body-s"><%= course.full_address %></p>
   <p class="govuk-body-s">Tel:<%= link_to(course.phone_number, "tel:#{course.phone_number}", class: 'govuk-!-padding-left-1 govuk-link' ) %></p>
-  <%= link_to 'Visit website', course.url, class: 'govuk-!-margin-bottom-6 govuk-button', aria: { label: "Visit website of #{course.provider}" }, data: { module: 'govuk-button', ga_event_category: course.topic } %>
+  <%= link_to 'Visit website', course.url, class: 'govuk-!-margin-bottom-6 govuk-button', aria: { label: "Visit website of #{course.provider}" }, data: { module: 'govuk-button', ga_event_category: "#{course.topic}_courses" } %>
```
cwrw: I would prefer to prepend the courses to the name, as it would be easier to navigate in the events in GA, and makes no difference for the analyst whilst populating their reports

tavy88: You say `I am checking for Maths Courses` not `I am checking Courses Maths`. I think we are wasting too much time on such an unimportant thing really.

cwrw: spoke in person, will leave as is for now

PR #388, opened 2019-12-17
Azure AD Authentication for Active Admin

tavy88:
### Context

The aim here is to separate the ActiveAdmin specific logic as much as possible from the rest. This PR implements the following:

- Add a sign-in page
- Configure active admin following resources:
  - authentication_method which acts as a before filter in
    every Active Admin controllers
  - current_user_method which creates the current_user concept
    just like devise
  - logout path which is a GET
- Implement the logout process:
  - reset the session
  - call Microsoft's logout API endpoint to ensure logout from
    their side as well, not just from ours as recommended: https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-web-app-sign-user-sign-in?tabs=aspnetcore#sign-out
- Rename the Omniauth strategy
- Move all logic regarding active admin under the admin namespace
- Move admin user session handling to AdminUserSession
- Switch to `cookie_storage` from `active_record_storage` when Admin Mode is ON.
- Store the user roles on the session when we can later on user them

#### Before reviewing

1. Please make sure you have an understanding of how Rack-based Omniauth works, what is an Authentication Strategy and how that is implemented on top of OAuth2 authentication protocol. Here's a link that does similar implementations for Twitter/Facebook: https://www.sitepoint.com/rails-authentication-oauth-2-0-omniauth/

2. Please make sure you understand the following phases that Omniauth uses to follow the Oauth2 protocol:
    - request_phase
    - callback_phase

3. Please make sure you understand how ActiveAdmin is configured to support a filter that checks if the user is authenticated and provides a current_user if so. Starting point: https://stackoverflow.com/questions/43637400/activeadmin-setting-config-logout-link-path-cause-undefined-method-error

4. Please check Microsoft's sample for AzureAD integration: https://docs.microsoft.com/en-us/graph/tutorials/ruby?tutorial-step=3

### Screenshots

Screen Shot 2019-12-20 at 19 33 30

### Ticket

https://dfedigital.atlassian.net/browse/GET-806


app/controllers/admin/user_sessions_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+      reset_session
+
+      sign_out_and_redirect_to(root_url)
+    end
+
+    private
+
+    def sign_out_and_redirect_to(path)
+      redirect_to "#{MICROSOFT_LOGOUT_PATH}?post_logout_redirect_uri=#{path}"
+    end
```
tavy88: TBD with Lois and see if we want to logout from Microsoft as well for better security to ensure the JWT is invalidated as Microsoft recommends (please check the PR description)

Gemfile
```diff
@@ -9,6 +9,10 @@ ruby File.read('.ruby-version').chomp
 gem 'rails', '~> 6.0.0'
 gem 'rails-i18n', '~> 6.0.0'
 
+# Azure AD integration with Omniauth and OAuth2
+gem 'omniauth-oauth2'
```
smoothcontract: This is cool I think. It's generally a good idea to use/re-use battle tested security code if possible so omniauth is a great choice from that point of view. It is a big gem though which introduces a lot of other dependencies. On balance I think the tradeoff is worthwhile, but we might need to keep an eye on infrastructure overhead (particularly memory) when this is deployed.

app/controllers/admin/auth_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+        redirect_to root_path
+      else
+        redirect_to admin_sign_in_path, alert: t('active_admin.errors.authentication')
+      end
+    end
+
+    private
+
+    def admin_user_roles
```
smoothcontract: I feel like this stuff doesn't really belong in a controller. One alternative would be to pass the omniauth hash directly into the `AdminUserSession` class and have that assign the individual attributes. That too doesn't feel like a spectacular fit, so perhaps the best place for this is the `AdminUser` model. Perhaps add a `roles` attribute to the `AdminUser` model that is built within the `from_omniauth` method. It need not be persisted though, just made available as an instance method.

app/lib/omni_auth/strategies/azure_ad_auth.rb
```diff
@@ -0,0 +1,61 @@
...
+      option :authorize_options, [:scope]
+
+      # Rescue any issue occuring during the request phase - TBD for when we have an error page available
+      # def call(env)
+      #   super
+      # rescue StandardError => e
+      #   Rails.logger.error("Admin Authentication Error: #{e.message}")
+
+      #   redirect '/admin/sign-in-error'
+      # end
```
smoothcontract: I think you can remove this commented section now - as discussed offline, the default error handling with a 500 page is sufficient to handle this.

app/models/admin_user.rb
```diff
@@ -0,0 +1,15 @@
+class AdminUser < RestrictedActiveRecordBase
+  validates :email, presence: true, uniqueness: { case_sensitive: false }
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP, allow_blank: true }
+  validates :name, presence: true
+
+  def self.from_omniauth(auth_hash)
+    user_info = auth_hash['extra']['user_info']
```
smoothcontract: As mentioned above, I think this would be a nice spot to parse out the list of user roles and make those available as an instance method array in the admin user. Maybe this is a good opportunity to map the Azure role GUIDs to something more friendly we will use internally (but that could wait until the authorisation story).

app/models/admin_user_session.rb
```diff
@@ -0,0 +1,23 @@
+class AdminUserSession
+  attr_reader :session
+
+  def initialize(session)
+    @session = session
+  end
+
+  def admin_user_id
```
smoothcontract: I feel that the `admin_user` prefix is superfluous as it's scoped within an `AdminUser` class. So perhaps consider renaming this to just `id` or `user_id`.

app/models/admin_user_session.rb
```diff
@@ -0,0 +1,23 @@
...
+    session[:admin_user_id]
+  end
+
+  def admin_user_id=(value)
+    session[:admin_user_id] = value
+  end
+
+  def admin_user_roles
```
smoothcontract: Similarly maybe rename to `roles` or `user_roles`. It's still then quite clear that this relates to an admin user.

db/restricted_migrate/20191218121428_create_admin_users.rb
```diff
@@ -0,0 +1,12 @@
+class CreateAdminUsers < ActiveRecord::Migration[6.0]
+  def change
+    enable_extension 'citext'
```
smoothcontract: The extension is already enabled, so this line is redundant. See `CreateUsers` migration.

app/models/admin_user.rb
tavy88: Oh yeah I only care about storing the roles in this story.

app/models/admin_user_session.rb
tavy88: Yep

app/models/admin_user_session.rb
tavy88: Yep.

Gemfile
tavy88: Agreed

app/controllers/admin/auth_controller.rb
tavy88: Yeah, it's a bit tricky. I did think about it.

app/views/layouts/admin/application.html.erb
```diff
@@ -0,0 +1,62 @@
...
+            </span>
+          </div>
+          <div class="govuk-footer__meta-item">
+            <a class="govuk-footer__link govuk-footer__copyright-logo" href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/">© Crown copyright</a>
+          </div>
+        </div>
+      </div>
+    </footer>
+  </body>
+</html>
```
tavy88: Will check with Lois to see if she's happy with this solution for now.

app/models/admin_user.rb
```diff
@@ -0,0 +1,26 @@
+class AdminUser < RestrictedActiveRecordBase
+  validates :email, presence: true, uniqueness: { case_sensitive: false }
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP, allow_blank: true }
+  validates :name, presence: true
```
tavy88: We need to add the `resource_id` here as well.

config/initializers/session_store.rb
```diff
@@ -1,2 +1,4 @@
-Rails.application.config.session_store :active_record_store, key: '_get_help_to_retrain_session'
-ActionDispatch::Session::ActiveRecordStore.session_class = Session
+unless Rails.configuration.admin_mode
+  Rails.application.config.session_store :active_record_store, key: '_get_help_to_retrain_session'
+  ActionDispatch::Session::ActiveRecordStore.session_class = Session
+end
```
tavy88: Testing in RSpec this was a bit tricky.

spec/features/admin/authentication_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+      expect(page).to have_content(content)
+    end
+  end
+
+  scenario 'Stores the user details on successful authentication' do
+    stub_omniauth
+
+    visit(auth_azure_ad_auth_callback_path)
+
+    expect(AdminUser.last.email).to eq('test@test.com')
```
tavy88: Maybe I want to do something like:

```ruby
 expect(described_class.last).to have_attributes(
        name: 'some.name',
        resource_id: '1111-111-11-1',
        email: 'test@test.com'
      )
```

app/views/layouts/admin/application.html.erb
tavy88: She is happy with this solution for now. She might ask Sophie about the copy.

app.json
```diff
@@ -36,6 +36,18 @@
     },
     "BING_SPELL_CHECK_API_KEY": {
       "required": false
+    },
+    "AZURE_CLIENT_ID": {
```
cwrw: Since the main Heroku app does not have those values, we do not need to specify azure specific ENV variables

app/lib/omni_auth/strategies/azure_ad_auth.rb
```diff
@@ -0,0 +1,52 @@
...
+             site: SITE,
+             authorize_url: "/#{Rails.application.config.azure_tenant_id}/oauth2/v2.0/authorize",
+             token_url: "/#{Rails.application.config.azure_tenant_id}/oauth2/v2.0/token"
+
+      # Send the scope parameter during authorize
+      option :authorize_options, [:scope]
+
+      # Unique ID for the user is the id field
+      uid { user_info['id'] }
+
```
cwrw: discussed briefly in person, but for posterity, the omniauth strategy guide contains some useful information and schema guides which I think would benefit us, see https://github.com/omniauth/omniauth/wiki/Strategy-Contribution-Guide#defining-the-callback-phase. An example would be to use info as per the [Auth Hash Schema](https://github.com/omniauth/omniauth/wiki/Auth-Hash-Schema)

config/initializers/omniauth_azure_ad.rb
```diff
@@ -0,0 +1,8 @@
+require 'omni_auth/strategies/azure_ad_auth'
+
+Rails.application.config.middleware.use OmniAuth::Builder do
+  provider :azure_ad_auth,
+           Rails.application.config.azure_client_id,
+           Rails.application.config.azure_client_secret,
+           scope: Rails.application.config.azure_scopes
```
cwrw: **Not a blocker** - spoke in person but for posterity I prefer if we keep our key vault secrets a store for sensitive data example API keys or data that tends to be different across different ENVs. As the scopes are unlikely to change or be different across ENVs, I would prefer we set it here directly

config/routes.rb
```diff
@@ -6,6 +6,9 @@
     ActiveAdmin.routes(self)
 
     get '/', to: 'admin/dashboard#index'
+    get 'admin/sign-in', to: 'pages#admin_sign_in'
```
cwrw: I feel that both sign in and sign out can be in the auth controller, as I think its the right place to encapsulate the logic, bonus is that we can even define the layout at the controller level itself as well `layout 'something'`, as we've done a good job so far separating the admin from the rest of the app logic

app/helpers/admin/authentication_helper.rb
```diff
@@ -0,0 +1,22 @@
+module Admin
+  module AuthenticationHelper
+    # Override active admin helper method that ensures user is authenticated before accessing the dashboard
+    def authenticate_active_admin_user!
+      redirect_to admin_sign_in_path unless admin_current_user.present?
+    end
+
+    # Override active admin helper method that ensures we have access to a current_user
+    def admin_current_user
+      if admin_user_session.user_id.present?
```
cwrw: we could return early here instead of setting the admin current user to nil

app.json
tavy88: Yeah, I would suggest we add them to the main app as this is an important piece which will be live. It's no longer something optional.

app/lib/omni_auth/strategies/azure_ad_auth.rb
```diff
@@ -0,0 +1,52 @@
...
+      # Get additional information after token is retrieved
+      extra do
```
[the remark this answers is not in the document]
tavy88: Hmm, let's discuss. I would not want to access the keys here as we move the error at rack level.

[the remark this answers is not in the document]
tavy88: Also `info` does not need to nest `resource_id` as that is covered by `uid` which is required on the schema.

config/routes.rb
tavy88: sing-in can move, but sign-out & sign-in are (create, destroy) actions under sessions_controller. auth_controller is only for Omniauth auth hash handling - it's a standard pattern.

app/models/admin_user.rb
```diff
@@ -0,0 +1,27 @@
+class AdminUser < RestrictedActiveRecordBase
+  validates :email, presence: true
+  validates :email, format: { with: URI::MailTo::EMAIL_REGEXP, allow_blank: true }
```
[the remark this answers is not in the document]
tavy88: nope

PR #392, opened 2019-12-20
Get 746 skills rarity

smoothcontract:
### Context
https://dfedigital.atlassian.net/browse/GET-746

Adds a new `rarity` attribute to `Skill` model. This is precalculated via a new rake task and represents the frequency of each skill by ranking them. The lowest rank of `1` represents unique skills, higher numbers represent more commonplace skills.

When job profiles have identical growth and number of skills matched, then the skills rarity is used to further prioritise matches. This requires updating the skills matcher to also join against skills table to determine appropriate rarity for each skill. However, In order to avoid returning duplicate job profile results the query returns only the lowest rarity of all skills matched to a job profile (i.e. the rarest).

Also added new rarity attribute to admin interface and added appropriate test coverage to demonstrate ordering works as expected.

### Guidance to review
* Add the database migrations
* Run the new rake task
* Try to test with a mixture of rare and common skills (fun!)

### DIFF
--- a/app/models/skills_matcher.rb
+++ b/app/models/skills_matcher.rb
@@ -23,7 +23,8 @@ class SkillsMatcher
 
   def skills_matched_query
     JobProfileSkill
-      .select(:job_profile_id, 'COUNT(job_profile_id) AS skills_matched')
+      .select(:job_profile_id, 'MIN(rarity) as skills_rarity', 'COUNT(job_profile_id) AS skills_matched')
+      .joins('LEFT JOIN skills ON skills.id = job_profile_skills.skill_id')
       .where(skill_id: user_session.skill_ids)
       .group(:job_profile_id)
       .to_sql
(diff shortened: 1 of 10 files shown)


app/models/skills_matcher.rb
```diff
@@ -23,7 +23,8 @@ def job_profile_scores
 
   def skills_matched_query
     JobProfileSkill
-      .select(:job_profile_id, 'COUNT(job_profile_id) AS skills_matched')
+      .select(:job_profile_id, 'MIN(rarity) as skills_rarity', 'COUNT(job_profile_id) AS skills_matched')
+      .joins('LEFT JOIN skills ON skills.id = job_profile_skills.skill_id')
```
tavy88: Hmm I wonder here if we should use eager loading instead of JOINs: https://axiomq.com/blog/better-performance-for-rails-app-with-joins-and-eager-loading/

Would be good to measure the performance impact.

cwrw: discussed in person, not an issue

PR #394, opened 2020-01-02
GET-783 Spell check service does not fail if no API key available

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-783

* Spell check service does not throw error when there is no API key
* Spell check service returns early if API key is not present
* Remove API key from test ENV, and only set in respective tests to stop making real calls in specs (which was happening previously)
* Fix tests in feature to avoid making real API calls when key set, also fix an old test as it will always pass otherwise

How to test:
tail the logs on heroku `heroku logs -t --app get-help-to-get-783-dis-ir3cm5` and perform a search, you should not see `Spell Check service error` in the logs

I will remove the bing keys from the vaults as we move along the pipeline


spec/features/check_your_skills_spec.rb
```diff
@@ -67,6 +67,7 @@ def fake_bing_api_call_with(response_body)
   end
 
   scenario 'User enters an incorrect word and a correction is returned' do
+    Rails.configuration.bing_spell_check_api_key = 'test'
```
tavy88: As discussed this can also be

```ruby
allow(Rails.configuration).to receive(:bing_spell_check_api_key).and_return('test')
```

This would eliminate the need to call 

```ruby
 ensure
    Rails.configuration.bing_spell_check_api_key = nil
```

However, this is a personal preference so it's up to you if you want to try my suggestion.

spec/features/check_your_skills_spec.rb
```diff
@@ -11,15 +11,15 @@
...
+  def fake_bing_api_call_with(response_body, text)
     request_headers = {
       'Content-Type' => 'application/x-www-form-urlencoded',
       'Ocp-Apim-Subscription-Key' => 'test'
     }
 
     stub_request(:get, SpellCheckService::API_ENDPOINT)
       .with(headers: request_headers,
-            query: URI.encode_www_form(mkt: 'en-gb', mode: 'spell', text: 'Gatas'))
+            query: URI.encode_www_form(mkt: 'en-gb', mode: 'spell', text: text))
```
tavy88: Cool for parametrizing.

However, because it's a test and we want to be specific usually in a test, not sure why we would need to use different values. But I get the flexibility for maybe future scenarios. So all good!

PR #396, opened 2020-01-02
SPIKE: GET-866-GA-location-override 

tavy88:
### Context

Finding a way to override the location in GA and use the client's actual IP + User-agent.


cwrw:
tackled in https://github.com/DFE-Digital/get-help-to-retrain/pull/404

PR #398, opened 2020-01-06
Get 805

mafonso:
### Context
https://dfedigital.atlassian.net/browse/GET-805

This PR introduces a conditional webapp resource with a matching conditional output variable.
This output variable will be used by the Azure DevOps pipeline in order to determine if the start/swap/process should run for that application.

### DIFF
--- a/azure-pipelines-deploy-template.yml
+++ b/azure-pipelines-deploy-template.yml
@@ -40,6 +40,7 @@ jobs:
 
             - task: AzureAppServiceManage@0
               displayName: 'Start Azure App Service: ${{parameters.environment}}'
+              condition: ne(variables['appServiceName'], 'disabled')
               inputs:
                 azureSubscription: '${{parameters.subscriptionServiceConnection}}'
                 Action: 'Start Azure App Service'
@@ -50,6 +51,7 @@ jobs:
 
             - task: AzureAppServiceManage@0
               displayName: 'Swap Slots: ${{parameters.environment}}'
+              condition: ne(variables['appServiceName'], 'disabled')
               inputs:
                 azureSubscription: '${{parameters.subscriptionServiceConnection}}'
                 WebAppName: '${{parameters.resourceGroupName}}-as'
@@ -58,6 +60,7 @@ jobs:
 
             - task: AzureAppServiceManage@0
               displayName: 'Stop Azure App Service: ${{parameters.environment}}'
+              condition: ne(variables['appServiceName'], 'disabled')
               inputs:
                 azureSubscription: '${{parameters.subscriptionServiceConnection}}'
                 Action: 'Stop Azure App Service'
--- a/azure/config/dev2.parameters.json
+++ b/azure/config/dev2.parameters.json
@@ -2,6 +2,9 @@
   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
+    "sku": {
+      "value": "P1V2"
+    },
     "domainDatabaseName": {
       "value": "domain"
     },
--- a/azure/config/dev2.parameters.json
+++ b/azure/config/dev2.parameters.json
@@ -105,6 +105,9 @@
     },
     "googleAnalyticsTrackingId": {
       "value": ""
+    },
+    "adminEnabled": {
+      "value": true
     }
   }
 }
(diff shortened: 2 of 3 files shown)


azure/config/dev2.parameters.json
```diff
@@ -2,6 +2,9 @@
   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
+    "sku": {
+      "value": "P1V2"
```
mafonso: We need more resources on this environment. Dev2 was still on a S1

cwrw: UT is the only environment left without this, should we add it there as well or is it not needed (yet)?

azure/template.json
```diff
@@ -160,12 +160,14 @@
...
     "environmentName": "[split(variables('resourceNamePrefix'),'-')[1]]",
     "appServiceName": "[concat(variables('resourceNamePrefix'), '-as')]",
+    "adminServiceName": "[concat(variables('resourceNamePrefix'), 'admin-as')]",
     "keyvaultCertificateName": "[if(greater(length(parameters('certificateName')),0), parameters('certificateName'), replace(parameters('customHostName'), '.', '-'))]",
     "storageAccountName": "[replace(concat(variables('resourceNamePrefix'), 'str'), '-', '')]",
     "appServicePlanName": "[concat(variables('resourceNamePrefix'), '-asp')]",
     "domainDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-', parameters('domainDatabaseName'),'Psql')]",
     "restrictedDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-', parameters('restrictedDatabaseName'),'Psql')]",
     "appServiceRuntimeStack": "[concat('COMPOSE|', base64(replace(base64ToString(parameters('dockerComposeFile')),'__CONTAINER_REFERENCE__',parameters('containerImageReference'))))]",
+    "adminServiceRuntimeStack": "[concat('DOCKER|', parameters('containerImageReference'))]",
```
cwrw: we might need multi-container for admin as well in the future, but I guess for now this is ok

azure/template.json
```diff
@@ -341,7 +343,7 @@
             "value": "[concat(variables('appServicePlanName'),'-AZURE_IP-')]"
           },
           "ipAddresses": {
-            "value": "[reference('app-service').outputs.possibleOutboundIpAddresses.value]"
+            "value": "[concat(reference('app-service').outputs.possibleOutboundIpAddresses.value)]"
```
cwrw: do we not have a whitelist for the restricted db? although I remember we could connect to it unless I'm mistaken 🤔

azure/template.json
```diff
@@ -556,12 +562,124 @@
       "dependsOn": [
         "app-service-plan"
       ]
+    },
+    {
+      "name": "admin-service",
```
cwrw: do we need to add something for admin logging as well? see https://github.com/DFE-Digital/get-help-to-retrain/blob/master/azure/template.json#L383

azure-pipelines-deploy-template.yml
```diff
@@ -40,6 +40,7 @@ jobs:
 
             - task: AzureAppServiceManage@0
               displayName: 'Start Azure App Service: ${{parameters.environment}}'
+              condition: ne(variables['appServiceName'], 'disabled')
```
cwrw: shouldn't this be `adminServiceName` ?

azure/config/dev2.parameters.json
mafonso: I don't think it's needed yet (or ever) it's usage is really low, and we probably won't have an admin area there. So it keeps the costs down until we need it.

azure/template.json
mafonso: We can swap the RunTimeStacks then.   If the worker goes into the admin area it is not needed on the app instance by then. So we can and should swap it.

azure/template.json
mafonso: No, diagnostic logs messages are sent to the same Event Hub, as long as the app logs to stdout it's fine.

azure-pipelines-deploy-template.yml
mafonso: I agree that the name is similar and confusing now that we have two.

That's the variable inside the TaskGroup

image

It used the match the name that is passed in, but now we have two.  That can't be ranamed because that's the internal variable that the task expects. We are setting it with the correct one depending on the instance.
 
image

PR #399, opened 2020-01-06
Upgrade to node version 12.14.0

cwrw:
* upgrade the node version file and yarn packages
* updgrade pipeline version


azure-pipelines.yml
```diff
@@ -43,7 +43,7 @@ stages:
     steps:
       - task: NodeTool@0
         inputs:
-          versionSpec: '10.14.2'
+          versionSpec: '12.13.0'
```
mafonso: Why 12.13.0 if the default `yarn install`went for 12.14.0? Should we not aim to this version as well?

`error @: The engine "node" is incompatible with this module. Expected version "10.x.x". Got "12.14.0"
`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "engines": {
-    "node": "10.x.x"
+    "node": "12.x.x"
```
mafonso: But this should cover the default installed version anyway.. so it will be fine.

azure-pipelines.yml
cwrw: I was going with what was on the vmimage pool, but it seems we can install our own, amending: https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md#nodejs

mafonso: Fair point. I was only thinking at the docker build aspect.
Let's see if this branch runs ok then.

PR #401, opened 2020-01-06
Enable admin mode on qa

mafonso:
### Context

This builds on top of the Admin Conditional deployment to enable it on QA

### DIFF
--- a/azure-pipelines-deploy-template.yml
+++ b/azure-pipelines-deploy-template.yml
@@ -36,11 +36,39 @@ jobs:
                 csmFile: '$(Pipeline.Workspace)\ARMTemplates\azure\template.json'
                 csmParametersFile: '$(Pipeline.Workspace)\ARMTemplates\azure\config\${{parameters.environment}}.parameters.json'
                 overrideParameters: '-containerImageReference "${{parameters.containerImageReference}}" -dockerComposeFile "$(dockerCompose)"  -supportEmailAddresses ${{parameters.supportEmailAddresses}}'
-                deploymentOutputs: DeploymentOutput
+                deploymentOutputs: ArmOutputs
+
+            - powershell: |
+                function Convert-ArmOutputToPsObject {
+                  param (
+                    [Parameter(Mandatory=$true)]
+                    [string]
+                    $ArmOutputString
+                  )
+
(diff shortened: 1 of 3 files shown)


azure-pipelines-deploy-template.yml
```diff
@@ -68,3 +68,34 @@ jobs:
                 SpecifySlotOrASE: true
                 resourceGroupName: '${{parameters.resourceGroupName}}'
                 Slot: staging
+
+            - task: AzureAppServiceManage@0
+              displayName: 'Start Azure Admin Service: ${{parameters.environment}}'
+              condition: ne(variables['appServiceName'], 'disabled')
+              inputs:
+                azureSubscription: '${{parameters.subscriptionServiceConnection}}'
+                Action: 'Start Azure App Service'
```
cwrw: can we change the action to admin service as well, here and below?

mafonso: The action is a reserved "keyword" for the task.  
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-app-service-manage.

azure-pipelines-deploy-template.yml
```diff
@@ -1,5 +1,5 @@
 parameters:
-  debug: false
+  debug: true
```
cwrw: do we need to keep this as true?

mafonso: No, I can remove it. It does not seem to work anyway. That should give you more details on the job output, which is handy to have enabled.
I'll disable it again and revisit this later.

azure-pipelines-deploy-template.yml
```diff
@@ -36,11 +36,39 @@ jobs:
...
+                      } elseif ($type -eq "string") {
+                          Write-Host "##vso[task.setvariable variable=$keyname]$value"
+                          Write-Host "Added Azure DevOps variable '$keyname' ('$type') with value '$value'"
+                      } else {
+                          Throw "Type '$type' is not supported for '$keyname'"
+                      }
+                  }
+                }
+                Convert-ArmOutputToPsObject -ArmOutputString '$(ArmOutputs)'
+              displayName: "Parsing outputs from ARM Deployment into pipeline variables"
```
tavy88: Interesting script - take a bit of time to digest.

PR #404, opened 2020-01-08
Fix inflated users stats and location in GA

tavy88:
To fix the inflated stats on users when sending custom
events and also the reported location we need to enrich
the payload we send as follows:

- cid will gold the _ga cookie value (as unique identifier)
- uip will hold the client's IP
- ua will hold the user agent information

### Ticket ID

https://dfedigital.atlassian.net/browse/GET-879

### DIFF
--- a/app/helpers/ga_tracking_helper.rb
+++ b/app/helpers/ga_tracking_helper.rb
@@ -11,6 +11,14 @@ module GaTrackingHelper
     )
   end
 
+  def client_tracking_data
+    {
+      ga_cookie: cookies['_ga'],
+      ip_address: request.remote_ip,
+      user_agent: request.env['HTTP_USER_AGENT']
+    }
+  end
+
   private
 
   def build_event_props(key:, label:, values:)
--- a/app/helpers/ga_tracking_helper.rb
+++ b/app/helpers/ga_tracking_helper.rb
@@ -13,7 +13,7 @@ module GaTrackingHelper
 
   def client_tracking_data
     {
-      ga_cookie: cookies['_ga'],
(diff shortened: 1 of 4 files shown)


app/helpers/ga_tracking_helper.rb
```diff
@@ -11,6 +11,14 @@ def track_selections(selected: {}, unselected: {})
     )
   end
 
+  def client_tracking_data
+    {
+      ga_cookie: cookies['_ga'],
```
cwrw: Tested everything locally and on Heroku and it looked good. The only issue I've noticed is that the front end events (i.e. outbound links) were not in sync with backend events and appeared as two users. On inspection of the payload sent it seems that the `cid` being sent from the frontend:

````
cid=336203544.1578517089
# correctly parsed by gtag
```
and from the backend:
```
cid=GA1.1.336203544.1578517089
# needs parsing to remove the `GA1.1`
```
are different. This is because the client id is the last two sections of the value we are extracting from the `_ga` cookie: `336203544.1578517089` rather than the full value including `GA1.1`.
To resolve this, we can parse this value to remove the first two sections, and then both will be in sync. We could use `gsub` as follows or maybe find a nicer way: `cookies['_ga']&.gsub(/^(.*?\..*?\.)/,"")`.

tavy88: Yeah I can see it in GA. The `GAX.Y` should be stripped. https://stackoverflow.com/questions/31854752/how-to-get-the-client-id-while-sending-data-to-ga-using-measurement-protocol

PR #405, opened 2020-01-09
Add default project tags to all resources

mafonso:
### Context

https://dfedigital.atlassian.net/browse/GET-873

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -2,6 +2,13 @@
   "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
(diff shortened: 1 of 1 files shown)


azure/template.json
```diff
@@ -2,6 +2,13 @@
   "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
+    "resourceTags": {
+        "type": "object",
+        "defaultValue": {
+            "Parent Business Service": "Teacher Training and Qualifications",
```
cwrw: 🦂

mafonso: This was what we were told to add..
image

PR #406, opened 2020-01-09
GET-809 Add Authorisation to Admin interface

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-809

Using the cancancan gem adapter as well as Azure AD groups, add authorisation to the admin interface.

A roles column has been added to users, which allows users to be linked to multiple roles. Roles are hardcoded and maintained in the code. We made this decision to avoid having to maintain in the database and the code, but are open to changing this if required by the business.

~For a user to hold multiple roles we use a roles mask, which uses bitmask logic to calculate an integer based on the role and position.~

Since we have access to a roles list from the authentication hash, map those ids to the hardcoded roles we have and determine the role a user belongs to. I removed the session mapping as that is no longer needed.

Finally we define the abilities for each roles, which plays nicely with active admin.

Useful links:
https://github.com/watson/active_admin/wiki/How-to-work-with-cancan
https://github.com/ryanb/cancan/wiki/Role-Based-Authorization
https://github.com/CanCanCommunity/cancancan/wiki/defining-abilities


app/models/admin_user.rb
```diff
@@ -1,25 +1,50 @@
...
+  # convert list of roles to bitmask
+  def roles=(roles)
+    self.roles_mask = (roles & ROLES).map { |r| 2**ROLES.index(r) }.sum
+  end
 
-    # We are extracting ids instead of names as ids will not change, while names could change
-    roles_data.pluck(:id)
+  # convert bitmask back to list of roles
+  def roles
+    ROLES.reject { |r| ((roles_mask || 0) & 2**ROLES.index(r)).zero? }
```
tavy88: Discussed in person, I feel that simply adding a `roles` array would help us get around the complexity of using bitmask which makes the code very hard to read (unless one knows exactly how the algorithm works).  Also, because our table is so small the efficiency we would get using bitmask doesn't I think justify the extra complexity introduced.

Here's a small snippet I had locally:

```ruby
irb(main):001:0> AdminUser.last
=> #<AdminUser id: 1, resource_id: "9434b582-757d-423c-80cb-30cee265f95c", name: "octavian.neguletu", email: [FILTERED], created_at: "2020-01-01 13:07:31", updated_at: "2020-01-10 16:50:27", roles_mask: 5, roles: ["management", "read"]>
```

This way we can get rid of `roles=` and `role` overrides.

app/models/admin_user.rb
```diff
@@ -1,25 +1,50 @@
...
 
-  def roles_from(auth_hash)
-    roles_data = auth_hash.dig(:extra, :user_roles, :value)
+  def self.groups_to_roles_mapping
+    {
+      Rails.configuration.azure_management_role_id => ROLES[0],
+      Rails.configuration.azure_readwrite_role_id => ROLES[1],
+      Rails.configuration.azure_read_role_id => ROLES[2]
+    }.reject { |k, _v| k.blank? }
+  end
```
tavy88: Let's see if we can slightly rethink this as it does look a bit weird as we discussed.

PR #408, opened 2020-01-09
GET-828 Growth nil job profiles should appear in the end rather than beginning

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-828

Use custom ordering for growth embed in other sorting orders. This
specifies that growth nil should come last, but maintains growth in
decsending order.
https://stackoverflow.com/questions/5826210/rails-order-with-nulls-last

Also add `No data available` on nil growth job profiles.
Regarding non recommended jobs, I've run the rake task on heroku, but we need to do this on azure and copy data when this is in qa

### DIFF
--- a/app/decorators/job_profile_decorator.rb
+++ b/app/decorators/job_profile_decorator.rb
@@ -87,8 +87,7 @@ class JobProfileDecorator < SimpleDelegator # rubocop:disable Metrics/ClassLengt
   end
 
   def growth_type
-    return unless growth
-
+    return 'No data available' unless growth
     return 'Falling' if growth <= -5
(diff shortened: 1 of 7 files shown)


tavy88:
@cwrw What about:

`4. For all jobs have no job growth data available, instead of ‘falling’ etc. always show “No data available”`

app/decorators/job_profile_decorator.rb
```diff
@@ -87,8 +87,7 @@ def growth_icon
   end
 
   def growth_type
-    return unless growth
-
+    return 'No data available' unless growth
```
tavy88: Perfect!

PR #410, opened 2020-01-10
Get 895  - Make custom hostname  parameterisable for both the app and the admin area 

mafonso:
### Context
https://dfedigital.atlassian.net/browse/GET-895

### DIFF
--- a/azure/config/dev1.parameters.json
+++ b/azure/config/dev1.parameters.json
@@ -147,8 +147,14 @@
         "secretName": "dev-azureReadRoleId"
       }
     },
-    "customHostName": {
-      "value": "dev1.nrs-ghtr.org.uk"
+    "appCustomHostName": {
+      "value": "dev1"
+    },
+    "adminCustomHostName": {
+      "value": "admin-dev1"
+    },
+    "baseDomainName": {
+      "value": "nrs-ghtr.org.uk"
     },
     "certificateName": {
       "value": "nrs-ghtr-org-uk"
--- a/azure/config/qa.parameters.json
+++ b/azure/config/qa.parameters.json
@@ -147,8 +147,14 @@
         "secretName": "qa-azureReadRoleId"
       }
     },
-    "customHostName": {
-      "value": "qa-origin.nrs-ghtr.org.uk"
+    "appCustomHostName": {
+      "value": "qa-origin"
+    },
+    "adminCustomHostName": {
+      "value": "admin-qa"
+    },
+    "baseDomainName": {
+      "value": "nrs-ghtr.org.uk"
     },
     "certificateName": {
       "value": "nrs-ghtr-org-uk"
--- a/azure/template.json
+++ b/azure/template.json
@@ -16,11 +16,25 @@
         "description": "The resource group for the key vault."
       }
     },
-    "customHostName": {
+    "appCustomHostName": {
       "type": "string",
       "defaultValue": "",
       "metadata": {
         "description": "The custom hostname to add to the app service."
+      }
+    },
+    "adminCustomHostName": {
+      "type": "string",
+      "defaultValue": "",
+      "metadata": {
+        "description": "The custom hostname to add to the admin service."
+      }
+    },
+    "baseDomainName": {
+      "type": "string",
+      "defaultValue": "",
+      "metadata": {
+        "description": "The base domain name for the service. the service hostname will be prefixed to this to compose a FQDN "
       }
     },
     "certificateName": {
@@ -214,7 +228,9 @@
     "environmentName": "[split(variables('resourceNamePrefix'),'-')[1]]",
     "appServiceName": "[concat(variables('resourceNamePrefix'), '-as')]",
     "adminServiceName": "[concat(variables('resourceNamePrefix'), 'admin-as')]",
-    "keyvaultCertificateName": "[if(greater(length(parameters('certificateName')),0), parameters('certificateName'), replace(parameters('customHostName'), '.', '-'))]",
+    "appCustomFQDN": "[concat(parameters('appCustomHostName'), '.', parameters('baseDomainName'))]",
+    "adminCustomFQDN": "[concat(parameters('adminCustomHostName'), '.', parameters('baseDomainName'))]",
+    "keyVaultCertificateName": "[if(greater(length(parameters('certificateName')),0), parameters('certificateName'), replace(parameters('baseDomainName'), '.', '-'))]",
     "storageAccountName": "[replace(concat(variables('resourceNamePrefix'), 'str'), '-', '')]",
     "appServicePlanName": "[concat(variables('resourceNamePrefix'), '-asp')]",
     "domainDatabaseServerName": "[concat(variables('resourceNamePrefix'), '-', parameters('domainDatabaseName'),'Psql')]",
@@ -228,7 +244,7 @@
     {
       "name": "storage-account",
       "type": "Microsoft.Resources/deployments",
-      "apiVersion": "2019-08-01",
+      "apiVersion": "2019-10-01",
       "properties": {
         "mode": "Incremental",
         "templateLink": {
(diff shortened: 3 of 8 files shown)


azure/config/dev1.parameters.json
```diff
@@ -123,8 +123,14 @@
...
-      "value": "dev1.nrs-ghtr.org.uk"
+    "appCustomHostName": {
+      "value": "dev1"
+    },
+    "adminCustomHostName": {
+      "value": "admin-dev1"
+    },
+    "baseDomainName": {
+      "value": "nrs-ghtr.org.uk"
     },
```
mafonso: Split the FQDN into a common base domain and specific hostnames

azure/template.json
```diff
@@ -16,13 +16,27 @@
...
+      }
+    },
+    "baseDomainName": {
+      "type": "string",
+      "defaultValue": "",
+      "metadata": {
+        "description": "The base domain name for the service. the service hostname will be prefixed to this to compose a FQDN "
+      }
+    },
     "certificateName": {
```
mafonso: Take the new split parameter organisation

azure/template.json
```diff
@@ -193,7 +210,9 @@
     "environmentName": "[split(variables('resourceNamePrefix'),'-')[1]]",
     "appServiceName": "[concat(variables('resourceNamePrefix'), '-as')]",
     "adminServiceName": "[concat(variables('resourceNamePrefix'), 'admin-as')]",
-    "keyvaultCertificateName": "[if(greater(length(parameters('certificateName')),0), parameters('certificateName'), replace(parameters('customHostName'), '.', '-'))]",
+    "appCustomFQDN": "[concat(parameters('appCustomHostName'), '.' ,parameters('baseDomainName'))]",
+    "adminCustomFQDN": "[concat(parameters('adminCustomHostName'), '.' ,parameters('baseDomainName'))]",
+    "keyVaultCertificateName": "[if(greater(length(parameters('certificateName')),0), parameters('certificateName'), replace(parameters('baseDomainName'), '.', '-'))]",
```
mafonso: Compose each FQDN from it's parts.

azure/template.json
```diff
@@ -193,7 +210,9 @@
     "environmentName": "[split(variables('resourceNamePrefix'),'-')[1]]",
     "appServiceName": "[concat(variables('resourceNamePrefix'), '-as')]",
     "adminServiceName": "[concat(variables('resourceNamePrefix'), 'admin-as')]",
-    "keyvaultCertificateName": "[if(greater(length(parameters('certificateName')),0), parameters('certificateName'), replace(parameters('customHostName'), '.', '-'))]",
+    "appCustomFQDN": "[concat(parameters('appCustomHostName'), '.' ,parameters('baseDomainName'))]",
+    "adminCustomFQDN": "[concat(parameters('adminCustomHostName'), '.' ,parameters('baseDomainName'))]",
```
mafonso: Compose each FQDN from it's parts.

azure/template.json
```diff
@@ -207,7 +226,7 @@
     {
       "name": "storage-account",
       "type": "Microsoft.Resources/deployments",
-      "apiVersion": "2019-08-01",
+      "apiVersion": "2019-10-01",
```
mafonso: Just API version bumping.

azure/template.json
```diff
@@ -387,10 +406,10 @@
       ]
     },
     {
-      "condition": "[greater(length(parameters('customHostName')), 0)]",
+      "condition": "[or(greater(length(variables('appCustomFQDN')), 0), greater(length(variables('adminCustomFQDN')), 0))]",
       "name": "app-service-certificate",
```
mafonso: We will need to get the certificate if either instance uses a custom hostname.

azure/config/dev1.parameters.json
```diff
@@ -123,8 +123,14 @@
         "secretName": "dev-azureTenantId"
       }
     },
-    "customHostName": {
-      "value": "dev1.nrs-ghtr.org.uk"
+    "appCustomHostName": {
+      "value": "dev1"
+    },
+    "adminCustomHostName": {
+      "value": "admin-dev1"
```
cwrw: I think this name read better, I was wondering what changed to allow us to use a dash as opposed to our initial `admindev1`?

azure/config/qa.parameters.json
```diff
@@ -123,8 +123,14 @@
         "secretName": "qa-azureTenantId"
       }
     },
-    "customHostName": {
-      "value": "qa-origin.nrs-ghtr.org.uk"
+    "appCustomHostName": {
+      "value": "qa-origin"
```
cwrw: I was wondering why are we going with `-origin` on test subscription? I was expecting origin only on prod

azure/template.json
```diff
@@ -16,13 +16,27 @@
...
       "defaultValue": "",
       "metadata": {
         "description": "The custom hostname to add to the app service."
       }
     },
+    "adminCustomHostName": {
+      "type": "string",
+      "defaultValue": "",
+      "metadata": {
+        "description": "The custom hostname to add to the app service."
```
cwrw: could we update this description to include admin?

azure/template.json
```diff
@@ -172,6 +186,9 @@
         "description:": "Azure Active Directory - TENANT ID"
       }
     },
+
```
cwrw: 🦂

azure/template.json
```diff
@@ -193,7 +210,9 @@
     "environmentName": "[split(variables('resourceNamePrefix'),'-')[1]]",
     "appServiceName": "[concat(variables('resourceNamePrefix'), '-as')]",
     "adminServiceName": "[concat(variables('resourceNamePrefix'), 'admin-as')]",
-    "keyvaultCertificateName": "[if(greater(length(parameters('certificateName')),0), parameters('certificateName'), replace(parameters('customHostName'), '.', '-'))]",
+    "appCustomFQDN": "[concat(parameters('appCustomHostName'), '.' ,parameters('baseDomainName'))]",
```
cwrw: could you amend the spacing of `'.' ,parameters` to `'.', parameters` :)?

azure/template.json
```diff
@@ -704,10 +723,10 @@
             ]
           },
           "customHostName": {
-            "value": "[concat('admin',parameters('customHostName'))]"
+            "value": "[variables('adminCustomFQDN')]"
           },
           "certificateThumbprint": {
-            "value": "[if(greater(length(parameters('customHostname')), 0), reference('app-service-certificate').outputs.certificateThumbprint.value, '')]"
+            "value": "[if(greater(length(parameters('adminCustomHostname')), 0), reference('app-service-certificate').outputs.certificateThumbprint.value, '')]"
```
cwrw: to align to line 605, should this also be `adminCustomFQDN`?

azure/config/qa.parameters.json
mafonso: We have a similar setup it on UAT and QA as well.   
UAT so that it a replication of Prod, and QA because we have/had to test changes with Frontdoor on occasion.  

However regarding UAT we don't have it (yet) because we pointed the Front Door directly to the internal azure hostname. 

The changes on this PR just keep things as they are, but there is a question if we need to have custom names at all and it we do then we need to make this consistent. 

We can make another PR to add the "origin" custom hostname to UAT for consistency reasons, as it seems to be the generality of the cases.

azure/template.json
mafonso: Good spot.

azure/template.json
mafonso: It can for alignment reasons. 
In practice it does not make any difference, as if the hostname is defined the FQDN will be as well.
But it should be consistent yes.

azure/config/dev1.parameters.json
mafonso: The initial form was due to a limitation of the previous setup.  The dash was the suggested for but it would make it harder to deal with the special case of the production environment. 
Since this approach gives us the flexibility I opted to go back to the dash form.

PR #413, opened 2020-01-13
Tell Omniauth which callback url to use

tavy88:
Because we have different envs we require multiple callback
urls defined in Azure AD Directory.

In order for rack to pick the right one, we need to
explicitly mention that URL.


tavy88:
Looks like it still doesn't do the trick for us.

PR #414, opened 2020-01-13
GET-859 Remove duplicate banner landmark

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-859

* Stop fixing cookie banner to top to allow it to show the header
* Remove duplicate header from cookie banner partial
* Place cookie banner after header, so now it shows.
Only difference now is that we see the blue border (checking with lois on this)

Testing:
point axe tests to local and run, don't show failure anymore


tavy88:
Hmm it's also loosing the gray background as you scroll down. Looks a bit weird and it looks like the site is gradually made available as you scroll down. Have you checked this behaviour with Lois? It doesn't look right.

PR #417, opened 2020-01-14
Trap focus inside cookies modal

tavy88:
# Context
Prevent users from tabbing anywhere outside the cookies
modal when this is visible.

Inspired from: https://stackoverflow.com/questions/32996817/vanilla-javascript-trap-focus-in-modal-accessibility-tabbing

And other similar resources.

# Ticket
https://dfedigital.atlassian.net/browse/GET-843

### DIFF
--- a/app/webpacker/packs/cookies-banner.js
+++ b/app/webpacker/packs/cookies-banner.js
@@ -5,7 +5,7 @@ function CookiesBanner () {
     if (typeof(cookiesModal) !== 'undefined' && cookiesModal != null) {
       if (getCookie('seen_cookie_message') !== 'true') {
         displayCookiesModal(cookiesModal);
-        trapFocus(cookiesModal)
+        trapFocus(cookiesModal);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


app/webpacker/packs/cookies-banner.js
```diff
@@ -40,6 +41,33 @@ function CookiesBanner () {
       setCookie('seen_cookie_message', 'true', 30);
     }
   }
+
+  function trapFocus(element) {
+    var focusableElements = element.querySelectorAll('a[href]:not([disabled])');
+    var acceptCookies = focusableElements[2];
```
cwrw: discussed in person but for posterity: would be good to include the header elements as well in tabbing, and also to investigate using specific elements instead of array.

tavy88: sure, keen to see how things look like after you've merged yours.

tavy88: I prefer a more generic approach where we don't care which the focusable elements are as long as we trap the focus between 1st and last. The ids can change so can the markup. I think this approach is more bullet proof for further changes.

PR #418, opened 2020-01-14
Fix bad location of tags parameter on admin webapp resource

mafonso:
### Context
https://dfedigital.atlassian.net/browse/GET-873

The parameter object for the tags was being passed into the wrong place on the ARM template structure for a particular resource.  This is a fix for that "bug"

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -674,15 +674,15 @@
       "condition": "[parameters('adminEnabled')]",
       "apiVersion": "2019-10-01",
       "properties": {
-        "resourceTags":{
-            "value": "[parameters('resourceTags')]"
-        },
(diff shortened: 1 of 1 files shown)


azure/template.json
```diff
@@ -674,15 +674,15 @@
...
-        },
         "mode": "Incremental",
         "templateLink": {
           "uri": "[concat(variables('deploymentUrlBase'), 'app-service-linux.json')]",
           "contentVersion": "1.0.0.0"
         },
         "parameters": {
+          "resourceTags":{
+              "value": "[parameters('resourceTags')]"
+          },
```
tavy88: I see so it was nested under the wrong key.

PR #419, opened 2020-01-14
GET-859 Remove duplicate banner landmark

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-859

* Remove duplicate header from cookie banner partial
* Use js to move header div inside fixed div when cookie banner is visible
  and move it back out when done. If cookie banner is not visible, it will
  appear as normal.
  I've played around with this and another option would be to add a div around
  both cookies and header and fix that, but this seemed like the least obtrusive way.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -26,12 +26,14 @@
     </script>
     <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
     <%= render 'shared/cookies_banner' unless current_page?(cookies_policy_path) %>
-    <header class="govuk-header" role="banner" data-module="govuk-header">
-      <div class="govuk-header__container govuk-width-container">
-        <%= render 'shared/navigation/header_logo' %>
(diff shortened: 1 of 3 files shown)


app/views/layouts/application.html.erb
```diff
@@ -26,12 +26,14 @@
...
-      </div>
-    </header>
+    <div id="govuk-header-container">
+      <header class="govuk-header" role="banner" data-module="govuk-header" id="govuk-header">
+        <div class="govuk-header__container govuk-width-container">
+          <%= render 'shared/navigation/header_logo' %>
+          <%= render 'shared/navigation/main_header' %>
+        </div>
+      </header>
+    </div>
```
tavy88: Discussed in person, another option would have been to use conditional statements e.g

```ruby
if cookies[:key] == 'accepted modal' 
  put the div here (and it would be hidden as long as the modal is present)
```
And then use something similar on the modal side. That would remove the JS complexity and fiddling with the DOM dynamically. However, happy with this approach as well.

PR #422, opened 2020-01-15
Audit searches on PID

tavy88:
### Context
Re-using the object changes field, and
capture the search query and results count while on
PID on auditing.

### Screenshot

Screen Shot 2020-01-15 at 11 50 41

### Notes
Useful resource for retrieving the full filtered collection (not just first page of results): https://stackoverflow.com/questions/26215985/activeadmin-access-filtered-collection

### Ticket
https://dfedigital.atlassian.net/browse/GET-878

### DIFF
--- a/app/admin/paper_trail_versions.rb
+++ b/app/admin/paper_trail_versions.rb
@@ -39,9 +39,16 @@ if defined?(ActiveAdmin)
         row :changes do |version|
           if version.object_changes.present?
             changes_hash = version.object_changes.except('updated_at')
-            changes_hash.map { |k, v|
-              "<strong>#{k}: #{v.compact.join(' -> ')}</strong>"
-            }.compact.join('<br>').html_safe
+            if version.event == 'search'
+              "<strong>Searched for</strong>:<br>
+                #{changes_hash['query'].map { |k, v| "#{k}: #{v}" }.join('<br>')}".concat(
+                  "<br><br><strong>Results found</strong>: #{changes_hash['results']}"
+                ).html_safe
+            else
+              changes_hash.map { |k, v|
+                "<strong>#{k}: #{v.compact.join(' -> ')}</strong>"
+              }.compact.join('<br>').html_safe
+            end
           end
         end
         row :created_at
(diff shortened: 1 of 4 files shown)


app/admin/paper_trail_versions.rb
```diff
@@ -39,9 +39,16 @@
         row :changes do |version|
           if version.object_changes.present?
             changes_hash = version.object_changes.except('updated_at')
-            changes_hash.map { |k, v|
-              "<strong>#{k}: #{v.compact.join(' -> ')}</strong>"
-            }.compact.join('<br>').html_safe
+            if version.event == 'search'
+              "<strong>Searched for</strong>:<br>
+                #{changes_hash['query'].map { |k, v| "#{k}: #{v}" }.join('<br>')}".concat(
```
cwrw: protect against empty filter

app/admin/user_personal_data.rb
```diff
@@ -23,6 +23,18 @@ def destroy
 
         super
       end
+
+      def index
+        if params[:commit] == 'Filter'
+          track_custom_event(
+            item_type: 'User Personal Data Page',
+            event: 'search',
+            changes: { query: params[:q], results: collection.count }
```
cwrw: count full dataset returned instead of first page

app/admin/paper_trail_versions.rb
cwrw: add test in features

app/admin/user_personal_data.rb
tavy88: yeah looks like we need to use ransack.

app/admin/user_personal_data.rb
```diff
@@ -23,6 +23,18 @@ def destroy
 
         super
       end
+
+      def index
+        if params[:commit] == 'Filter'
+          track_custom_event(
+            item_type: 'User Personal Data Page',
+            event: 'search',
+            changes: { query: params[:q], results: UserPersonalData.ransack(params[:q]).result.count }
```
cwrw: could we use `collection.total_count` instead of rerunning the query? `total_count` is available from https://github.com/kaminari/kaminari#the-per-scope

tavy88: Yeah could do. The query will probably be cached but yeah we can use total_count as well.

PR #426, opened 2020-01-16
Enable admin on UAT

mafonso:
### Context

https://dfedigital.atlassian.net/browse/GET-901

Unable Admin Area on UAT environment.

### DIFF
--- a/azure/config/uat.parameters.json
+++ b/azure/config/uat.parameters.json
@@ -114,6 +114,9 @@
     },
     "appServicePlanInstances": {
       "value": 2
+    },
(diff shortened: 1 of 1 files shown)


azure/config/uat.parameters.json
```diff
@@ -114,6 +114,9 @@
     },
     "appServicePlanInstances": {
       "value": 2
+    },
+    "adminEnabled": {
+      "value": true
```
tavy88: We'll need to add the keys in the vault as well before this goes in.

PR #428, opened 2020-01-17
Fix the following Axe failure:

tavy88:
### Context

Error: the array [
  {
    "issues": [
      {
        "description": "Ensures every form element has a label"
        "help": "Form elements must have labels"

https://dequeuniversity.com/rules/axe/3.3/label?application=axeAPI

### Test

Should not see the above error.


tavy88:
Ok let's do this stretch on the same ticket why not.

PR #430, opened 2020-01-17
Update Job Profile page markup

tavy88:
### Context
Following changes are introduced:

1. Select type of work changes into Apply to become section
2. This new section since it's the last and has a special style
for its delimiter, we are moving the dynamic sections' delimiters
on top of each section rather than after, so that the last dynamic
section's delimiter does not collide with the custom one that belongs
to Apply to become section.
3. Since concat/+ in ruby escape strings, we need to call an html_safe
on the our dynamic sections' bodies when prepending the delimiter.

### Ticket
https://dfedigital.atlassian.net/browse/GET-888

### DIFF
--- a/app/decorators/job_profile_decorator.rb
+++ b/app/decorators/job_profile_decorator.rb
@@ -29,7 +29,11 @@ class JobProfileDecorator < SimpleDelegator # rubocop:disable Metrics/ClassLengt
   end
 
   def how_to_become
-    html_body.xpath(HOW_TO_BECOME_XPATH).text
+    @how_to_become ||= html_body.xpath(HOW_TO_BECOME_XPATH).text
+  end
+
+  def become
+    how_to_become.gsub('How to', '').strip
   end
 
   def salary_range
@@ -62,6 +66,7 @@ class JobProfileDecorator < SimpleDelegator # rubocop:disable Metrics/ClassLengt
     html_body.xpath(ADDITIONAL_COPY_XPATH).children.map(&:text)
   end
 
+  # TODO: Refactor this method and separate the apprenticeship logic
   def section(key)
     return unless (xpath = SECTIONS_XPATH[key])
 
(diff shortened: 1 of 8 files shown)


app/views/job_profiles/_jobs_in_your_area.html.erb
```diff
@@ -1,8 +1,21 @@
-<% return unless @job_vacancy_count.present? %>
```
cwrw: since this is no longer only a jobs in your area section could we rename this?

app/decorators/job_profile_decorator.rb
```diff
@@ -72,7 +76,9 @@ def section(key)
     mutate_apprenticeship_content if key == :apprenticeship
     mutate_html_body
 
-    @doc.to_html.gsub(%r{<a.*?>(.+?)</a>}, '\1').concat(separator_line)
+    return @doc.to_html.gsub(%r{<a.*?>(.+?)</a>}, '\1').html_safe if key == :apprenticeship
```
cwrw: could you add a test for this?

app/decorators/job_profile_decorator.rb
```diff
@@ -72,7 +76,9 @@ def section(key)
...
+
+    separator_line.concat(@doc.to_html.gsub(%r{<a.*?>(.+?)</a>}, '\1').html_safe)
```
cwrw: did you have a chance to look at `to_html/html_safe`?

app/decorators/job_profile_decorator.rb
cwrw: also `@doc.to_html.gsub(%r{<a.*?>(.+?)</a>}, '\1').html_safe` seems to be shared, we could create a variable for that to be used either for apprenticeships or all other cases, what do you think?

app/views/job_profiles/_jobs_in_your_area.html.erb
tavy88: I would not want to rename the partial really. Let's chat about the second suggestion as it's not quite clear.

app/decorators/job_profile_decorator.rb
tavy88: yeah coud do that.

app/decorators/job_profile_decorator.rb
tavy88: not yet no.

app/decorators/job_profile_decorator.rb
tavy88: > could you add a test for this? (if its not terrible to test)

We weren't testing this bit before either as the test would only test the position of the delimiter so not sure if it's worth the complication.

PR #436, opened 2020-01-21
GET-908 Allow passing in ipSecurityRestrictions per webapp

mafonso:
### Context
https://dfedigital.atlassian.net/browse/GET-908

This PR allows passing an ipSecurityRestrictions object for each of the WebApps (app and admin) 

### Guidance to review

Uses https://github.com/DFE-Digital/get-help-to-retrain-devops/commit/6cc89a562e4973a4b73b12ceb930071892c1606e

### DIFF
--- a/azure/config/qa.parameters.json
+++ b/azure/config/qa.parameters.json
@@ -43,6 +43,18 @@
         "secretName": "qa-restrictedDatabasePassword"
       }
     },
+    "appIpSecurityRestrictions": {
+      "value": [
+        {
+          "ipAddress": "147.243.0.0/16",
+          "name": "FrontDoor ingress IPv4"
+        },
+        {
+          "ipAddress": "2a01:111:2050::/44",
+          "name": "FrontDoor ingress IPv6"
+        }
+      ]
+    },
     "secretKeyBase": {
       "reference": {
         "keyVault": {
--- a/azure/template.json
+++ b/azure/template.json
@@ -94,6 +94,20 @@
         "description": "List of IP Addresses to whitelist on the database firewall"
       }
     },
+    "appIpSecurityRestrictions": {
+      "type": "array",
+      "defaultValue": [],
+      "metadata": {
+        "description": "List of IP Address to whitelist on the Application instance Azure Webapp service "
+      }
+    },
+    "adminIpSecurityRestrictions": {
+      "type": "array",
+      "defaultValue": [],
+      "metadata": {
+        "description": "List of IP Address to whitelist on the Admin instance Azure Webapp service "
+      }
+    },
     "containerImageReference": {
       "type": "string",
       "metadata": {
@@ -568,6 +582,9 @@
           "runtimeStack": {
             "value": "[variables('appServiceRuntimeStack')]"
           },
+          "ipSecurityRestrictions": {
+            "value": "[parameters('appIpSecurityRestrictions')]"
+          },
           "appServiceAppSettings": {
             "value": [
               {
@@ -694,6 +711,9 @@
           },
           "runtimeStack": {
             "value": "[variables('adminServiceRuntimeStack')]"
+          },
+          "ipSecurityRestrictions": {
+            "value": "[parameters('adminIpSecurityRestrictions')]"
           },
           "appServiceAppSettings": {
             "value": [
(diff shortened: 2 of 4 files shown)


azure/config/dev1.parameters.json
```diff
@@ -43,6 +43,30 @@
...
+    "appIpSecurityRestrictions": {
+      "value": [
+        {
+          "ipAddress": "147.243.0.0/16",
+          "name": "FrontDoor ingress IPv4"
+        },
+        {
+          "ipAddress": "2a01:111:2050::/44",
+          "name": "FrontDoor ingress IPv6"
+        },
```
mafonso: From https://docs.microsoft.com/en-us/azure/frontdoor/front-door-faq#how-do-i-lock-down-the-access-to-my-backend-to-only-azure-front-door

azure/config/qa.parameters.json
```diff
@@ -43,6 +43,29 @@
         "secretName": "qa-restrictedDatabasePassword"
       }
     },
+    "appIpSecurityRestrictions": {
+      "value": [
+        {
+          "ipAddress": "147.243.0.0/16",
+          "name": "FrontDoor ingress IPv4"
+        },
```
mafonso: From: https://docs.microsoft.com/en-us/azure/frontdoor/front-door-faq#how-do-i-lock-down-the-access-to-my-backend-to-only-azure-front-door

azure/config/qa.parameters.json
```diff
@@ -43,6 +43,29 @@
...
+        {
+          "ipAddress": "2a01:111:2050::/44",
+          "name": "FrontDoor ingress IPv6"        },
+        {
```
mafonso: From https://docs.microsoft.com/en-us/azure/frontdoor/front-door-faq#how-do-i-lock-down-the-access-to-my-backend-to-only-azure-front-door

azure/template.json
```diff
@@ -94,6 +94,20 @@
         "description": "List of IP Addresses to whitelist on the database firewall"
       }
     },
+    "appIpSecurityRestrictions": {
+      "type": "array",
+      "defaultValue": [],
+      "metadata": {
+        "description": "List of IP Address to whitlist on the Application instance Azure Webapp service "
```
tavy88: Typo here: `whitlist`

azure/template.json
```diff
@@ -94,6 +94,20 @@
...
+      }
+    },
+    "adminIpSecurityRestrictions": {
+      "type": "array",
+      "defaultValue": [],
+      "metadata": {
+        "description": "List of IP Address to whitlist on the Admin instance Azure Webapp service "
```
tavy88: Typo here: `whitlist`

azure/config/dev1.parameters.json
```diff
@@ -43,6 +43,22 @@
         "secretName": "dev-restrictedDatabasePassword"
       }
     },
+    "appIpSecurityRestrictions": {
+      "value": [
+        {
+          "ipAddress": "217.38.175.68/32",
```
cwrw: Don't office IPs change, are we not risking to have to change the code when that happens?

azure/config/qa.parameters.json
```diff
@@ -43,6 +43,29 @@
...
+          "ipAddress": "217.38.175.68/32",
+          "name": "Office"
+        }
+      ]
+    },
+    "adminIpSecurityRestrictions": {
+      "value": [
+        {
+          "ipAddress": "217.38.175.68/32",
```
tavy88: Discussed in person about not doing this code wise and doing it from UI but I think you're right we should review changes like this anyway.

azure/config/dev1.parameters.json
mafonso: Yes that is correct. Ideally I would even add the office address here, but I can see how it is convenient.  But if it changes or you need to add more ranges to the admin area (likely) they should be added as needed.

tavy88: Ok

cwrw: discussed in person: agreed to add range instead, and confirm that DFE kit can access as well

tavy88:
Discussed in person and happy with these rules for more flexible access.

PR #437, opened 2020-01-21
GET-897 Fix progression on question pages bug

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-897

If user goes back mid journey, we should show progress indicator, until user gets
to action plan page.

Remaining behaviour remains the same.

Testing:
1) Normal journey shows normal progress
2) Interrupted journey shows only page where journey abandoned and progress
3) Going back shows progress mid journey
4) Action plan editing question shows no progress

### DIFF
--- a/app/helpers/question_helper.rb
+++ b/app/helpers/question_helper.rb
@@ -1,8 +1,6 @@
 module QuestionHelper
   def action_plan_or_questions_path
-    training_questions ||
-      it_training_questions ||
-      job_hunting_questions ||
+    questions_path ||
       action_plan_path
   end
 
@@ -18,13 +16,19 @@ module QuestionHelper
     it_training_questions_path if user_session.it_training.nil?
   end
 
-  def progression_indicator(show_indicator:, step:)
-    return unless show_indicator
+  def progression_indicator(step:)
+    return unless questions_path
 
     content_tag(
       :span,
       "Step #{step} of 3",
       class: 'govuk-caption-xl'
     )
+  end
+
+  def questions_path
+    training_questions ||
+      it_training_questions ||
+      job_hunting_questions
   end
 end
--- a/spec/helpers/question_helper_spec.rb
+++ b/spec/helpers/question_helper_spec.rb
@@ -81,14 +81,26 @@ RSpec.describe QuestionHelper do
   end
 
   describe '#progression_indicator' do
-    it 'returns nil when show indictator is false' do
-      expect(helper.progression_indicator(show_indicator: nil, step: 1)).to be_nil
+    it 'shows progress indicator and step when no questions page have been visited' do
+      expect(helper.progression_indicator(step: 1)).to eq(
+        '<span class="govuk-caption-xl">Step 1 of 3</span>'
+      )
     end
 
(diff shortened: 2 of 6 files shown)


app/helpers/question_helper.rb
```diff
@@ -18,13 +18,19 @@ def it_training_questions
...
+    return unless show_progression_indicator
 
     content_tag(
       :span,
       "Step #{step} of 3",
       class: 'govuk-caption-xl'
     )
   end
+
+  def show_progression_indicator
```
tavy88: If we find a better name for this we can dry the code even more by reusing this inside:

```ruby
def action_plan_or_questions_path
    training_questions ||
      it_training_questions ||
      job_hunting_questions ||
      action_plan_path
  end
```
We could use something like: 

```ruby
def action_plan_or_questions_path
  show_progression_indicator || action_plan_path
end
```
So you get my point why the name needs to be perhaps more intuitive.

spec/helpers/question_helper_spec.rb
```diff
@@ -81,14 +81,26 @@ def user_session
...
-      expect(helper.progression_indicator(show_indicator: nil, step: 1)).to be_nil
+    it 'shows progress indicator and step when no questions page have been visited' do
+      expect(helper.progression_indicator(step: 1)).to eq(
+        '<span class="govuk-caption-xl">Step 1 of 3</span>'
+      )
     end
 
-    it 'returns progress indicator when show indicator is true with correct page' do
-      expect(helper.progression_indicator(show_indicator: true, step: 2)).to eq(
+    it 'shows progress indicator and step when atleast one question page has been visited' do
```
tavy88: space here : `atleast`

spec/helpers/question_helper_spec.rb
```diff
@@ -81,14 +81,26 @@ def user_session
...
+
+      expect(helper.progression_indicator(step: 2)).to eq(
         '<span class="govuk-caption-xl">Step 2 of 3</span>'
       )
     end
+
+    it 'hides progress indicator if all question pages have been visited' do
+      user_session.training = ['english_skills']
+      user_session.it_training = ['computer_skills']
+      user_session.job_hunting = ['cv']
```
tavy88: So no `Step 3 of 3` ?

cwrw: not sure I understand?

cwrw: If this is regarding adding a test for step 3, at the unit level we already tested that passing in a variable appears in the html, at the feature level we are testing that we get to step 3 (check feature specs)

app/helpers/question_helper.rb
```diff
@@ -1,8 +1,6 @@
 module QuestionHelper
   def action_plan_or_questions_path
-    training_questions ||
-      it_training_questions ||
-      job_hunting_questions ||
+    questions_path ||
```
tavy88: Great looks good now!

spec/helpers/question_helper_spec.rb
tavy88: Yeah ignore me I didn't see the Step 3 out of 3

PR #443, opened 2020-01-27
New copy changes for job profiles search

tavy88:
### Context
Specific job search: update head title
Specific job search results:
- update head title
- add new paragraph below search box
Specific job search - no results:
- only the content below the search box should now change

### Ticket
https://dfedigital.atlassian.net/browse/GET-877

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -105,8 +105,8 @@ en-GB:
       no_results: 0 results found - try again using a different job title
   job_profiles:
     index:
-      title: Search for a type of work
-      body: Enter a job title to find out how well your skills match to it
+      title: Search for a type of work you're interested in
+      body: Enter a job title to find out how well your skills match to it.
       placeholder: Enter a job title
     vacancies:
       one: is
--- a/app/views/job_profiles/_no_results.html.erb
+++ b/app/views/job_profiles/_no_results.html.erb
@@ -1,0 +1,12 @@
+<h1 class="govuk-heading-xl"><%= t('title', scope: scope) %></h1>
+<p class="govuk-body-l"><%= t('body', scope: scope) %></p>
+<%= render "shared/search/results_form", search_path: { action: :results }, scope: scope %>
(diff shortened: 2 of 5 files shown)


config/locales/en.yml
```diff
@@ -105,7 +105,7 @@ en-GB:
       no_results: 0 results found - try again using a different job title
   job_profiles:
     index:
-      title: Search for a type of work
+      title: Search for a type of work you're interested in
       body: Enter a job title to find out how well your skills match to it
```
cwrw: missing full stop as per prototype

app/views/shared/search/_no_results.html.erb
```diff
@@ -1,8 +1,11 @@
 <h1 class="govuk-heading-xl">No results found</h1>
```
cwrw: This should no longer be different that the main results page as per ticket and prototype (everything above search box remains the same):
https://get-help-to-get-877-sea-4kpfjp.herokuapp.com/job-profiles/results?search=sdfsdfds
https://get-help-to-retrain-prototype.herokuapp.com/main/releaseX_e2e_v2/specific_job_search_no_results

tavy88: Yeah I understood something else from the ticket but I think it makes sense what you're saying.

app/views/shared/search/_no_results.html.erb
```diff
@@ -1,8 +1,12 @@
-<h1 class="govuk-heading-xl">No results found</h1>
+<h1 class="govuk-heading-xl"><%= t('title', scope: scope) %></h1>
```
cwrw: We are sharing this partial across both the check your skills journey as well as job profile search in the skills matcher journey. I've had a chat with Lois to confirm, but the content on both those will now be different, so we can't share the partial anymore unfortunately and need to only update the job profile search no results content for now. Happy to chat in person for clarity

tavy88: I see. I will chat with her as well.

PR #447, opened 2020-01-28
GET-899 Change content for returning from saved results to returning from saved progress instead.

jcavell:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-899

* Replace return to saved results with return to saved progress
* Keep urls the same for now
* Fix cookies page as well as error pages

### DIFF
--- a/app.json
+++ b/app.json
@@ -65,7 +65,7 @@
     }
   },
   "addons": [
-    "heroku-postgresql"
+    "     heroku-postgresql"
   ],
   "buildpacks": [
     {
--- a/app.json
+++ b/app.json
@@ -65,7 +65,7 @@
     }
   },
   "addons": [
-    "     heroku-postgresql"
+    "heroku-postgresql"
   ],
   "buildpacks": [
     {
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -20,7 +20,7 @@ en-GB:
     current_job_skills: Job skills
     user_personal_data: Your information
     save_your_results: Save your progress
-    return_to_saved_results: Return to saved results
+    return_to_saved_results: Return to saved progress
     link_sent: Link sent
     results_saved: Progress saved
     email_sent_again: Email sent again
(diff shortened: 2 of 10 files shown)


app.json
```diff
@@ -65,7 +65,7 @@
     }
   },
   "addons": [
-    "heroku-postgresql"
+    "     heroku-postgresql"
```
cwrw: could you remove this change please, as it's breaking the review app deploy (invalid json)

app/views/errors/save_results_error.html.erb
```diff
@@ -15,7 +15,7 @@
     <h1 class="govuk-heading-xl">Your results have been saved</h1>
     <p class="govuk-body">We have saved your results using the email address you have provided.</p>
     <p class="govuk-body">We cannot send you a confirmation email right now because of an error with our system. Your results are still saved.</p>
-    <p class="govuk-body">To see your saved skills and job matches in future, just return to the service and enter your email address. You will then be sent a link so you can return to your saved results.</p>
+    <p class="govuk-body">To see your saved skills and job matches in future, just return to the service and enter your email address. You will then be sent a link so you can return to your saved progress.</p>
```
cwrw: this error page is for the save your results (not return) and I don't believe we should change it

config/locales/en.yml
```diff
@@ -62,10 +62,10 @@ en-GB:
     errors_internal_server_error: Get help to retrain - Internal server error
     errors_postcode_search_error: Get help to retrain - Postcode search error
     errors_jobs_near_me_error: Get help to retrain - Jobs near me error
-    errors_return_to_saved_results_error: Get help to retrain - Return to saved results error
+    errors_return_to_saved_results_error: Get help to retrain - Return to saved progress error
```
cwrw: need to confirm page title change with lois

cwrw: need to remove those as not changing url and page name

PR #449, opened 2020-01-29
Update content for save your results journey

tavy88:
### Context
Replace save your results with save your progress.

### Ticket
https://dfedigital.atlassian.net/browse/GET-900

### DIFF
--- a/app/views/errors/save_results_error.html.erb
+++ b/app/views/errors/save_results_error.html.erb
@@ -12,10 +12,10 @@
 
 <div class="govuk-grid-row govuk-!-margin-top-7">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Your results have been saved</h1>
-    <p class="govuk-body">We have saved your results using the email address you have provided.</p>
-    <p class="govuk-body">We cannot send you a confirmation email right now because of an error with our system. Your results are still saved.</p>
-    <p class="govuk-body">To see your saved skills and job matches in future, just return to the service and enter your email address. You will then be sent a link so you can return to your saved results.</p>
+    <h1 class="govuk-heading-xl">Your progress has been saved</h1>
+    <p class="govuk-body">We have saved your progress using the email address you have provided.</p>
+    <p class="govuk-body">We cannot send you a confirmation email right now because of an error with our system. Your progress is still saved.</p>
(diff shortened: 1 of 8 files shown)


app/views/errors/save_results_error.html.erb
```diff
@@ -12,10 +12,10 @@
...
 <div class="govuk-grid-row govuk-!-margin-top-7">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Your results have been saved</h1>
-    <p class="govuk-body">We have saved your results using the email address you have provided.</p>
-    <p class="govuk-body">We cannot send you a confirmation email right now because of an error with our system. Your results are still saved.</p>
-    <p class="govuk-body">To see your saved skills and job matches in future, just return to the service and enter your email address. You will then be sent a link so you can return to your saved results.</p>
+    <h1 class="govuk-heading-xl">Your progress has been saved</h1>
+    <p class="govuk-body">We have saved your progress using the email address you have provided.</p>
+    <p class="govuk-body">We cannot send you a confirmation email right now because of an error with our system. Your progress is still saved.</p>
+    <p class="govuk-body">To see your saved skills and job matches in future, just return to the service and enter your email address. You will then be sent a link so you can return to your saved progress.</p>
```
tavy88: The prototype on the ticket is wrong as we don't use the email in the copy so I've adapted the content.

app/views/users/new.html.erb
```diff
@@ -12,15 +12,15 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl"><%= t('.title') %></h1>
     <p class="govuk-body">If you like, you can save your skills list and job matches so you can return to them later.</p>
```
cwrw: content here needs updating as well, see https://get-help-to-retrain-prototype.herokuapp.com/main/releaseX_e2e_v2/create_account

tavy88:
I don't think we want to change the path as it will skew the GA results.

config/locales/en.yml
```diff
@@ -19,7 +19,7 @@ en-GB:
     location_ineligible: No courses in your area
     current_job_skills: Job skills
     user_personal_data: Your information
-    save_your_results: Save your results
+    save_your_results: Save your progress
     return_to_saved_results: Return to saved results
     link_sent: Link sent
     results_saved: Results saved
```
cwrw: The content needs to change on the breadcrumb for the error page as well as the confirmation page

PR #450, opened 2020-01-30
GET-791 Add error summary to all forms with validation

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-791

* Add method to accept object and return an error summary listing all errors from the object
* taken from: https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/88101aec9760233f0734de69a78e4db71112733e/lib/govuk_design_system_formbuilder/elements/error_summary.rb
* Formatting adheres to https://design-system.service.gov.uk/components/error-summary/
* Errors link to where they are on form on the page. We are building the error id the same way we build it within the form
* Add helper to pages:
user personal data:
check your skills search and results:
skill builder:
job profile search and results:
courses:
job vacancy search:
save and return to saved results:

### DIFF
--- a/app/helpers/form_helper.rb
+++ b/app/helpers/form_helper.rb
@@ -1,4 +1,17 @@
 module FormHelper
+  def error_summary(object)
+    return unless object.errors.any?
+
+    content_tag('div', class: 'govuk-error-summary', **error_summary_attributes) do
+      safe_join(
+        [
+          tag.h2('There is a problem', id: 'error-summary-title', class: 'govuk-error-summary__title'),
+          error_list(object)
+        ]
+      )
+    end
+  end
+
   def form_group_tag(object, attribute, tag_class: [], &_block)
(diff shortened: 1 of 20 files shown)


app/helpers/form_helper.rb
```diff
@@ -36,4 +49,42 @@ def messages(object, attribute)
...
+      link_to(
+        message,
+        same_page_link(error_id(object_name, attribute))
+      )
+    end
+  end
+
+  def same_page_link(target)
+    '#'.concat(target)
+  end
```
tavy88: I don't think this method is necessary. How about:

```ruby
def error_list_item(object_name, attribute, message)
    content_tag('li') do
      link_to(
        message,
        "#{error_id(object_name, attribute)}"
      )
    end
  end
```

tavy88:
Hmm also make sure the build is green.

PR #451, opened 2020-01-31
Remove bottom spacing for buttons on action plan

tavy88:
### Context

Moving the spacing on top of the training and job
hunting options removes the weird spacing happening
when only one option is selected for each section.

### Ticket
https://dfedigital.atlassian.net/browse/GET-778


cwrw:
Instead of doing this on each div, we could leverage the `last-child`  ability in css. Something like this:
```
.action-plan-list  div.govuk-grid-row:last-child a {
  margin-bottom: govuk-spacing(2) !important;
}
```
and then wrap the two sections each with that div (and remove the margins from the last button):
```
<div class="action-plan-list">
</div>
```

tavy88:
It is an option however it's less readable. Also using !important; is usually not a good practice.

PR #452, opened 2020-01-31
GET-971 User unlocks save to results on PID form instead of selecting skills

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-971

* User unlocks save your results after filling in PID form
* Fix all tests for return to save results/save your results


tavy88:
You have some merge conflicts.

PR #465, opened 2020-02-11
GET-956 Track save and return events

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-956

* Track if user saves their progress
* Track if user returns to their saved progress
* Track if existing user saves their progress
* Do not track non existing user trying to save their progress
* Do not track if user enters invalid email
* Change track event method parameters to accept a scope. We
  need to do this to allow the same key to be mapped to different
  values for the save and return.

GA key on heroku for testing

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,8 +27,8 @@ class ApplicationController < ActionController::Base
 
   private
 
-  def track_event(event_key, event_value)
-    event_label = I18n.t(event_key, scope: :events)
+  def track_event(event_key, event_value, scope = 'events')
+    event_label = I18n.t(event_key, scope: scope)
 
     track_events(
       [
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -27,8 +27,8 @@ def redirect_unless_target_job
 
   private
 
-  def track_event(event_key, event_value)
-    event_label = I18n.t(event_key, scope: :events)
+  def track_event(event_key, event_value, scope = :events)
```
tavy88: I think this can be

```ruby
scope = 'events'
```

No need for it to be a symbol. I get now the complication, it all originates from the request that both events have to be under the same key.

spec/features/user_registration_spec.rb
```diff
@@ -262,4 +262,52 @@ def fill_in_user_personal_information
...
+    expect(tracking_service).to have_received(:track_events).with(
+      props:
+      [
+        {
+          key: :progress,
+          label: 'Returned to progress',
+          value: 'return_journey'
+        }
+      ]
+    )
```
tavy88: I think we should count here and ensure tracking_service has received  track_events twice.

tavy88: Discussed in person and I get the logic here. All good. You can ignore.

PR #466, opened 2020-02-12
GET-753 Add new information sources page

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-753

* Add new page and add path to routes
* Add back button which utilises the same logic as the continue
from the user save your results. This makes sure that you return to the
previous page if it exists, return to root if you didn't come from a previous
page, and return to root if you came from a page outside the app
* Override text underlining for footer lists as list should be underlined

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -6,4 +6,14 @@ class PagesController < ApplicationController
   def task_list
     @skills_summary_path = user_session.job_profile_skills? ? skills_path : check_your_skills_path
   end
+
+  def information_sources
+    @back_url = url_parser.get_redirect_path || root_path
+  end
+
+  private
+
+  def url_parser
+    @url_parser ||= UrlParser.new(request.referer, request.host)
+  end
 end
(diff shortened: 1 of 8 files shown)


app/controllers/pages_controller.rb
```diff
@@ -6,4 +6,14 @@ def action_plan
...
+
+  def information_sources
+    @back_url = url_parser.get_redirect_path || root_path
+  end
+
+  private
+
+  def url_parser
+    @url_parser ||= UrlParser.new(request.referer, request.host)
+  end
```
tavy88: Same method is defined in `users_controller.rb`. How about moving this to an appropriate helper? Also, I think the Back functionality might be re-used i other places as well and we dry the code a bit more.

cwrw: I think we should take it step by step, the next page we add will see this structured differently and so on, so I would keep it here until we find it necessary to move out

tavy88: I'm not going to insist on it but if there was a change needed to this method, you would have to change the same code in 2 places, rather than just one which would be the helper.

cwrw: So my reasoning is the next ticket I'm picking up, will also reuse this code, and there might be logic there which will help drive out the way I dry this up, so I'm agreeing with you but I think in this pr its too early, it's better to see a pattern evolve which will help structure the code in a better way and DRY it up in the appropriate manner then

PR #469, opened 2020-02-13
GET-959 Update privacy policy

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-959

* update policy content
* add back button

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -27,7 +27,6 @@ en-GB:
     link_sent_again: Link sent again
     link_expired: Link expired
     cookies_policy: Cookies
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -156,7 +155,7 @@ en-GB:
     cookies_policy:
       title: Cookies
     privacy_policy:
-      title: Privacy policy
+      title: Privacy policy for Get help to retrain
```
tavy88: So are we starting that new page title convention with the service at the end? Or is it just for this page?

tavy88:
Can you rebase?

PR #473, opened 2020-02-17
Error message (no results found) are consistent

tavy88:
### Context

Targetted pages:

- Current job skills search
- Previous job skills search
- Previous job skills search has changes to h1
and is now consistent in prototype - wasn’t mentioned in description above
- Breadcrumbs can remain the same as they are for now as this can be addressed in Sophie’s ticket

### Ticket

https://dfedigital.atlassian.net/browse/GET-913

### DIFF
--- a/app/views/job_profiles/_no_results.html.erb
+++ b/app/views/job_profiles/_no_results.html.erb
@@ -1,4 +1,4 @@
-<h1 class="govuk-heading-xl"><%= t('title', scope: scope) %></h1>
+<h1 class="govuk-heading-xl govuk-!-margin-bottom-7"><%= t('title', scope: scope) %></h1>
 <p class="govuk-body-l"><%= t('body', scope: scope) %></p>
 <%= render "shared/search/results_form", search_path: { action: :results }, scope: scope %>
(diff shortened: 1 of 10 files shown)


app/views/job_profiles/_no_results.html.erb
```diff
@@ -8,5 +8,5 @@
   <li>checking your spelling</li>
   <li>using a different name for this type of work</li>
   <li>searching for a different type of work</li>
-  <li>going back and looking at your job matches instead</li>
+  <li>going back and looking at your job matches instead</li>
```
cwrw: ✂️

tavy88:
I did think about the scope thing but either way, you want to tackle it would bring its cons and pros. We need to pass the scope as we use the same page to render different text. I would like to keep it as it is now. We might need to do a bit more refactoring which I do not want to risk rushing right now as it has the potential to introduce regressions.

tavy88:
Also, Job Profiles defect is not part of the scope of this ticket. That bit of code was not touched in the PR. It was always like that. I will create a ticket to address that small spacing defect.

PR #475, opened 2020-02-20
Get 1019 create courses tables

tavy88:
### Context

This PR creates the necessary database structure to ingest the CSV Course data. This also creates the model associations that are needed as follows:

#### Table names
- cvs_courses - our lookup table including precalculated values
- cvs_course_details - this will contain the data from courses CSV file
- csv_providers - this will contain the data from providers CSV file
- csv_opportunities - this will contain the data from opportunities CVS file
- csv_venues - this will contain the data from venues CVS file
- csv_opportunity_dates - this will contain the data from opportunity dates CVS file

#### Model associations

[primary_erd.pdf](https://github.com/DFE-Digital/get-help-to-retrain/files/4231683/primary_erd.pdf)

### Ticket
https://dfedigital.atlassian.net/browse/GET-1019

### DIFF
--- a/app/models/csv/provider.rb
+++ b/app/models/csv/provider.rb
@@ -1,6 +1,8 @@
-class Csv::Provider < PrimaryActiveRecordBase
-  self.table_name = 'csv_providers'
+module Csv
+  class Provider < PrimaryActiveRecordBase
+    self.table_name = 'csv_providers'
 
-  belongs_to :course_detail
-  has_many :courses, as: :addressable
+    belongs_to :course_detail
+    has_many :courses, as: :addressable
+  end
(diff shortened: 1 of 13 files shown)


app/models/csv/provider.rb
```diff
@@ -0,0 +1,8 @@
+module Csv
+  class Provider < PrimaryActiveRecordBase
+    self.table_name = 'csv_providers'
+
+    belongs_to :course_detail
+    has_many :courses, as: :addressable
```
cwrw: has many venues

tavy88: true

PR #477, opened 2020-02-24
Update CSV Course Models and add Factories

cwrw:
* Rename `csv_courses` table to `csv_courses_lookups` and update relations 
* Rename `csv_course_details` table to `csv_courses` and update relations
* Change `csv_course_lookups` to belong to opportunity rather than course
* Add factories for all the new csv course tables


spec/factories/csv/course_lookups.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :course_lookup, class: Csv::CourseLookup do
+    opportunity { create(:opportunity) }
+    addressable { [create(:venue), create(:provider)].sample }
+    hours { ['full time', 'part time', 'flexible'].sample }
+    delivery_type { ['online', 'work based', 'distance learning'].sample }
+    postcode { ['NW11 8QE', 'NW11 7PT', 'NW11 7HB', 'NW10 7SF', 'NW10 7NZ', 'NW10 6HJ'].sample }
+    subject { %w[maths esol english].sample }
```
tavy88: There is a variance element introduced here which could cause flakiness in our tests. How about using traits for say `esol` and `english` and let this factory maybe `math` by default? At least on my side of testing, I need to be sure exactly which is math/english etc.

cwrw: This is defining the basic factories. Traits usually come from reuse, here the initial factory will create random values. If your test depends on a certain value you will have to add it in the specific factory itself in the test. If you find that you will reuse this value feel free to add a trait for the different subjects then.

tavy88: Yeah, I know. It was just the fact that we do sampling and to even get to a course page whether it is a math/English or ESOL you need to know already know if your factory has generated any say math during the sampling process. But all good. I will be more specific on my branch.

PR #485, opened 2020-02-27
GET-1025 Add Course Filters

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-1025

* Amend course lookup importer to ignore certain study modes and attendance patterns
* Map filters directly to lookup table
* Fix tests and factories to include new mappings, as we don't want factories creating lookups that fail to be created
* Add new filter to frontend
* Amend geocoding to accept all new filter params
* Filter should respond to different media views: hide on mobile, show on desktop/tablet


app/views/courses/csv/_filter.html.erb
```diff
@@ -0,0 +1,40 @@
...
+            <% end %>
+            <div class="govuk-form-group">
+              <%= label_tag :delivery_type, 'Course type', for: 'delivery_type', class: 'govuk-label' %>
+              <%= select_tag :delivery_type, @delivery_type_options, class: 'govuk-select govuk-!-width-full', id: 'delivery_type' %>
+            </div>
+          </div>
+          <div class="govuk-grid-column-one-half">
+            <div class="govuk-form-group govuk-!-margin-bottom-3">
+              <%= label_tag :distance, 'Distance', for: 'distance', class: 'govuk-label' %>
+              <%= errors_tag_minimal @search, :postcode %>
```
cwrw: move this above label

app/models/csv/course_geospatial_search.rb
```diff
@@ -22,7 +24,15 @@ def find_courses
     private
 
     def scope
-      topic.present? ? Csv::CourseLookup.where(subject: topic) : Csv::CourseLookup.all
+      topic.present? ? Csv::CourseLookup.where(**selected_options) : Csv::CourseLookup.all
```
tavy88: We won't need `Csv::CourseLookup.all` but since we bin this eventually it makes no difference.

app/views/courses/csv/_filter.html.erb
```diff
@@ -0,0 +1,40 @@
+<div class="govuk-accordion" data-module="govuk-accordion" id="accordion-default">
+  <div class="govuk-accordion__section ">
+    <div class="govuk-accordion__section-header">
+      <h2 class="govuk-accordion__section-heading">
+        <span class="govuk-accordion__section-button" id="accordion-default-heading-1">
+          Filters
+        </span>
+      </h2>
+    </div>
+    <div id="accordion-default-content-1" class="govuk-accordion__section-content govuk-!-padding-top-0" aria-labelledby="accordion-default-heading-1">
```
tavy88: id name should be customizable. have another 👁

spec/features/csv/find_training_courses_spec.rb
```diff
@@ -54,12 +68,211 @@
...
+    course_lookup = create(:course_lookup, latitude: 0.1, longitude: 1.5, subject: 'maths')
+    capture_user_location('NW6 8ET')
+    visit(courses_path(topic_id: 'maths'))
+    select('Up to 40 miles', from: 'distance')
+    click_on('Apply filters')
+
+    expect(page).to have_text(course_lookup.course.name)
+  end
+
+  scenario 'User does not see unselected distances' do
```
tavy88: I think this can be a bit more descriptive.

app/views/courses/csv/_filter.html.erb
cwrw: confirmed doesn't work

PR #494, opened 2020-03-02
Get 927 tabbed panel

tavy88:
### Context

Add tabs to skills matcher page

Also, add tabs to job profiles search page. The way
the original GDS component works needs a bit of tweaking
as follows:

The inactive tab has the link to the destination page and
not an anchor link as the component would expect. To achieve
that the inactive tab must not have the govuk-tabs__tab class,
otherwise, it will error as it's expecting something like:
'#element-id' which would not work well in our case.

### Screenshots

Screen Shot 2020-03-01 at 17 09 08
Screen Shot 2020-03-01 at 17 09 15
Screen Shot 2020-03-01 at 17 09 23
Screen Shot 2020-03-01 at 17 09 30

### Ticket

https://dfedigital.atlassian.net/browse/GET-927

### DIFF
--- a/app/views/job_profiles/_search_form.html.erb
+++ b/app/views/job_profiles/_search_form.html.erb
@@ -1,8 +1,7 @@
-<h1 class="govuk-heading-xl govuk-!-margin-bottom-7"><%= t(:title, scope: 'job_profiles.index') %></h1>
-<p class="govuk-body-l"><%= t(:body, scope: 'job_profiles.index') %></p>
+<p class="govuk-body"><%= t(:body, scope: 'job_profiles.index') %></p>
 <%= render "shared/search/results_form", search_path: results_job_profiles_path, scope: 'job_profiles.index' %>
-<p class="govuk-body">Select the job title that best matches what you're looking for.</p>
-<p class="govuk-body">If you can't see the job title you're looking for, try searching with different words - for example, checkout operator instead of cashier.</p>
 <p class="govuk-body">
   <%= "#{pluralize(@results.total_count, 'result', locale: I18n.locale.to_s)} found" %>
 </p>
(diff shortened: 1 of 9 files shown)


app/views/skills_matcher/index.html.erb
```diff
@@ -12,25 +12,37 @@
...
+        <% end %>
+      </section>
+    </div>
   </div>
   <div class="govuk-grid-column-one-third">
     <%= render "shared/contact_us" %>
     <%= render 'shared/save_or_return_to_your_results' %>
   </div>
 </div>
+
```
cwrw: 🦂

app/views/job_profiles/_search_form.html.erb
```diff
@@ -1,8 +1,7 @@
-<h1 class="govuk-heading-xl govuk-!-margin-bottom-7"><%= t(:title, scope: 'job_profiles.index') %></h1>
-<p class="govuk-body-l"><%= t(:body, scope: 'job_profiles.index') %></p>
+<p class="govuk-body"><%= t(:body, scope: 'job_profiles.index') %></p>
 <%= render "shared/search/results_form", search_path: results_job_profiles_path, scope: 'job_profiles.index' %>
-<p class="govuk-body">Select the job title that best matches what you're looking for.</p>
-<p class="govuk-body">If you can't see the job title you're looking for, try searching with different words - for example, checkout operator instead of cashier.</p>
 <p class="govuk-body">
   <%= "#{pluralize(@results.total_count, 'result', locale: I18n.locale.to_s)} found" %>
 </p>
+<p class="govuk-body">If you can’t see the job title you want, try searching with different words. For example, search for checkout operator instead of cashier.</p>
```
cwrw: this appears different in the prototype than the mockups in the ticket (ordering of lines) not sure which one is the correct one?

tavy88: It is confirmed with Lois.

PR #495, opened 2020-03-02
Get 523 enable bing

tavy88:
### Context

Re-enable Bing Spell Check service.

Also add a new key for bing spell check api endpoint,
as with cognitive services enabled on all our
subscriptions the API also changes per subscription.

### Screenshots

Screen Shot 2020-03-02 at 07 34 11

### Ticket

https://dfedigital.atlassian.net/browse/GET-523

### DIFF
--- a/app/services/spell_check_service.rb
+++ b/app/services/spell_check_service.rb
@@ -1,8 +1,6 @@
 class SpellCheckService
   SpellCheckServiceError = Class.new(StandardError)
 
-  API_ENDPOINT = 'https://api.cognitive.microsoft.com/bing/v7.0/spellcheck'.freeze
-
   attr_reader :response, :api_key
 
   def initialize(api_key: Rails.configuration.bing_spell_check_api_key)
@@ -51,7 +49,7 @@ class SpellCheckService
   end
 
   def uri
-    @uri ||= URI.parse(API_ENDPOINT)
+    @uri ||= URI.parse(Rails.configuration.bing_spell_check_api_endpoint)
   end
 
   def request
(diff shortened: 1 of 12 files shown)


app/services/spell_check_service.rb
```diff
@@ -1,8 +1,6 @@
 class SpellCheckService
   SpellCheckServiceError = Class.new(StandardError)
 
-  API_ENDPOINT = 'https://api.cognitive.microsoft.com/bing/v7.0/spellcheck'.freeze
```
cwrw: spoke in person, for posterity: pass in the endpoint in the initialiser and guard against it the same way as the api key

spec/services/spell_check_service_spec.rb
```diff
@@ -20,9 +20,13 @@
   let(:search_term) { 'hallo wrld' }
   let(:status) { 200 }
 
+  before do
+    allow(Rails.configuration).to receive(:bing_spell_check_api_endpoint).and_return('https://s111-bingspellcheck.cognitiveservices.azure.com/bing/v7.0/spellcheck')
```
cwrw: instead of this can you pass the endpoint here? `subject(:service) { described_class.new(api_key: 'test') }`

spec/services/spell_check_service_spec.rb
```diff
@@ -312,4 +316,12 @@
       expect(service.scan(search_term: search_term)).to be nil
     end
   end
+
+  context 'when the API endpoint is missing' do
+    subject(:service) { described_class.new(api_key: 'test', api_endpoint: nil) }
```
cwrw: can you do the same on line 313 (but the opposite)?

tavy88: 👍

PR #504, opened 2020-03-10
GET-1009 jmeter performance tests

bkarakashev:
### Context
JMeter-performance-tests


azure/jmeter/test.sh
```diff
@@ -0,0 +1,32 @@
...
+	-Jhost=$host \
+	-n -t /$testFile -l $T_DIR/test-plan.jtl -j $T_DIR/jmeter.log \
+	-e -o $R_DIR
+
+# echo "==== jmeter.log ===="
+# cat $T_DIR/jmeter.log
+
+# echo "==== Raw Test Report ===="
+# cat $T_DIR/test-plan.jtl
+
```
tavy88: Do we still need these ☝️ ?

bkarakashev: nope I will clean it up

azure-pipelines-jmeter-performance-tests.yml
```diff
@@ -0,0 +1,19 @@
...
+  displayName: Run JMeter tests
+  steps:
+  - task: Bash@3
+    displayName: Execute JMeter tests
+    inputs:
+      targetType: filePath
+      filePath: 'azure/jmeter/test.sh'
+      arguments: '$PWD test/jmeter-performance-tests.jmx $(host)'
+      workingDirectory: azure/jmeter
+      failOnStderr: false
```
cwrw: do we not want to fail this if it errors?

bkarakashev: It is a bit odd. as it is failing even if it is all pass. I will try to investigate and see if I can find out why is failing if there is no error... I assume is reading in the report where it says error % and its things that is an error... and marks the build as failed.

azure/jmeter/run.sh
```diff
@@ -0,0 +1,13 @@
...
+#
+# Run JMeter Docker image with options
+
+NAME="jmetertest"
+IMAGE="justb4/jmeter:latest"
+ROOTPATH=$1
+
+echo "$ROOTPATH"
+# Finally run
+docker stop $NAME > /dev/null 2>&1
```
cwrw: discussed in person: attempt to remove

azure-pipelines-jmeter-performance-tests.yml
cwrw: ignore for now, but docker fix might work, not a blocker

PR #508, opened 2020-03-11
Get-1008 Refine JMeter Test

bkarakashev:
### Context
Refine JMeter Test


tavy88:
Discussed in-person to run the job from the container.

PR #515, opened 2020-03-13
Remove keys from dev1, as we will use it as a stress test env

cwrw:
:shipit:


tavy88:
The description is empty btw.

PR #516, opened 2020-03-13
Add the courses near me error page for search

tavy88:
### Context
If the NCS API is down we want to make sure users are
still able to use the service and get a relevant error
message when trying to find a relevant course.

### Ticket
https://dfedigital.atlassian.net/browse/GET-1061

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -4,17 +4,11 @@ class CoursesController < ApplicationController
   ].freeze
 
   DELIVERY_TYPES = [
-    %w[All all],
-    ['Classroom based', '1'],
-    %w[Online 2],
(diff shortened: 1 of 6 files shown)


app/controllers/courses_controller.rb
```diff
@@ -45,7 +47,8 @@ def find_courses
     redirect_to course_postcode_search_error_path
   end
 
-  def find_csv_courses
+  # TODO: Move FindACourseService::APIError rescue to index when we decomission the the CSV Courses feature
+  def find_csv_courses # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
tavy88: This hopefully can be solved when the filter persistence PR is merged as it has a bit of refactoring.

PR #525, opened 2020-03-18
Render the decorated course description

tavy88:
### Context

It looks like we were not using the decorated course description when rendering.

### Screenshot

Screen Shot 2020-03-18 at 12 52 39

### Ticket
https://dfedigital.atlassian.net/browse/GET-1078

### DIFF
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -62,10 +62,10 @@
         <% end %>
       </tbody>
     </table>
-    <% if @decorated_course_details.description.present? %>
+    <% if @decorated_course_details.course_description.present? %>
       <h2 class="govuk-heading-m">Course description</h2>
       <p class="govuk-body">
-        <%= @decorated_course_details.description %>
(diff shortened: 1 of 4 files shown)


app/views/courses/show.html.erb
```diff
@@ -62,10 +62,10 @@
         <% end %>
       </tbody>
     </table>
-    <% if @decorated_course_details.description.present? %>
+    <% if @decorated_course_details.course_description.present? %>
```
cwrw: I think this should remain a check on description itself as when we get to the decorator stage it will fail on the `size` check if it's nil

tavy88: I don't think so as the check for size is done in the decorated method. If I keep `description.present?` with a description = '.' the condition will pass. The decorated method return if the size is < 15.

tavy88: But I need to guard against a description being nil in the decorated method as well.

cwrw: ok that's fair

PR #526, opened 2020-03-18
GET-1085 Search by math instead of maths for courses

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-1085

* Search by keyword math instead of maths for courses API to
get more results

### DIFF
--- a/spec/models/course_search_spec.rb
+++ b/spec/models/course_search_spec.rb
@@ -105,7 +105,7 @@ RSpec.describe CourseSearch do
       course_search.search
       expect(find_a_course_service).to have_received(:search).with(
         options: {
-          keyword: 'maths',
+          keyword: 'math',
           distance: '20',
           qualification_levels: %w[1 2 X E],
           postcode: 'NW6 8ET',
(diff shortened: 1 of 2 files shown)


spec/models/course_search_spec.rb
```diff
@@ -279,6 +279,62 @@
...
+      )
+      allow(FindACourseService).to receive(:new).and_return(find_a_course_service)
+
+      course_search = described_class.new(
+        postcode: 'NW6 8ET',
+        topic: 'physics'
+      )
+
+      course_search.search
+      expect(find_a_course_service).to have_received(:search).with(
```
tavy88: Hmm I thought we were rejecting anything but English and maths from the routing level?

cwrw: from the routing level yes, but this is a search class though, which will accept any type of course

tavy88: Discussed in person. My personal view is that leveraging the test that uses `english` as keyword is enough as english <> math,  and so is physics. I believe there is a bit redundancy here but not going to block it. So 👍

PR #527, opened 2020-03-18
GET-1050 Track course type and course hours

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-1050

* Track course hours if selected from dropdown and different than all
* Track course type if selected from dropdown and different than all
* Reuse the constant for the dropdown but change to hash to reverse
lookup the values for the tracking

GA qa key added to heroku app for testing

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -1,4 +1,4 @@
-class CoursesController < ApplicationController
+class CoursesController < ApplicationController # rubocop:disable Metrics/ClassLength
   DISTANCE = [
     ['Up to 10 miles', '10'], ['Up to 20 miles', '20'], ['Up to 30 miles', '30'], ['Up to 40 miles', '40']
   ].freeze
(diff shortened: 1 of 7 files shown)


app/controllers/courses_controller.rb
```diff
@@ -1,4 +1,4 @@
-class CoursesController < ApplicationController
+class CoursesController < ApplicationController # rubocop:disable Metrics/ClassLength
```
cwrw: keeping this for now until we remove the flag, which will allow us to refactor this controller

app/controllers/courses_controller.rb
```diff
@@ -112,9 +104,23 @@ def course_details
...
+    track_course_filter_for(
+      parameter: courses_params[:delivery_type],
+      value_mapping: DELIVERY_TYPES,
+      label: 'events.course_type_filter'
+    )
+    track_course_filter_for(
+      parameter: courses_params[:hours],
+      value_mapping: HOURS,
+      label: 'events.course_hours_filter'
+    )
```
tavy88: Suggestion:

If you had another method:

```ruby
def track_course_filters_for(options: [])
options.each do |option|
  track_course_filter_for(
      parameter: options[:key],
      value_mapping: options[:key][:mappings],
      label: options[:key][:locale]
    )
end
end
```

That would reduce the method length which is a bit of a smell (as per Rubocop). Thoughts?

I did something similar with GA Backend Tracking.

PR #529, opened 2020-03-19
Courses details formatting

tavy88:
### Context
Following formatting rules are introduced:

1. Add a 0 to prices like "£362.0"
2. Map QualificationLevel

 ```ruby
     E = Entry Level
     X = Unknown
     1 - Level 1
     2 - Level 2
      ...
```

3. Map CourseHours

```ruby
    PartTime = Part-time
    FullTime = Full-time
    Flexible = Flexible
```

4. Map CourseType

```ruby
    ClassroomBased = Classroom based
    WorkBased = Work based
    Online = Online
```

5. Don’t show CourseHours row if CourseHours = undefined

### Screenshots

Screen Shot 2020-03-19 at 13 45 06

### Ticket
https://dfedigital.atlassian.net/browse/GET-1084

### DIFF
--- a/app/decorators/course_details_decorator.rb
+++ b/app/decorators/course_details_decorator.rb
@@ -42,6 +42,13 @@ class CourseDetailsDecorator < SimpleDelegator
     description
   end
 
+  def course_qualification_level
+    return 'Unknown' if qualification_level == 'X'
+    return 'Entry Level' if qualification_level == 'E'
+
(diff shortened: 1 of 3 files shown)


tavy88:
I did think about that and that is something we need to discuss with content team/Steve and it's not part of the scope of this ticket. It's worth raising with Lois (I will do it and maybe create a ticket).

PR #530, opened 2020-03-20
GET-1083 Courses near me results formatting

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-1083

* Show start date/flexible date on results page
* Hide undefined course hours
* Format location and remove n/a values

### DIFF
--- a/app/views/courses/_search_course.html.erb
+++ b/app/views/courses/_search_course.html.erb
@@ -1,5 +1,5 @@
 <li>
-  <h2 class="govuk-heading-m govuk-!-margin-bottom-0 govuk-heading-l"><%= link_to(course.name, course_details_path(topic_id: params[:topic_id], course_id: course.course_id, course_run_id: course.course_run_id), class: 'govuk-link') %></h2>
+  <h2 class="govuk-heading-m govuk-!-margin-bottom-2"><%= link_to(course.name, course_details_path(topic_id: params[:topic_id], course_id: course.course_id, course_run_id: course.course_run_id), class: 'govuk-link') %></h2>
   <h3 class="govuk-heading-s govuk-!-margin-bottom-0"><span class="govuk-visually-hidden">Provider:</span><%= course.provider_name %></h3>
   <% if course.distance %>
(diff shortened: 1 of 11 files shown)


app/views/courses/_search_course.html.erb
```diff
@@ -5,7 +5,14 @@
     <p class="govuk-body govuk-!-margin-bottom-0">Distance: <strong><%= course.distance.round(1) %> miles</strong></p>
   <% end %>
   <p class="govuk-body govuk-!-margin-bottom-0">Course type: <strong><%= course.course_type  %></strong></p>
-  <p class="govuk-body govuk-!-margin-bottom-0">Course hours: <strong><%= course.course_hours %></strong></p>
-  <p class="govuk-body-s govuk-!-margin-top-2"><%= course.full_address %></p>
-  <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible govuk-!-margin-top-0">
+  <% if course.course_hours != 'Undefined' %>
```
tavy88: Could you use downcase here for safety as we are not 100% sure it will always come capitalised like this? I did something similar to mine.

spec/decorators/search_course_decorator_spec.rb
```diff
@@ -2,48 +2,70 @@
...
 
-      expect(described_class.new(search_course).full_address).to eq('Queens Gardens Sites HU1 3DG')
+      expect(described_class.new(search_course).full_address).to eq('Queens Gardens Sites, HU1 3DG')
     end
 
-    it 'adds line break to one comma' do
+    it 'removes n/a values followed by comma' do
       search_course = SearchCourse.new(
-        'venueAddress' => 'Kingston Upon Hull, HU1 3DG'
+        'venueAddress' => 'n/a, n/a, WA2 8QA'
```
tavy88: 'venueAddress' => 'n/a, n/a , WA2 8QA' will become: 

```ruby
 , WA2 8QA
```

I would maybe add the extra space before the comma in the regex as I think it's a likely case.

tavy88: Talked in person - will ignore this for now and see how future data might look like. No cases like this as it stands were seen yet.

PR #534, opened 2020-03-24
All page titles to follow the new convention

tavy88:
### Context
Change page titles where necessary
Change URL where necessary
Don’t do anything to breadcrumbs in this ticket

### Ticket
https://drive.google.com/file/d/1GGfYS5LNXugi5ua5xdfvQYHmrjSDtgUl/view

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -45,7 +45,8 @@ Rails.application.routes.draw do
     get 'job-hunting-questions', to: 'questions#job_hunting'
     post 'job-hunting-questions', to: 'questions#job_hunting_answers'
     get 'course-postcode-search-error', to: 'errors#course_postcode_search_error'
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -45,12 +45,12 @@
     get 'job-hunting-questions', to: 'questions#job_hunting'
     post 'job-hunting-questions', to: 'questions#job_hunting_answers'
     get 'course-postcode-search-error', to: 'errors#course_postcode_search_error'
-    get 'return-to-saved-results-error', to: 'errors#return_to_saved_results_error'
-    get 'save-results-error', to: 'errors#save_results_error'
+    get 'return-to-saved-progress-error', to: 'errors#return_to_saved_results_error'
+    get 'save-your-progress-error', as: 'save-results-error', to: 'errors#save_results_error'
```
tavy88: This needs to drop the alias of the old route that is no longer used.

PR #535, opened 2020-03-25
GET- 974 Add new cookie banner

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-974

* update policy page as we no longer use azure app insights and to
reflect new cookie policy guidelines
* User's selected values for cookies should be saved to a session. We
are using boolean now to show a user accepted all cookies or partial
cookies
*  show/hide cookie banner as a session available on for user and avoid creating session on home page
* Amend cookies modal to not rely on js and appear in different forms
with and without js. With js we see two options, without only one option.
* Hide/show content according to js being enabled or disabled
* When js is disabled we are ok for now not to trap focus
* When a user accepts all cookies user session set to true, when
partial set to false

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -25,8 +25,9 @@
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     </script>
     <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
-    <%= render 'shared/cookies_banner' if show_cookie_banner? %>
-    <% if !show_cookie_banner? %>
+    <% if show_cookie_banner? %>
+      <%= render 'shared/cookies_banner' %>
+    <% else %>
       <div id="govuk-header-container">
         <header class="govuk-header" role="banner" data-module="govuk-header" id="govuk-header">
           <div class="govuk-header__container govuk-width-container">
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -29,12 +29,7 @@
       <%= render 'shared/cookies_banner' %>
     <% else %>
       <div id="govuk-header-container">
-        <header class="govuk-header" role="banner" data-module="govuk-header" id="govuk-header">
-          <div class="govuk-header__container govuk-width-container">
-            <%= render 'shared/navigation/header_logo' %>
-            <%= render 'shared/navigation/main_header' %>
-          </div>
-        </header>
+        <%= render partial: 'shared/header', locals: { container_class: nil } %>
       </div>
     <% end %>
 
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -25,15 +25,17 @@
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     </script>
     <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
-    <%= render 'shared/cookies_banner' unless current_page?(cookies_policy_path) %>
-    <div id="govuk-header-container">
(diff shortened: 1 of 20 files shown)


app/views/layouts/application.html.erb
```diff
@@ -25,15 +25,17 @@
...
-    <%= render 'shared/cookies_banner' unless current_page?(cookies_policy_path) %>
-    <div id="govuk-header-container">
-      <header class="govuk-header" role="banner" data-module="govuk-header" id="govuk-header">
-        <div class="govuk-header__container govuk-width-container">
-          <%= render 'shared/navigation/header_logo' %>
-          <%= render 'shared/navigation/main_header' %>
-        </div>
-      </header>
-    </div>
+    <%= render 'shared/cookies_banner' if show_cookie_banner? %>
```
cwrw: if/else here

app/views/pages/cookies_policy.html.erb
```diff
@@ -12,20 +12,23 @@
 <div class="govuk-grid-row govuk-!-margin-top-7 govuk-body">
   <div class="govuk-grid-column-full">
```
cwrw: update content with lois

app/webpacker/styles/_cookies.scss
```diff
@@ -1,5 +1,19 @@
...
+
+  .cookies-modal-summary {
+    display: none;
+  }
+}
+
+.cookies-modal-details {
   display: none;
+}
+
```
cwrw: remove space

spec/features/google_analytics_tracking_spec.rb
```diff
@@ -13,6 +14,16 @@
     allow(Rails.configuration).to receive(:google_analytics_tracking_id).and_return(nil)
 
     visit(root_path)
+    click_on('Accept all cookies')
+
+    expect(page.source).not_to include('gtag')
+  end
+
+  scenario 'snippet is not included when user only accepts necessary cookies' do
+    allow(Rails.configuration).to receive(:google_analytics_tracking_id).and_return(nil)
```
cwrw: change to value instead of nil

app/views/layouts/application.html.erb
```diff
@@ -25,15 +25,18 @@
...
+    <% if show_cookie_banner? %>
+      <%= render 'shared/cookies_banner' %>
+    <% else %>
+      <div id="govuk-header-container">
+        <header class="govuk-header" role="banner" data-module="govuk-header" id="govuk-header">
+          <div class="govuk-header__container govuk-width-container">
+            <%= render 'shared/navigation/header_logo' %>
+            <%= render 'shared/navigation/main_header' %>
+          </div>
+        </header>
```
tavy88: I think you can put this in a partial:

```ruby
<header class="govuk-header" role="banner" data-module="govuk-header" id="govuk-header">
    <div class="govuk-header__container govuk-width-container  <%= margin.present? : 'govuk-!-padding-bottom-2' : '' %>">
        <%= render 'shared/navigation/header_logo' %>
        <%= render 'shared/navigation/main_header' %>
    </div>
  </header>
```

And then can do here:

```ruby
render partial: 'header'
```

and in the cookies banner:

```ruby
render partial: 'header', locals: { margin: 'govuk-!-padding-bottom-2'}
```

Should dry things a bit more.

config/routes.rb
```diff
@@ -99,6 +99,8 @@
     get 'link-expired', to: 'users#link_expired'
     get '/sign-in/:token', to: 'passwordless/sessions#show', authenticatable: 'user', as: :token_sign_in
 
+    post 'update-cookie-preferences', to: 'cookies#update'
```
tavy88: I get the choice for update here. We discussed in person I think it's fine to keep it this way even if it's more like create from a Rails convention pov.

PR #538, opened 2020-03-30
Get 1087 update breadcrumbs

tavy88:
### Context

For more consistency across user journey we would like
to update the breadcrumbs across the whole service as follows:

Home > Your information
Home > Steps
Home > Steps > Check your skills
Home > Steps > Check your skills > Check your skills results
Home > Steps > Check your skills > Search results > Current job skills
Home > Steps > Your skills
Home > Steps > Check your skills
Home > Steps > Your job matches
Home > Steps > Your job matches > Job profile details
Home > Steps > Action plan
Home > Steps > Action plan > Courses near you
Home > Steps > Action plan > Courses near you
Home > Steps > Action plan > CV advice
Home > Steps > Action plan > Cover letter advice
Home > Steps > Action plan > Interview advice
Home > Steps > Action plan > Offers near you
Home > Steps > Action plan > Jobs near you

Add Back functionality for the following sections:

- user registration
- training questions
- user sign in
- save progress error page
- link expired page

The logic needs updated as follows:

1. Save your progress journey
2. Return to your saved progress journey
3. Training questions

Because of a few edge cases, the back_link is now able
to accept a custom url for redirection and can also ignore
paths.

The current Back link implementation needs to ignore
paths like:

- email-sent-again which is only accessible via POST
- link-sent again which is only accessible via POST

Training questions, because they come in sequence, and
according to GDS pattern the user needs to be able to navigate
throuhg every question using the Back link, so adding
explicit redirect routes fixes this issue.

### Ticket
https://dfedigital.atlassian.net/browse/GET-1087

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -43,8 +43,7 @@ module ApplicationHelper
   def back_url(paths_to_ignore)
     url = url_parser.get_redirect_path(paths_to_ignore: paths_to_ignore)
 
-    return root_path unless url.present?
-    return task_list_path if request.original_url.include?(url)
+    return root_path if url.nil? || request.original_url.include?(url)
 
     url
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -29,22 +29,14 @@ module ApplicationHelper
   end
 
   def back_link(custom_url: nil, paths_to_ignore: [])
-    link_to('Back', custom_url || back_url(paths_to_ignore), class: 'govuk-back-link')
+    back_url = custom_url || url_parser.get_redirect_path(paths_to_ignore: paths_to_ignore) || root_path
+
+    link_to('Back', back_url, class: 'govuk-back-link')
   end
 
   def show_cookie_banner?
     return if current_page?(cookies_policy_path)
 
     cookies['_get_help_to_retrain_session'].blank? || user_session.cookies.nil?
-  end
-
-  private
-
-  def back_url(paths_to_ignore)
-    url = url_parser.get_redirect_path(paths_to_ignore: paths_to_ignore)
-
-    return root_path unless url
-
-    url
   end
 end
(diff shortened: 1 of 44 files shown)


app/helpers/application_helper.rb
```diff
@@ -28,15 +28,23 @@ def target_job
     @target_job ||= JobProfile.find_by(id: user_session.target_job_id)
   end
 
-  def back_link
-    back_url = url_parser.get_redirect_path || root_path
-
-    link_to('Back', back_url, class: 'govuk-back-link')
+  def back_link(custom_url: nil, paths_to_ignore: [])
+    link_to('Back', custom_url || back_url(paths_to_ignore), class: 'govuk-back-link')
```
cwrw: I think moving the logic here might be easier to parse:
```
url = custom_url || url_parser.get_redirect_path(paths_to_ignore: paths_to_ignore) || root_path
```
what do you think?

config/features.rb
```diff
@@ -5,5 +5,5 @@
 
   feature :foo, description: 'Example feature flag', default: true
   feature :health_check, description: 'Dummy feature for split.io health check'
-  feature :csv_courses, description: 'Enable service to use the new CSV courses data'
+  feature :csv_courses, description: 'Enable service to use the new CSV courses data', default: true
```
cwrw: 🔥 🔥

spec/features/questions_spec.rb
```diff
@@ -9,13 +9,38 @@
...
+  scenario 'User can go back to previous page from IT training questions page' do
+    user_targets_job
+    click_on('Continue')
+
+    click_on('Back')
+
+    expect(page).to have_current_path(training_questions_path)
+  end
+
+  scenario 'User goes back to training questions path when deeplinking to it training questions page on a non empty session and then hitting Back' do
```
cwrw: not sure I understand this test?

app/helpers/application_helper.rb
tavy88: 👍

config/features.rb
tavy88: It doesn't make much difference tbh since our feature is on anyway and next sprint we are removing it completely.

spec/features/questions_spec.rb
tavy88: Yeah, I think this is no longer needed as we changed the strategy.

PR #541, opened 2020-03-31
GET-1042 Improve job profile search

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-1042

* Add new columns sector and hierarchy to job profiles table
* Add new rake task which accepts the hierarchy/sector spreadsheet and populates those fields in job profiles table
* Improve exact match in job profile search and separate ranking
* Boost job profile search when a work matches the profile sector
* Push down (secondary) profile when matching hierarchy


app/models/job_profile_search.rb
```diff
@@ -1,6 +1,22 @@
...
+    alternative_titles: { weight: 'AB', score: '4|32' },
+    specialism: { weight: 'ABC', score: '4|1|32' },
+    hidden_titles: { weight: 'ABC', score: '4|1|32' },
+    description: { weight: 'ABCD', score: '4|8|32' }
+  }.freeze
+
+  PARTIAL_MATCH_RANK = {
+    name: { weight: 'A', score: '4|2' },
+    alternative_titles: { weight: 'B', score: '8|2' },
+    specialism: { weight: 'C', score: 0 },
```
cwrw: put in string

app/models/job_profile_search.rb
```diff
@@ -1,6 +1,27 @@
-class JobProfileSearch
+class JobProfileSearch # rubocop:disable Metrics/ClassLength
   include ActiveModel::Validations
 
+  EXACT_MATCH_RANK = {
+    name: { weight: 'A', score: '4|1' },
+    alternative_titles: { weight: 'AB', score: '4|32' },
+    specialism: { weight: 'ABC', score: '4|1|32' },
+    hidden_titles: { weight: 'ABC', score: '4|1|32' },
+    description: { weight: 'ABCD', score: '4|8|32' }
```
tavy88: So `specialism` and `hidden_titles` seem to have exactly the same priority which makes me think how important their contribution is and if we need both? I believe there should be a priority of one on top of the other? Have you tried fiddling with their ranking scores to see if you get any improvement?

app/models/job_profile_search.rb
```diff
@@ -1,6 +1,27 @@
...
+    specialism: { weight: 'C', score: '0' },
+    hidden_titles: { weight: 'C', score: '0' },
+    description: { weight: 'D', score: '4|8' },
+    sector: { weight: 'A', score: '0' }
+  }.freeze
+
+  HIERARCHY_MATCH_RANK = {
+    hierarchy: { weight: 'A', score: '0' }
+  }.freeze
+
```
tavy88: Discussed in person about documenting the rationale of picking these scores for each of the cases above.

app/services/job_profile_search_service.rb
```diff
@@ -0,0 +1,78 @@
...
+        hierarchy: nil,
+        sector: nil
+      ).count
+    }
+  end
+
+  private
+
+  def update_job_profile(data)
+    return if data[:type] == 'doubtful' || data[:type] == 'meaningless'
```
tavy88: Could also do 

```ruby
return if %w(doubtful meaningless).include?(data[:type])
```

app/services/job_profile_search_service.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    update_job_profiles!(job_profiles, data[:type], words) if job_profiles.any?
+  end
+
+  def build_word_array(data)
+    words = [data[:word]]
+    alternatives = data[:alternatives]&.split(', ')
+    words += alternatives if alternatives
+
+    words
```
tavy88: I think this can be done in less steps:

```ruby
def build_word_array(data)
   alternatives = data[:alternatives]&.split(', ')

   [data[:word], alternatives].flatten.compact
end
```

spec/models/job_profile_search_spec.rb
```diff
@@ -215,6 +243,55 @@
...
+        name: 'Office assistant'
+      )
+
+      job_profiles_in_order = [
+        business_assistant,
+        office_assistant,
+        kitchen_assistant
+      ]
+
+      expect(described_class.new(term: 'Business assistant').search).to eq(job_profiles_in_order)
```
tavy88: We could use:

```ruby
subject { described_class.new(term: term) }
let(:term) { 'term per context' }
```

app/models/job_profile_search.rb
cwrw: discussed in person: both fields are of the same weight and serve different purposes in the search context

PR #544, opened 2020-04-02
GET-1091 Show validation error on postcode API error

cwrw:
jira-ticket: https://dfedigital.atlassian.net/browse/GET-1091

* Add response body to error message
* Raise postcode not found error when error body from API contains
the relevant message
* Catch the error message at the course search level and add
the validation error postcode not found to the object. We have
to check for errors then validation as `valid?` will overwrite
the existing errors and remove them.
* Add authenticity token to params to avoid errors

### DIFF
--- a/app/models/course_search.rb
+++ b/app/models/course_search.rb
@@ -17,7 +17,7 @@ class CourseSearch
   end
 
   def search
-    return [] unless valid?
+    return [] if errors.present? || invalid?
 
     find_a_course_response
       .fetch('results', [])
@@ -25,7 +25,7 @@ class CourseSearch
   end
 
   def count
-    return unless valid?
+    return 0 if errors.present? || invalid?
 
     find_a_course_response['total'] || 0
   end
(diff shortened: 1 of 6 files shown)


app/models/course_search.rb
```diff
@@ -82,8 +82,19 @@ def coordinates
...
+        )
+      rescue FindACourseService::PostcodeNotFoundError
+        add_non_existing_postcode_error
+
+        {}
+      end
+  end
+
+  def add_non_existing_postcode_error
+    errors.add(:postcode, I18n.t('courses.nonexisting_postcode_error'))
```
tavy88, line 98: I wonder if we should say `Postcode not found` instead of `Enter a real postcode`? Because our postcode does exist but it's deemed as not found by NCS.

cwrw: After discussion in planning, we decided that their error maps to our postcode not real error, just with different wording, so I'm keen to keep as is for now

tavy88: 👍

PR #547, opened 2020-04-02
Authentication process to ignore the resource name

tavy88:
### Context
We recently noticed that resources in Azure Active Directory
can have the name attribute changed breaking our current
implementation which relied on the combination of
resource_id + email + name as the unique identifier
of authorised users for the Admin feature.

Relying instead on just resource_id + email as the unique
identifier will fix this issue.

### Ticket
https://dfedigital.atlassian.net/browse/GET-1095

### Testing
Need to manually update the name of your user in the database and then try to login via AAD.

### DIFF
--- a/spec/features/admin/authentication_spec.rb
+++ b/spec/features/admin/authentication_spec.rb
@@ -72,6 +72,18 @@ RSpec.feature 'Admin Users Authentication' do
     end
   end
 
+  scenario 'Authenticates returning user correctly when the name in Active Azure Directory is different from the one persisted in our database' do
+    create(:admin_user, name: 'james.bond', email: 'test@test.com', resource_id: '1111-111-11-1')
+
+    stub_omniauth
+
+    visit(auth_azure_ad_auth_callback_path)
+
+    ['Dashboard', 'some.name', 'Logout'].each do |content|
+      expect(page).to have_content(content)
+    end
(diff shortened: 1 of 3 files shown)


spec/features/admin/authentication_spec.rb
```diff
@@ -72,6 +72,18 @@
     end
   end
 
+  scenario 'Authenticates returning user correctly when the name in Active Azure Directory is different from the one persisted in our database' do
+    create(:admin_user, name: 'james.bond', email: 'test@test.com', resource_id: '1111-111-11-1')
```
cwrw, line 76: 🔫

spec/models/admin_user_spec.rb
```diff
@@ -68,6 +68,20 @@
       expect(active_admin.valid?).to be true
     end
 
+    it 'correctly presists the identified resource name' do
```
cwrw, line 71: I think this test is a duplicate of the above (it returns all attributes including name)

tavy88: I wanted explicitly to make the distinction between storing and updating a record. Because you might set something, but then maybe have a condition that doesn't override the initially set value. I know there is a bit of redundancy here but just wanted to illustrate these 2 cases.

PR #558, opened 2020-04-07
Update copy on the Action Plan page

tavy88:
### Context

Update copy on the Action Plan page.

### Ticket
https://dfedigital.atlassian.net/browse/GET-1111


tavy88:
Yeah, I think I'll keep the change in one place. We will review again on QA anyway.

PR #607, opened 2020-06-12
SPIKE: Improve the skills matching

tavy88:
### Context

This is spiking a potential improvement over skills matching.


tavy88:
Spike has been done.

PR #608, opened 2020-06-18
Concept: Override target job when an alternative job title is passed

tavy88:
### Context
We want to allow users to surface additional job vacancies,
by being able to override the targetted job on the session.

Also fix the following bug:
- until now, if a page/distance param was not being passed
at the controller level, it was still explicitly passed as nil in
the JobVacancySearch causing the initial page
number to be nil.to_i = 0 instead of 1.
- similar situation for distance.

### Screenshot
Screen Shot 2020-06-18 at 16 32 04

### Ticket
https://dfedigital.atlassian.net/browse/GET-1163


tavy88:
This will no longer be needed.

